PR #6, opened 2017-11-14
feat(pubsub): initial asynchronous pubsub subscriber

TheKevJames:


pubsub/gcloud/aio/pubsub/subscription.py
```diff
@@ -0,0 +1,68 @@
...
+                google.gax.errors.RetryError):
+            # common intermittent communication errors
+            if not retries:
+                raise
+
+            await asyncio.sleep(1)
+        except Exception:
+            raise
+
+        return await self.acknowledge(ack_ids, client=client,
```
linusmixson: Should we avoid recursion here and opt for more informative exceptions? This could make a pretty hard-to-read stacktrace. Otherwise, looks great!

PR #12, opened 2017-11-27
feat(taskqueue): add perf-timing hookins to LocalTaskQueue

TheKevJames:
Allows users to capture processing time information when using the LocalTaskQueue emulation layer.

### DIFF
--- a/taskqueue/gcloud/aio/taskqueue/taskqueue.py
+++ b/taskqueue/gcloud/aio/taskqueue/taskqueue.py
@@ -5,6 +5,7 @@ import asyncio
 import datetime
 import json
 import logging
+import time
 
 import ujson
 from gcloud.aio.auth import Token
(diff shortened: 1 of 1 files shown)


taskqueue/gcloud/aio/taskqueue/taskqueue.py
```diff
@@ -228,21 +229,22 @@ def __init__(self, project, service_file, task_queue, session=None,
...
+        self.deleted = {}
         self.leased = {}
         self.ready = {}
-        self.deleted = {}
 
-        self.task_queue = kwargs.get('task_queue') or 'q:{}'.format(
-            self.next_id()
-        )
+        self.duration = 0
+        self.start_time = None
```
jonathan-johnston: `start_time` is the time the task queue first leased a task and `duration` is the length of time the queue has been operating with continuous tasks (i.e. not empty)? Is this for analytics (ex. recording utilization)?

TheKevJames: Yes, indeed -- we're using this in a few internal projects to measure performance of processing _n_ tasks off the queue.

PR #13, opened 2018-01-09
feat(taskqueue): migrate to cloudtasks v2beta2

TheKevJames:


TheKevJames:
@jonathan-johnston Ah, no, there are a huge number of changes here. Looks like there are so many changes that the file renames got picked up as completely new files.. give me a sec and I'll split the PR into a seperate commit for that, to make it easier to review. My bad!

jonathan-johnston:
Thanks @TheKevJames! That would make it a lot easier to visually parse :)

TheKevJames:
@jonathan-johnston fixed, sorry about that (also, rebased onto #14, FYI)

taskqueue/gcloud/aio/taskqueue/taskmanager.py
```diff
@@ -2,486 +2,159 @@
...
-                },
-                'created_at': str(self.created_at),
-                'state': getattr(self.state, 'name', None)
-            }
+        properties = {
+            'bucket': payload.get('bucket'),
+            'error': str(exception),
+            'generation': None,
+            'metageneration': None,
+            'time_created': datetime.datetime.now(),
```
jonathan-johnston: Do you have to worry about local time vs UTC conversions?

taskqueue/gcloud/aio/taskqueue/taskmanager.py
```diff
@@ -2,486 +2,159 @@
...
+            for task in tasks:
+                futures.append(asyncio.ensure_future(self.process(task)))
 
-    def try_to_lease(self):
+            await asyncio.gather(*futures)
 
-        if self.no_tasks >= self.max_concurrency:
+    async def process(self, task):
+        name = task['name']
+        payload = ujson.loads(decode(task['pullMessage']['payload']))
```
jonathan-johnston: Did you mention that the payload isn't always a JSON object?

taskqueue/gcloud/aio/taskqueue/taskmanager.py
```diff
@@ -2,486 +2,159 @@
...
 
-        self.hold_onto(id_, payload)
+            autorenew.cancel()
+            with contextlib.suppress(asyncio.CancelledError):
+                await autorenew
 
-        half_lease = self.lease_seconds / 2
-        call_later(half_lease, self.renew, id_, half_lease)
+            task = self.tasks[name]
+            del self.tasks[name]
```
jonathan-johnston: Any reason you're repeating these calls inside and outside the try-except block here? I might suggest something like this:
```
try:
    # code
except FirstError:
    # code
except SecondError:
    # code
finally:
    autorenew.cancel()
    with contextlib.suppress(asyncio.CancelledError):
        await autorenew

    task = self.tasks[name]
    del self.tasks[name]
```

This will trigger prior to the early returns in the `except` blocks, and will also trigger if no exception if hit.

taskqueue/gcloud/aio/taskqueue/taskmanager.py
TheKevJames: Yes indeed. We weren't doing this before, but we may as well do it now.

taskqueue/gcloud/aio/taskqueue/taskmanager.py
TheKevJames: Adding a comment.

taskqueue/gcloud/aio/taskqueue/taskmanager.py
TheKevJames: Wouldn't the order be incorrect? I need to cancel the `autorenew` thread and grab the task from it after it has been cancelled in order to use that task object for cancelling/acking/deleting.

```
kevin@atomicbox » python3.6
>>> try:
...  raise ValueError(42)
... except Exception:
...  print('except')
... finally:
...  print('finally')
...
except
finally
```

TheKevJames: Though I would love to de-dupe this block if we can think of a way to get the ordering right.

taskqueue/tests/integration/taskmanager_test.py
```diff
@@ -0,0 +1,61 @@
...
+
+        # DRAIN
+        await tm.tq.drain()
+
+        # START
+        tm.start()
+
+        # INSERT
+        for task in tasks:
+            await tm.tq.insert(encode(json.dumps(task)))
```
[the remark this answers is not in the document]
TheKevJames: Indeed! The `TaskQueue` class is now fully featured, instead of only having a `delete` method.

taskqueue/tests/integration/taskqueue_test.py
```diff
@@ -0,0 +1,72 @@
...
+                assert v != leased['tasks'][0][k]
+            else:
+                assert v == leased['tasks'][0][k]
+
+        # ack?
+        # cancel?
+
+        # DELETE
+        result = await tq.delete(renewed['name'])
+        assert not result
```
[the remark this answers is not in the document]
TheKevJames: True.. by standard JSON API spec delete 204s return no-content (see [jsonapi](http://jsonapi.org/format/#crud-deleting-responses-204)). In this case, I suppose I really just want to assert no exception was raised.

taskqueue/gcloud/aio/taskqueue/taskmanager.py
jonathan-johnston: Ah, gotcha, I missed the fact that you were using that return value. I can't think of a function name for the repeated code block so it's not worth pulling out into a function. Otherwise there are no other straight-forward and readable ways to de-dupe this block. This is fine!

PR #22, opened 2018-04-24
feat(storage): implement "list blobs" feature

TheKevJames:
This file hadn't been touched in forever, so I included a bunch of cleanup here. The functional changes are `Storage.list_objects` and `Bucket.list_blobs`, which can be used with an optional prefix to list all objects in a given bucket and/or in a given folder in a bucket.

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -18,45 +18,43 @@ READ_WRITE_SCOPE = 'https://www.googleapis.com/auth/devstorage.read_write'
 log = logging.getLogger(__name__)
 
 
-class Storage(object):
-
+class Storage:
     def __init__(self, project, service_file, token=None, session=None):
-
         self.service_file = service_file
+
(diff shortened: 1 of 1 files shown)


storage/gcloud/aio/storage/storage.py
```diff
@@ -141,17 +120,25 @@ def __init__(self, storage, name):
...
+        status, content = await self.storage.list_objects(self.name,
+                                                          params=params,
+                                                          session=session)
 
-        return blob
+        if status < 200 or status >= 300:
+            log.error('Could not list %s/%s: %s', self.name, prefix, content)
+            return
 
+        return [x['name'] for x in content.get('items', list())]
```
jonathan-johnston: Do you have to `content = ujson.loads(content)`?

TheKevJames: Nope, calling the `get` in the `Storage` method with `json_response=True` already does this.

PR #26, opened 2018-05-14
refactor(tests): use `pytest-asyncio`

TheKevJames:


TheKevJames:
Hmm, I'm having trouble debugging this. That integration test is behaving differently locally that in CI -- if #27 exhibits the same failure, we'll know this is a networking / CircleCI config issue.

PR #30, opened 2018-07-26
Reuse session.

Natim:


TheKevJames:
@Natim nice catch -- thanks for the contribution!

PR #39, opened 2018-10-11
feature(delete): add ability to delete specific object

overset:
~~I'll continue to retry circle tests to make sure this is passing~~

Edit: Circle tests passing


TheKevJames:
Would be great a test for this, but looks good either way.

overset:
> Would be great a test for this, but looks good either way.

Refactored the smoke test to go through more of the object life-cycle. Found an issue with https://github.com/talkiq/gcloud-aio/commit/8864d6d9a7fb2f987b0c94c41f9819902e5f2975 causing non-json file data being uploaded not able to be json serialized.

PR #40, opened 2018-10-12
fix(encoding): fix URI path encoding for GCP REST API

overset:
Added support for properly encoding Storage object names with special paths to follow https://cloud.google.com/storage/docs/json_api/#encoding

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -59,8 +59,6 @@ class Storage:
         session = session or self.session
         resp = await session.delete(url, headers=headers, params=params or {},
                                     timeout=60)
-        log.warning(url)
-        log.warning(await resp.json())
         resp.raise_for_status()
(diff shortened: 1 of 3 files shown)


storage/gcloud/aio/storage/storage.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import urllib
```
TheKevJames: Needs to be `import urllib.parse`

TheKevJames:
```
kevin@kjames0164 ~/c/p/t/docker-nox » python3
>>> import urllib
>>> urllib.parse.quote
AttributeError: module 'urllib' has no attribute 'parse'
>>>
kevin@kjames0164 ~/c/p/t/docker-nox » python3
>>> import urllib.parse
>>> urllib.parse.quote
<function quote at 0x10c442510>
>>>
```

PR #60, opened 2019-02-06
Update gcloud-datastore to use local development emulator

nickfirespotter:
Check for the DATASTORE_EMULATOR_HOST env variable and if it exists assume working in dev mode. Then update the API_ROOT to point to the emulator as well as not requiring Authorization header or the need to obtain an auth Token.

### DIFF
--- a/datastore/gcloud/aio/datastore/datastore.py
+++ b/datastore/gcloud/aio/datastore/datastore.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from typing import Any
 from typing import Dict
 from typing import List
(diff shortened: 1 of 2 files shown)


datastore/gcloud/aio/datastore/datastore.py
```diff
@@ -43,13 +52,21 @@ def __init__(self, project: Optional[str] = None,
...
+            self.token = None
+        else:
+            self.token = token or Token(service_file=service_file, session=session,
+                                        scopes=SCOPES)
 
     async def project(self) -> str:
         if self._project:
             return self._project
 
+        if _IS_DEV:
```
TheKevJames: I _think_ (but could be convinced either way) that this should take precedence over any `_project` value we've explicitly set, given we're effectively hijacking other user args such as `service_account` when an emulator is detected. Thoughts?

nickfirespotter: Worth consideration...
I think in general if a developer is going to be changing the project name between dev and prod (or staging) they have to be obtaining that project id from some variable. Either passed in to the running program or pulled from an env var. So I largely would think it wouldn't really matter.
I guess the question is if there are cases where someone would be connecting to 2 different projects in their code. And if so the emulator supports that or not... I can go either way.

TheKevJames: > 2 different projects
That's a great thought. We do that in a bunch of our codebases, to communicate between, say, the engineering production env and the datascience one. I suppose that would just be unsupported by the emulator. I can't really think of a decent way to support that sort of cross-project work involving an emulator, though, so I might put that off for a later version only if need be.

TheKevJames:
Merged and rolling out as release `gcloud-aio-datastore==3.1.0`. Thanks!

PR #62, opened 2019-02-06
auth: credentials file not found raise a `FileNotFoundError`

JensMadsen:


auth/gcloud/aio/auth/auth.py
```diff
@@ -55,6 +55,9 @@ class Token:
     def __init__(self, service_file: Optional[str] = None,
                  session: aiohttp.ClientSession = None,
                  scopes: List[str] = None) -> None:
+        if not os.path.isfile(service_file):
```
TheKevJames: This line throws a typeerror when `service_file` is `None` (ie. the default case). We should probably only perform this check if `service_file is not None`.

Do you think its also worth throwing this error if we fallback to `GOOGLE_APPLICATION_CREDENTIALS` and/or `CLOUDSDK_CONFIG` and they are explicitly set but not present?

JensMadsen: sorry about no checking this properly(typeerror instead of filenotfounderror). 

I did this pr because I wasted a lot of time due to a typo in my path to the credentialsfile. I just wanted to avoid similar thing to happen again. In that line I think it is a good idea to check if the user has specified it specifically. At least throw a warning

TheKevJames:
Thanks for the contribution, this error message will be helpful. Will roll this out in the next release :)

PR #48, opened 2018-11-23
Storage streams

JensMadsen:
#44 storage upload handle streams

PR #65, opened 2019-02-15
feat(storage): upload via streams

TheKevJames:
Memory efficient upload method first provided by @JensMadsen in #48.

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -1,11 +1,12 @@
 import asyncio
 import enum
+import io
 import logging
 import mimetypes
+import os
 from typing import Any
 from typing import Optional
 from typing import Tuple
-from typing import Union
 from urllib.parse import quote
 
 import aiohttp
@@ -111,8 +112,8 @@ class Storage:
                                                  read_timeout=10)
         session = session or self.session
 
-        data = self._preprocess_data(file_data)
-        content_length = len(data)
+        stream = self._preprocess_data(file_data)
+        content_length = self._get_stream_len(stream)
 
         # mime detection method same as in aiohttp 3.4.4
         content_type = content_type or mimetypes.guess_type(object_name)[0]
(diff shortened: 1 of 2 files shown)


storage/gcloud/aio/storage/storage.py
```diff
@@ -129,22 +130,34 @@ def get_bucket(self, bucket_name: str) -> Bucket:
...
         raise TypeError(f'upload type {upload_type} not supported')
 
     @staticmethod
-    def _preprocess_data(data: Any) -> Union[str, bytes]:
+    def _get_stream_len(stream: io.IOBase) -> int:
+        current = stream.tell()
+        try:
+            return stream.seek(0, os.SEEK_END)
+        finally:
+            stream.seek(current)
```
jonathan-johnston: Note:
Interesting use of the `try-except` block to rewind the stream. If you're curious, doing an assignment would be 2 instructions cheaper:
```
>>> def foo(data):
...     try:
...             return data
...     finally:
...             print(data)
... 
>>> dis.dis(foo)
  2           0 SETUP_FINALLY            4 (to 6)

  3           2 LOAD_FAST                0 (data)
              4 RETURN_VALUE

  5     >>    6 LOAD_GLOBAL              0 (print)
              8 LOAD_FAST                0 (data)
             10 CALL_FUNCTION            1
             12 POP_TOP
             14 END_FINALLY
             16 LOAD_CONST               0 (None)
             18 RETURN_VALUE
>>> def bar(data):
...     x = data
...     print(data)
...     return x
... 
>>> dis.dis(bar)
  2           0 LOAD_FAST                0 (data)
              2 STORE_FAST               1 (x)

  3           4 LOAD_GLOBAL              0 (print)
              6 LOAD_FAST                0 (data)
              8 CALL_FUNCTION            1
             10 POP_TOP

  4          12 LOAD_FAST                1 (x)
             14 RETURN_VALUE

```

TheKevJames: Good to know! In this case, I think its worth the extra ops, since that will additionally guarantee an exception raised by `stream.seek` does not affect whether we rewind or not.

PR #69, opened 2019-03-28
[WIP] feat(datastore) add non-GQL queries

KevinGDialpad:
[WIP] I still need to figure out how to run the integration tests - I wouldn't merge this before test_query in `smoke_test.py` is implemented (and the docs are updated). But the code is there so I think there is value in giving time to review while I work out the integration tests and the pre-commit hook (pylint did find some issues that I fixed, but it currently raises an Attribute error for some reason)

I added classes for Query, Filter, and Value. This adds support for non-GQL queries.

### DIFF
--- a/datastore/gcloud/aio/datastore/constants.py
+++ b/datastore/gcloud/aio/datastore/constants.py
@@ -2,6 +2,11 @@ import enum
 from datetime import datetime as dt
 
 
+class CompositeFilterOperator(enum.Enum):
+    OPERATOR_UNSPECIFIED = 'OPERATOR_UNSPECIFIED'
+    AND = 'AND'
+
+
 class Consistency(enum.Enum):
     EVENTUAL = 'EVENTUAL'
     STRONG = 'STRONG'
@@ -27,6 +32,16 @@ class Operation(enum.Enum):
     INSERT = 'insert'
     UPDATE = 'update'
     UPSERT = 'upsert'
+
+
+class PropertyFilterOperator(enum.Enum):
+    OPERATOR_UNSPECIFIED = 'OPERATOR_UNSPECIFIED'
+    LESS_THAN = 'LESS_THAN'
+    LESS_THAN_OR_EQUAL = 'LESS_THAN_OR_EQUAL'
+    GREATER_THAN = 'GREATER_THAN'
+    GREATER_THAN_OR_EQUAL = 'GREATER_THAN_OR_EQUAL'
+    EQUAL = 'EQUAL'
+    HAS_ANCESTOR = 'HAS_ANCESTOR'
 
 
 class ResultType(enum.Enum):
--- a/datastore/gcloud/aio/datastore/filter.py
+++ b/datastore/gcloud/aio/datastore/filter.py
@@ -1,0 +1,111 @@
+from typing import Any
+from typing import Dict
+from typing import List
+
+from gcloud.aio.datastore.constants import CompositeFilterOperator
+from gcloud.aio.datastore.constants import PropertyFilterOperator
+from gcloud.aio.datastore.value import Value
+
+
+class _BaseFilter:
+    json_key = ''
+
+    def __repr__(self) -> str:
+        return str(self.to_repr())
+
+    @classmethod
+    def from_repr(cls, data: Dict[str, Any]) -> '_BaseFilter':
+        raise NotImplementedError
+
+    def to_repr(self) -> Dict[str, Any]:
+        raise NotImplementedError
+
+
+# https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/runQuery#Filter
+class Filter:
+    def __init__(self, inner_filter: _BaseFilter):
+        self.inner_filter = inner_filter
+
+    def __repr__(self) -> str:
+        return str(self.to_repr())
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Filter):
+            return False
+        return self.inner_filter == other.inner_filter
+
+    @classmethod
+    def from_repr(cls, data: Dict[str, Any]) -> 'Filter':
+        (inner_filter_name, inner_filter_data), = data.items()
+        if inner_filter_name == 'compositeFilter':
+            inner_filter = CompositeFilter.from_repr(inner_filter_data)
+        elif inner_filter_name == 'propertyFilter':
+            inner_filter = PropertyFilter.from_repr(inner_filter_data)
+        else:
+            raise ValueError(f'Invalid filter name: {inner_filter_name}')
+        return cls(inner_filter=inner_filter)
+
+    def to_repr(self) -> Dict[str, Any]:
+        return {
+            self.inner_filter.__class__.json_key: self.inner_filter.to_repr()
+        }
+
+
+class CompositeFilter(_BaseFilter):
+    json_key = 'compositeFilter'
+
+    def __init__(
+            self, operator: CompositeFilterOperator, filters: List[Filter]):
+        self.operator = operator
+        self.filters = filters
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, CompositeFilter):
+            return False
+        return bool(
+            self.operator == other.operator
+            and self.filters == other.filters)
+
+    @classmethod
+    def from_repr(cls, data: Dict[str, Any]) -> 'CompositeFilter':
+        operator = CompositeFilterOperator(data['op'])
+        filters = [Filter.from_repr(f) for f in data['filters']]
+        return cls(operator=operator, filters=filters)
+
+    def to_repr(self) -> Dict[str, Any]:
+        return {
+            'op': self.operator.value,
+            'filters': [f.to_repr() for f in self.filters]
+        }
+
+
+class PropertyFilter(_BaseFilter):
+    json_key = 'propertyFilter'
+
+    def __init__(
+            self, prop: str, operator: PropertyFilterOperator, value: Value):
+        self.prop = prop
+        self.operator = operator
+        self.value = value
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, PropertyFilter):
+            return False
+        return bool(
+            self.prop == other.prop
+            and self.operator == other.operator
+            and self.value == other.value)
+
+    @classmethod
+    def from_repr(cls, data: Dict[str, Any]) -> 'PropertyFilter':
+        prop = data['property']['name']
+        operator = PropertyFilterOperator(data['op'])
+        value = Value.from_repr(data['value'])
+        return cls(prop, operator, value)
+
+    def to_repr(self) -> Dict[str, Any]:
+        return {
+            'property': {'name': self.prop},
+            'op': self.operator.value,
+            'value': self.value.to_repr()
+        }
--- a/datastore/gcloud/aio/datastore/query.py
+++ b/datastore/gcloud/aio/datastore/query.py
@@ -5,11 +5,60 @@ from typing import List
 from gcloud.aio.datastore.constants import MoreResultsType
 from gcloud.aio.datastore.constants import ResultType
 from gcloud.aio.datastore.entity import EntityResult
+from gcloud.aio.datastore.filter import Filter
 from gcloud.aio.datastore.utils import make_value
 from gcloud.aio.datastore.utils import parse_value
 
 
-class GQLQuery:
+class _BaseQuery:
+    json_key = ''
+
+    def __repr__(self) -> str:
+        return str(self.to_repr())
+
+    @classmethod
+    def from_repr(cls, data: Dict[str, Any]) -> '_BaseQuery':
+        raise NotImplementedError
+
+    def to_repr(self) -> Dict[str, Any]:
+        raise NotImplementedError
+
+
+# https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/runQuery#Query
+class Query(_BaseQuery):
+    json_key = 'query'
+
+    def __init__(self, kind: str, query_filter: Filter = None):
+        if not kind:
+            raise NotImplementedError('Kindless queries are not supported')
+        self.kind = kind
+        self.query_filter = query_filter
+
+    def __eq__(self, other: Any) -> bool:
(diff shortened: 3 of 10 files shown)


datastore/gcloud/aio/datastore/query.py
```diff
@@ -28,12 +77,9 @@ def __eq__(self, other: Any) -> bool:
             and self.named_bindings == other.named_bindings
             and self.positional_bindings == other.positional_bindings)
 
-    def __repr__(self) -> str:
-        return str(self.to_repr())
-
     @classmethod
     def from_repr(cls, data: Dict[str, Any]) -> 'GQLQuery':
-        allow_literals = data['allowLiteralls']
+        allow_literals = data['allowLiterals']
```
KevinGDialpad: Found this in passing

datastore/gcloud/aio/datastore/value.py
```diff
@@ -0,0 +1,63 @@
+from typing import Any
+from typing import Dict
+
+
+_JSON_KEYS = {
```
KevinGDialpad: I do intend to add the other types before this gets merged

datastore/gcloud/aio/datastore/query.py
TheKevJames: Nice catch, thanks

TheKevJames:
@KevinGDialpad looks like `pylint` had a version mismatch issue -- fixed it on master, so a rebase should do the trick for ya.

datastore/gcloud/aio/datastore/value.py
```diff
@@ -0,0 +1,63 @@
...
+    # TODO timestampValue
+    # TODO keyValue
+    str: 'stringValue',
+    # TODO blobValue
+    # TODO geoPointValue
+    # TODO entityValue
+    # TODO arrayValue
+}
+
+_VALUE_TYPES = {v: k for k, v in _JSON_KEYS.items()}
```
KevinGDialpad: This is not strictly necessary but I find that it makes `from_repr()` easier to read

datastore/gcloud/aio/datastore/constants.py
```diff
@@ -2,6 +2,11 @@
 from datetime import datetime as dt
 
 
+class CompositeFilterOperator(enum.Enum):
+    OPERATOR_UNSPECIFIED = 'OPERATOR_UNSPECIFIED'
```
TheKevJames: nit: the key value for all these unspecified types should just be `UNSPECIFIED`

datastore/gcloud/aio/datastore/filter.py
```diff
@@ -0,0 +1,111 @@
+from typing import Any
+from typing import Dict
+from typing import List
+
+from gcloud.aio.datastore.constants import CompositeFilterOperator
+from gcloud.aio.datastore.constants import PropertyFilterOperator
+from gcloud.aio.datastore.value import Value
+
+
+class _BaseFilter:
```
TheKevJames: Huh, I've never seen abstract base classes with a leading underscore before. We've just used `BaseClassName` everywhere else, so consistency would be great here, but if there's a good reason for the leading underscore, we'd be happy to use this model?

datastore/gcloud/aio/datastore/filter.py
```diff
@@ -0,0 +1,111 @@
...
+    json_key = ''
```
TheKevJames: This can just be `json_key: str` to be abstract rather than giving it the default value.

datastore/tests/unit/filter_test.py
```diff
@@ -0,0 +1,165 @@
...
+        assert repr(query_filter) == str(query_filter.to_repr())
+
+    @staticmethod
+    def _create_property_filter() -> PropertyFilter:
+        return TestFilter._create_property_filters(count=1)[0]
+
+    @staticmethod
+    def _create_property_filters(count: int) -> List[PropertyFilter]:
+        if count > 2:
+            raise NotImplementedError('You may need to implement a way to generate any number of property filters :)')
```
KevinGDialpad: This might look a bit sloppy but it gets the benefits of a [Creation Method](http://xunitpatterns.com/Creation%20Method.html) that works very nicely with the tests above, without writing a method that is more general than I need (YAGNI). The day you actually need that, this message will let you know :)

Also, it just occurred to me that pylint didn't complain about this line being too long. I think it's either because it doesn't check test files or it's related to the fact that its code is raising an exception on my machine.

datastore/gcloud/aio/datastore/filter.py
```diff
@@ -0,0 +1,111 @@
...
+        return str(self.to_repr())
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Filter):
+            return False
+        return self.inner_filter == other.inner_filter
+
+    @classmethod
+    def from_repr(cls, data: Dict[str, Any]) -> 'Filter':
+        (inner_filter_name, inner_filter_data), = data.items()
```
TheKevJames: We generally prefer using early returns a bit more to try to keep the flows clear. I think using key lookups directly rather than doing the clever deserialization would be a bit more readable:

```
    @classmethod
    def from_repr(cls, data: Dict[str, Any]) -> 'Filter':
        if 'compositeFilter' in data:
            return cls(CompositeFilter.from_repr(data['compositeFilter']))
        if 'propertyFilter' in data:
            return cls(PropertyFilter.from_repr(data['propertyFilter']))

        raise ValueError(f'invalid filter name: {data.keys()}')
```

datastore/gcloud/aio/datastore/filter.py
```diff
@@ -0,0 +1,111 @@
...
+            inner_filter = CompositeFilter.from_repr(inner_filter_data)
+        elif inner_filter_name == 'propertyFilter':
+            inner_filter = PropertyFilter.from_repr(inner_filter_data)
+        else:
+            raise ValueError(f'Invalid filter name: {inner_filter_name}')
+        return cls(inner_filter=inner_filter)
+
+    def to_repr(self) -> Dict[str, Any]:
+        return {
+            self.inner_filter.__class__.json_key: self.inner_filter.to_repr()
```
TheKevJames: I don't think you need the class accessor here, `self.inner_filter.json_key` should be enough.

datastore/gcloud/aio/datastore/filter.py
```diff
@@ -0,0 +1,111 @@
...
+    @classmethod
+    def from_repr(cls, data: Dict[str, Any]) -> 'CompositeFilter':
+        operator = CompositeFilterOperator(data['op'])
+        filters = [Filter.from_repr(f) for f in data['filters']]
+        return cls(operator=operator, filters=filters)
+
+    def to_repr(self) -> Dict[str, Any]:
+        return {
+            'op': self.operator.value,
+            'filters': [f.to_repr() for f in self.filters]
```
TheKevJames: nit: please use trailing commas in dictionaries, it makes diffing much nicer later on.

datastore/gcloud/aio/datastore/query.py
```diff
@@ -5,11 +5,60 @@
...
+    def to_repr(self) -> Dict[str, Any]:
+        data = {
+            'kind': [{'name': self.kind}]
+        }
+        if self.query_filter:
+            data['filter'] = self.query_filter.to_repr()
+        return data
+
+
+class GQLQuery(_BaseQuery):
```
TheKevJames: Please add a comment point to `https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/runQuery#GqlQuery`, thanks! We must have missed it in the first pass.

datastore/gcloud/aio/datastore/query.py
```diff
@@ -5,11 +5,60 @@
...
+        raise NotImplementedError
+
+
+# https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/runQuery#Query
+class Query(_BaseQuery):
+    json_key = 'query'
+
+    def __init__(self, kind: str, query_filter: Filter = None):
+        if not kind:
+            raise NotImplementedError('Kindless queries are not supported')
```
TheKevJames: It'd be nice to implement this. If I'm reading the docs correctly, looks like we'd just need to change the `to_repr` method to attach an empty list. Something like:
```
data = {'kind': [{'name': self.kind}] if self.kind else []}
```

datastore/gcloud/aio/datastore/value.py
```diff
@@ -0,0 +1,63 @@
...
+        for json_key, value_type in _VALUE_TYPES.items():
+            if json_key in data:
+                if json_key == 'nullValue':
+                    value = None
+                else:
+                    value = value_type(data[json_key])
+                break
+        else:
+            raise NotImplementedError(
+                f"Dict '{str(data)}' does not contain a supported value key")
```
TheKevJames: It'd be great to be a bit more specific here -- rather than serializing the entire dictionary, we could just print out the actual type which is failing, eg, something like
```
raise NotImplementedError(f'{data.keys()} is not a supported value type')
```

FYI: No need to call `str` within an f-string, that'll be done automatically.

datastore/tests/unit/filter_test.py
TheKevJames: Interesting, I hadn't seen this before! What's the benefit to using these magic methods and property creation methods rather than using the standard test fixture workflow?

For example:
```
@pytest.fixture()
def property_filters() -> List[PropertyFilter]:
    return [
            PropertyFilter(prop='prop1', operator=PropertyFilterOperator.LESS_THAN, value=Value('value1')),
            PropertyFilter(prop='prop2', operator=PropertyFilterOperator.GREATER_THAN, value=Value(1234)),
        ]

@pytest.fixture()
def compose_filter(property_filters) -> CompositeFilter:
            return CompositeFilter(
            operator=CompositeFilterOperator.AND,
            filters=[
                Filter(property_filters[0]),
                Filter(property_filters[1])
            ])
```

which can be used in tests with:

```
def test_filter_to_repr(composite_filter):
    test_filter = Filter(inner_filter=composite_filter)
    r = test_filter.to_repr()
    assert r['compositeFilter'] == test_filter.inner_filter.to_repr()
```

KevinGDialpad: None! I'm not used to pytest because we use unittest in the other projects.
But this syntax looks great, I'll have a look and update the tests.

datastore/gcloud/aio/datastore/filter.py
KevinGDialpad: Ah sorry, didn't see that there was a convention.

It is partly because the class is abstract but more because it's internal to the library. I just wanted it to look wrong if somebody tries to do `filter = _BaseFilter()`. A bit like trying to access `foo._bar()`

But now that you point it out, this isn't actually necessary since it's not exposed to the outside world. I still like the fact that `filter = _BaseFilter()` looks wrong. Anyway, I think it's better to follow the convention so I'll make the change.

As a side note, I read a bit about abstract methods in Python and found out that there is something called [abc](https://docs.python.org/3.7/library/abc.html) that could be useful if you wanted to be strict about that kind of thing. (See also [PEP3119](https://www.python.org/dev/peps/pep-3119/))

datastore/gcloud/aio/datastore/filter.py
KevinGDialpad: Nice!

datastore/gcloud/aio/datastore/filter.py
KevinGDialpad: Ah yes, that's much more readable. I think I was trying too hard to not repeat myself

datastore/gcloud/aio/datastore/filter.py
KevinGDialpad: It is - not sure why I added it

datastore/gcloud/aio/datastore/filter.py
TheKevJames: `abc` is neat, but the thing I think would be most useful about marking off `ABC`s isn't handled by it (ie. preventing instantiation of the base class) :( I suppose it might be worth adding an `__init__()` which raises `NotImplemented` to handle that...

```
>>> import abc
>>> class Foo(abc.ABC):
...   pass
...
>>> x = Foo()
>>> x
<__main__.Foo object at 0x108bc4b38>
```

datastore/gcloud/aio/datastore/value.py
KevinGDialpad: The problem for this specific error message is that we also end up here if the type is missing altogether. That's why I print the whole dict. But I guess printing the dict is not that useful without also printing the list of supported values.

The best way to handle errors is probably to
1.  Count the number of fields and raise an error if it's different than 2, then
2. Validate the field that is not called 'excludeFromIndexes'

There is also a 'meaning' field that Google says 'should only be populated for backwards compatibility', and I'm not sure if it's actually required - still need to experiment when I can run end-to-end tests. But that means I may not know how many fields are expected (could be 2 or 3)

datastore/gcloud/aio/datastore/filter.py
KevinGDialpad: Ah, smart. I prefer to do that in __new__ rather than __init__, but that's a good way to make it explicit. Will make the change.

datastore/gcloud/aio/datastore/value.py
KevinGDialpad: Hey actually, I can just use constants.TYPES! Then I don't need those 2 dicts

datastore/gcloud/aio/datastore/filter.py
TheKevJames: New could make sense for sure, if that seems cleaner!

datastore/gcloud/aio/datastore/value.py
TheKevJames: That might just just point to a difference in the error message: do you think raising
```
raise NotImplementedError(f'{data.keys()} does not contain a supported value type (any of: {_VALUE_TYPES.keys()})')
```
would do the trick?

KevinGDialpad: Yes, I think this is the best. The error message is not that precise but it does give all the information needed

TheKevJames:
Closing in favor of #70

PR #70, opened 2019-04-02
feat(datastore) add non-GQL queries

KevinGDialpad:
Add support for non-GQL queries

I realized after rebasing that it would have been better to add commits to the previous code review and squash at the end, just before merging. Sorry about that. It's not obvious but I addressed your comments from #69 and added a smoke test and a documentation example for the queries.

### DIFF
--- a/datastore/README.rst
+++ b/datastore/README.rst
@@ -19,9 +19,14 @@ started:
 .. code-block:: python
 
     from gcloud.aio.datastore import Datastore
+    from gcloud.aio.datastore import Filter
+    from gcloud.aio.datastore import GQLQuery
     from gcloud.aio.datastore import Key
     from gcloud.aio.datastore import PathElement
-    from gcloud.aio.datastore import GQLQuery
+    from gcloud.aio.datastore import PropertyFilter
+    from gcloud.aio.datastore import PropertyFilterOperator
+    from gcloud.aio.datastore import Query
+    from gcloud.aio.datastore import Value
 
     ds = Datastore('my-gcloud-project', '/path/to/creds.json')
     key1 = Key('my-gcloud-project', [PathElement('Kind', 'entityname')])
--- a/datastore/README.rst
+++ b/datastore/README.rst
@@ -55,9 +55,13 @@ started:
     await ds.reserveIds(allocated_keys)
 
     # query support
-    query = GQLQuery('SELECT * FROM the_meaning_of_life WHERE answer = @answer',
-                     named_bindings={'answer': 42})
+    property_filter = PropertyFilter(prop='answer', operator=PropertyFilterOperator.EQUAL, value=Value(42))
+    query = Query(kind='the_meaning_of_life', query_filter=Filter(property_filter))
     results = await ds.runQuery(query, session=s)
+
+    gql_query = GQLQuery('SELECT * FROM the_meaning_of_life WHERE answer = @answer',
+                        named_bindings={'answer': 42})
+    results = await ds.runQuery(gql_query, session=s)
 
 Custom Subclasses
 ~~~~~~~~~~~~~~~~~
--- a/datastore/gcloud/aio/datastore/filter.py
+++ b/datastore/gcloud/aio/datastore/filter.py
@@ -13,8 +13,8 @@ class BaseFilter:
     # pylint: disable=unused-argument
     def __new__(cls, *args, **kwargs):
         if cls is BaseFilter:
-            raise TypeError('BaseFilter may not be instantiated')
-        return object.__new__(cls)
+            raise NotImplementedError('BaseFilter may not be instantiated')
+        return super().__new__(cls)
 
     def __repr__(self) -> str:
         return str(self.to_repr())
@@ -29,7 +29,7 @@ class BaseFilter:
 
 # https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/runQuery#Filter
 class Filter:
-    def __init__(self, inner_filter: BaseFilter):
+    def __init__(self, inner_filter: BaseFilter) -> None:
         self.inner_filter = inner_filter
 
     def __repr__(self) -> str:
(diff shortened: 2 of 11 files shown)


datastore/tests/integration/smoke_test.py
```diff
@@ -78,6 +83,35 @@
...
+
+        property_filter = PropertyFilter(
+            prop='value', operator=PropertyFilterOperator.EQUAL,
+            value=Value(42))
+        query = Query(kind=kind, query_filter=Filter(property_filter))
+
+        before = await ds.runQuery(query, session=s)
+        num_results = len(before.entity_results)
+
+        transaction = await ds.beginTransaction(session=s)
```
KevinGDialpad: I was really tempted to create a method for the object insertion here, but it ended up taking 5 or 6 parameters and didn't actually look much cleaner

datastore/README.rst
```diff
@@ -55,10 +55,14 @@ started:
     await ds.reserveIds(allocated_keys)
 
     # query support
-    query = GQLQuery('SELECT * FROM the_meaning_of_life WHERE answer = @answer',
-                     named_bindings={'answer': 42})
+    property_filter = PropertyFilter(prop='answer', operator=PropertyFilterOperator.EQUAL, value=Value(42))
```
TheKevJames: Please add the necessary imports to the top of this code block.

datastore/gcloud/aio/datastore/filter.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class BaseFilter:
+    json_key: str
+
+    # pylint: disable=unused-argument
+    def __new__(cls, *args, **kwargs):
+        if cls is BaseFilter:
+            raise TypeError('BaseFilter may not be instantiated')
+        return object.__new__(cls)
```
TheKevJames: Use `super()` rather than `object` -- always good to use the builtins, in case we have `BaseFilter` inherit from something else in the future.

datastore/gcloud/aio/datastore/filter.py
```diff
@@ -0,0 +1,115 @@
...
+from gcloud.aio.datastore.constants import CompositeFilterOperator
+from gcloud.aio.datastore.constants import PropertyFilterOperator
+from gcloud.aio.datastore.value import Value
+
+
+class BaseFilter:
+    json_key: str
+
+    # pylint: disable=unused-argument
+    def __new__(cls, *args, **kwargs):
```
TheKevJames: Looks like `__new__` doesn't take any additional args -- you can remove `args` and `kwargs` here (as well as the pylint disable).

datastore/gcloud/aio/datastore/filter.py
```diff
@@ -0,0 +1,115 @@
...
+        if cls is BaseFilter:
+            raise TypeError('BaseFilter may not be instantiated')
```
TheKevJames: This should be a `NotImplementedError`

datastore/gcloud/aio/datastore/filter.py
```diff
@@ -0,0 +1,115 @@
...
+    def from_repr(cls, data: Dict[str, Any]) -> 'BaseFilter':
+        raise NotImplementedError
+
+    def to_repr(self) -> Dict[str, Any]:
+        raise NotImplementedError
+
+
+# https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/runQuery#Filter
+class Filter:
+    def __init__(self, inner_filter: BaseFilter):
```
TheKevJames: nit: the return type of `__init__` is always `None`.

datastore/gcloud/aio/datastore/filter.py
```diff
@@ -0,0 +1,115 @@
...
+        if 'compositeFilter' in data:
+            return cls(CompositeFilter.from_repr(data['compositeFilter']))
+        if 'propertyFilter' in data:
+            return cls(PropertyFilter.from_repr(data['propertyFilter']))
+
+        raise ValueError(f'invalid filter name: {data.keys()}')
+
+    def to_repr(self) -> Dict[str, Any]:
+        return {
+            self.inner_filter.json_key: self.inner_filter.to_repr()
```
TheKevJames: nit: please include trailing commas in all dictionaries

datastore/gcloud/aio/datastore/query.py
```diff
@@ -5,11 +5,63 @@
...
 from gcloud.aio.datastore.utils import make_value
 from gcloud.aio.datastore.utils import parse_value
 
 
-class GQLQuery:
+class BaseQuery:
+    json_key = ''
+
+    # pylint: disable=unused-argument
+    def __new__(cls, *args, **kwargs):
```
TheKevJames: Same comments as for the other `__new__`

datastore/gcloud/aio/datastore/query.py
```diff
@@ -5,11 +5,63 @@
...
 from gcloud.aio.datastore.constants import ResultType
 from gcloud.aio.datastore.entity import EntityResult
+from gcloud.aio.datastore.filter import Filter
 from gcloud.aio.datastore.utils import make_value
 from gcloud.aio.datastore.utils import parse_value
 
 
-class GQLQuery:
+class BaseQuery:
+    json_key = ''
```
TheKevJames: Just `json_key: str`, like the other ABC.

datastore/gcloud/aio/datastore/filter.py
KevinGDialpad: That's what I did at the beginning, and I got `Unexpected keyword argument 'operator' in constructor call (unexpected-keyword-arg)` on `cls(operator=operator, filters=filters)` in CompositeFilter. And a similar one in PropertyFilter. Pycharm also complains about `__init__`'s signature being incompatible with `__new__`'s

So this is kind of containing the problem... I'm starting to think that this is just not worth it and that if we don't want to do this, we should probably just remove this method. As much as I would like to learn exactly how init and new work in Python, this might not be the best use of my time :)

TheKevJames: Interesting.. yeah, I agree, this is not worth your time right now. Let's just nuke the `__new__` / `__init__` methods on the abstract base classes entirely and rely on the fact that they're not publicly exposed.

PR #71, opened 2019-04-10
feat(datastore) Add ordering, cursors, offset, and limit on queries

KevinGDialpad:
Add ordering on queries, along with the simple types (cursors, offset and limit).

All that is missing now is projections and distinctOn.


datastore/gcloud/aio/datastore/query.py
```diff
@@ -28,9 +29,18 @@ def to_repr(self) -> Dict[str, Any]:
...
 
-    def __init__(self, kind: str = '', query_filter: Filter = None) -> None:
+    # TODO support projection and distinctOn
+    def __init__(self, kind: str = '', query_filter: Filter = None,
+                 order: List[PropertyOrder] = None, start_cursor: str = '',
+                 end_cursor: str = '', offset: int = None, limit: int = None
+                 ) -> None:
         self.kind = kind
         self.query_filter = query_filter
+        self.orders = order or []
```
kylepad: I could be missing something but it seems like a user could create two colliding `PropertyOrder` objects.  As in two copies with the same property, but two different directions. 
Would it be worth performing some sort of validation here to prevent that?

KevinGDialpad: I've tried and this is actually allowed - it sorts in whichever order comes first.

But even if it didn't, I think I would still want the bad request to reach Google and let Google create an error and the client deal with it, no? This way gcloud-aio-datastore wouldn't make an assumption on what is or isn't allowed in the JSON request.

kylepad: 🤔 yeah that makes sense to me 👍

PR #72, opened 2019-05-03
feat(datastore) Support entityValue and keyValue

KevinGDialpad:
Add support for entityValue and keyValue.

Works for the default Key and Entity classes, as well as for custom classes defined in a subclass of Datastore.


datastore/gcloud/aio/datastore/constants.py
```diff
@@ -61,23 +61,14 @@ class TypeName(enum.Enum):
...
     DOUBLE = 'doubleValue'
+    ENTITY = 'entityValue'
     INTEGER = 'integerValue'
+    KEY = 'keyValue'
     NULL = 'nullValue'
     STRING = 'stringValue'
     TIMESTAMP = 'timestampValue'
 
 
-# TODO: support more than just scalars
```
KevinGDialpad: I moved this to utils since some items are added dynamically now

datastore/gcloud/aio/datastore/utils.py
```diff
@@ -1,18 +1,35 @@
...
+
+
 def infer_type(value: Any) -> TypeName:
     kind = type(value)
 
     try:
         return TYPES[kind]
     except KeyError:
+        for supported_type, name in TYPES.items():
+            # Subclasses of supported entity types are also supported
```
KevinGDialpad: It's acceptable to have a custom entity and manipulate data (e.g. run queries) using subclasses of that entity

datastore/tests/unit/datastore_test.py
```diff
@@ -1,5 +1,23 @@
-import gcloud.aio.datastore.datastore as datastore  # pylint: disable=unused-import
+import gcloud.aio.datastore.datastore as datastore
+from gcloud.aio.datastore.constants import TypeName
+from gcloud.aio.datastore.utils import TYPES
 
 
-def test_importable():
-    assert True
+class TestDatastore:
```
jonathan-johnston: Do we need to make a class for this? With `pytest` you can just use a function to represent the test you want to run.

datastore/gcloud/aio/datastore/utils.py
```diff
@@ -1,18 +1,35 @@
...
+}
+
+
 def infer_type(value: Any) -> TypeName:
     kind = type(value)
 
     try:
         return TYPES[kind]
     except KeyError:
+        for supported_type, name in TYPES.items():
```
jonathan-johnston: If this is running concurrently against the global `TYPES` and the dictionary changes size after, for example, `TYPES[value] = TypeName.KEY` on a newly created subclass type, does this loop fail?

datastore/gcloud/aio/datastore/datastore.py
```diff
@@ -37,7 +39,36 @@
...
+class DatastoreMeta(type):
+    _entity_result_kind = None
+    _key_kind = None
+
+    @property
+    def entity_result_kind(cls):
+        return cls._entity_result_kind
+
+    @entity_result_kind.setter
+    def entity_result_kind(cls, value):
```
jonathan-johnston: Functions added should have type hints so `mypy` can validate in CI.

KevinGDialpad: Yes. This one, in particular, is difficult because `value` is a type that contains an attribute named `entity_kind` that is also a type.

datastore/gcloud/aio/datastore/utils.py
KevinGDialpad: What do you mean by fail?

That the user would be running something that calls this method while updating `entity_result_kind` in Datastore and the would be a race condition and the possibility that the new entity kind would not appear in this loop because it has been updated too late?

I'm not sure what we could do about that.

If your concern is what happens to `items()` if the list is updated externally, then I'm not sure but I'd be surprised if it raised an exception.

datastore/tests/unit/datastore_test.py
KevinGDialpad: That's true but that's not what we do in the other test files, so I thought I'd add a class here too. There may be more tests in this file in the future.

datastore/gcloud/aio/datastore/utils.py
TheKevJames: It would indeed raise an exception: `RuntimeError: dictionary changed size during iteration` in particular. When writing threadsafe code, we absolutely need to avoid modifying globals like this -- you'll need to have this `TYPES` value passed in or stored somewhere threadsafe, instead.

datastore/gcloud/aio/datastore/constants.py
TheKevJames: See other comment -- you'll probably need to keep this as a comment and have the dynamic modifications stored separately.

datastore/gcloud/aio/datastore/utils.py
KevinGDialpad: Ahhh... My bad.

Well, I don't actually need to support entityValue so I can just remove that and add it later in a different PR. If I remove those lines and the entity_result_kind in DatastoreMeta, then TYPES will be constant, except that the key_kind setter will add the key type. Would that be acceptable?

I think that there is a lot more complexity in adding entityValue, so I shouldn't add it before I need it.

TheKevJames: I agree wholeheartedly with that -- YAGNI and all that!

I think overall that sounds reasonable? So long as we're avoiding the global multi-thread mutations, we should be good here.

datastore/tests/unit/datastore_test.py
TheKevJames: @KevinGDialpad yeah, I'd been meaning to change that -- we prefer test functions over test classes, unless there's some reason why the test class would be necessary for this case (which I don't think is the case here). Please nuke the class!

datastore/gcloud/aio/datastore/utils.py
```diff
@@ -1,12 +1,25 @@
...
 from typing import Dict
 
 from gcloud.aio.datastore.constants import FORMATTERS
 from gcloud.aio.datastore.constants import TypeName
-from gcloud.aio.datastore.constants import TYPES
 from gcloud.aio.datastore.constants import UNFORMATTERS
 
 
+# TODO: add types for entityValue, geoPointValue and arrayValue
+# NOTE: TypeName.KEY is added dynamically by the datastore
```
TheKevJames: I think I'm missing the reason behind all this metaprogramming and adding the typename for this constant. Let me know where I'm misunderstanding this requirement:
- we want to support `Value.from_repr()` and `Value.to_repr()` for `keyValue` types

Is this not just plumbing? ie. in `Value.from_repr()` we'd check if the entity is a `keyValue` type, and if so we'd set: `self.value = Key.from_repr(...)`. Then, in `Value.to_repr()` we'd emit `self.value.to_repr()` if it is a subclass/instanceof `Key`. I don't think we need to modify the `TYPES` dictionary dynamically at all, we just need to add `Key: TypeName.KEY` to it.

datastore/gcloud/aio/datastore/utils.py
TheKevJames: Quick note that I think given [this comment](https://github.com/talkiq/gcloud-aio/pull/72#discussion_r281422024), adding `entityValue` support would just be roughly the one-liner of adding `Entity: TypeName.ENTITY` to the `TYPES` dict and changing the `if .. == 'keyValue'` checks to `if .. in {'keyValue', 'entityValue'}`.

datastore/gcloud/aio/datastore/utils.py
KevinGDialpad: The problem is that you can subclass the datastore and have a key type that's completely unrelated to Key.

I can do something like this:
```
from gcloud.aio.datastore import Datastore

class MyKey:
  pass

class MyDatastore(Datastore):
  key_kind = MyKey
```
But when the client tries to run a query with a filter on a value of type MyKey, it fails inside `Value.to_repr()` because MyKey is an unknown type.

That's why I need Value to know about the datastore's key_kind, which has been kind of a headache. But I like this implementation - the TYPES dict is basically known at the creation time of your custom datastore.

datastore/tests/unit/datastore_test.py
KevinGDialpad: I see. I'll remove this one then!

datastore/gcloud/aio/datastore/utils.py
KevinGDialpad: That would support Entity but not custom entity types. But you're right, that part is very easy to do so I can just add it. Custom entity types can be supported later.

TheKevJames: Gotcha, did not think of the child classes case. Good call, we can hold off on that for now!

datastore/gcloud/aio/datastore/utils.py
TheKevJames: Makes sense! Let's try to solve it the same way as we've done elsewhere -- by adding `key_kind` (and `entity_kind`, if we wanted to support both`) as class attributes to `Value`, like we do for all the other classes.

Then, within and `Value` methods, we'd always reference `self.key_kind` instead of `Key`. The `Entity` and `EntityResult` classes have good examples of how to do this.

KevinGDialpad: I like the idea.
I would still like infer_type, make_value and parse_value to work with `key_kind` though. At least make_value, since that would add support for `key_kind` to GQL queries.

The problem is that there is some duplication between Value's form_repr/to_repr and make_value/perse_value. We can probably find a way to use only one of them.

datastore/gcloud/aio/datastore/utils.py
```diff
@@ -1,29 +0,0 @@
...
-def infer_type(value: Any) -> TypeName:
-    kind = type(value)
-
-    try:
-        return TYPES[kind]
-    except KeyError:
-        raise Exception(f'unsupported value type {kind}')
-
-
-def make_value(value: Any) -> Dict[str, Any]:
```
KevinGDialpad: So actually, `make_value()` and `parse_value()` predate the Value class. They were used to write/parse Values. Now that there is a Value class, I moved that logic there.

datastore/gcloud/aio/datastore/constants.py
```diff
@@ -61,29 +61,21 @@ class TypeName(enum.Enum):
...
     dt: TypeName.TIMESTAMP,
     float: TypeName.DOUBLE,
     int: TypeName.INTEGER,
     str: TypeName.STRING,
-    bool: TypeName.BOOLEAN,
     type(None): TypeName.NULL,
 }
-
-FORMATTERS = {
-    TypeName.TIMESTAMP: lambda d: d.strftime('%Y-%m-%dT%H:%S:%M.%f000Z'),
```
KevinGDialpad: Pretty sure the `:%S:%M` was a typo, so I swapped them when I moved this line. Same for line 88.

datastore/gcloud/aio/datastore/value.py
```diff
@@ -24,33 +28,58 @@ def __repr__(self) -> str:
...
         return cls(value=value, exclude_from_indexes=exclude_from_indexes)
 
     def to_repr(self) -> Dict[str, Any]:
-        value_type = type(self.value)
-        if value_type not in TYPES:
-            raise NotImplementedError(f'type "{value_type}" is not supported '
-                                      'by the Value type')
-
+        value_type = self._infer_type(self.value)
+        if value_type == TypeName.KEY:
```
KevinGDialpad: I considered keeping the FORMATTERS dict. I don't really mind either way. This is not very pretty but I find it more readable.

datastore/gcloud/aio/datastore/value.py
```diff
@@ -2,11 +2,15 @@
...
 from typing import Dict
 
+from gcloud.aio.datastore.constants import TypeName
 from gcloud.aio.datastore.constants import TYPES
+from gcloud.aio.datastore.key import Key
 
 
 # https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/runQuery#value
 class Value:
+    key_kind = Key
```
KevinGDialpad: We could add `enitty_kind` here.
There is a circular dependency between Value and Entity so I decided to deal with that when I have to and not before :)

datastore/tests/unit/datastore_test.py
```diff
@@ -1,5 +1,15 @@
...
+def test_defining_custom_subclass_updates_value_key_kind():
+    class CustomKey:
+        pass
+
+    class CustomDatastore(datastore.Datastore):  # pylint: disable=unused-variable
+        key_kind = CustomKey
+
+    assert Value.key_kind == CustomKey
+
+    Value.key_kind = Key  # Teardown
```
KevinGDialpad: The fact that this line is necessary highlights the fact that you can shoot yourself in the foot if you're using more than one Datastore class since `Value.key_kind` will take the `key_kind` of whichever was defined last (even if it's not used at all).

Without this line, `Value.key_kind` stays set to `CustomKey` for whatever tests you are running next. So if you run the Value tests after this one, some of them will fail.

I think it's an edge case that demands a lot of attention so my preference would be to open a ticket to take care of that later. I expect the normal workflow to define your own Datastore class (or use the default one) and set `key_kind` there once and for all.

TheKevJames: Yes, that's a great thought to call out -- I agree this is an important edge case. Could you please open a GitHub issue on this repo so we don't forget to look into this?

datastore/gcloud/aio/datastore/constants.py
```diff
@@ -61,29 +61,21 @@ class TypeName(enum.Enum):
     BLOB = 'blobValue'
     BOOLEAN = 'booleanValue'
     DOUBLE = 'doubleValue'
+    ENTITY = 'entityValue'
```
TheKevJames: Let's not add this until we add full support for this.

datastore/gcloud/aio/datastore/datastore.py
```diff
@@ -37,7 +38,27 @@
 log = logging.getLogger(__name__)
 
 
-class Datastore:
+class DatastoreMeta(type):
```
TheKevJames: I think rather than doing this metaclass stuff we should stick with the solution we already have for now -- though this might be a better way to do it going forwards, I'd like to evaluate whether this is a worthwhile change on its own merits rather than at the same time as reviewing the `keyValue` logic. Note that this solution has some problems: it modifies the literal un-instantiated `Value` class for everyone, so as you called out that can cause several edge-cases.

Instead of doing this metaclass, let's give `Datastore`, `Entity`, and `Query` a `value_kind` entity to allow for custom subclassing. Should be the same as we have for all the other custom-subclass-able items; docs are [here](https://github.com/talkiq/gcloud-aio/tree/master/datastore#custom-subclasses). (and it would be great if you could update those docs with your changes!)

datastore/gcloud/aio/datastore/value.py
TheKevJames: Overall, I like this format much better. Thanks for clearing out some of our tech debt!

datastore/gcloud/aio/datastore/datastore.py
KevinGDialpad: Fair enough.
It's going to be a bit odd because it requires the client to specify a `value_kind` when they have a custom Key type. That also makes it a breaking change.

datastore/tests/unit/datastore_test.py
KevinGDialpad: Okay, I'll see what this edge case looks like when I've made the other changes, and create an issue

datastore/README.rst
```diff
@@ -120,10 +120,46 @@ The full list of classes which may be overridden in this way is:
...
+        value_kind = MyValue
+
+    class MyEntityResult(gcloud.aio.datastore.EntityResult):
+        entity_kind = MyEntity
+
+    class MyQueryResultBatch(gcloud.aio.datastore.QueryResultBatch):
+        entity_result_kind = MyEntityResult
+
+    class MyDatastore(gcloud.aio.datastore.Datastore):
+        key_kind = MyKey
```
KevinGDialpad: This is starting to look a bit messy and that could probably be cleaned up.

For example, in `MyEntity`, it shouldn't be possible to change `value_kind` without changing `key_kind` because things will break. So we should be able to remove `Entity.key_kind` completely and rely on `value_kind.key_kind` inside that class.
It's probably true for some combinations of `Datastore`'s attributes as well. It would require some careful thinking before removing some of them but I'm sure it's possible.

But that's not backward-compatible.

datastore/gcloud/aio/datastore/value.py
```diff
@@ -24,33 +28,60 @@ def __repr__(self) -> str:
...
+            supported = [name.value for name in supported_types.values()]
             raise NotImplementedError(
                 f'{data.keys()} does not contain a supported value type'
-                f' (any of: {supported_types})')
+                f' (any of: {supported})')
 
-        exclude_from_indexes = bool(data['excludeFromIndexes'])
+        # Google may not populate that field. This can happen with both
+        # indexed and non-indexed fields.
+        exclude_from_indexes = bool(data.get('excludeFromIndexes', False))
```
KevinGDialpad: There's more than one solution here. I don't think it matters much when parsing, so defaulting to False seems sensible.

TheKevJames:
@KevinGDialpad heads up that I had to create #74 to add some cleanups/docs to this branch before merging. Closing this PR, but your changes are on master. :)

PR #73, opened 2019-05-09
Auth sign_blob

nickfirespotter:
Note: The smoke test for sign_blob hasn't been fully tested

### DIFF
--- a/auth/gcloud/aio/auth/auth.py
+++ b/auth/gcloud/aio/auth/auth.py
@@ -11,13 +11,14 @@ from typing import Any
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import Union
 from urllib.parse import quote_plus
 from urllib.parse import urlencode
 
 import aiohttp
 import backoff
 import jwt
-from gcloud.aio.auth.utils import valid_base64
+from gcloud.aio.auth.utils import encode
 
 
 GCE_METADATA_BASE = 'http://metadata.google.internal/computeMetadata/v1'
@@ -197,15 +198,21 @@ class Token:
 SCOPES = ['https://www.googleapis.com/auth/iam']
 
 
-class IamCredentialsClient:
+class IamClient:
     def __init__(self, service_file: Optional[str] = None,
                  session: Optional[aiohttp.ClientSession] = None,
                  token: Optional[Token] = None) -> None:
-        self.session = session
+        self._session = session
         self.token = token or Token(service_file=service_file, session=session, scopes=SCOPES)
 
         if self.token.token_type != Type.SERVICE_ACCOUNT:
             raise TypeError('IAM Credentials Client is only valid for use with Service Accounts.')
+
+    @property
+    def session(self):
+        if not self._session:
+            self._session = aiohttp.ClientSession(conn_timeout=10, read_timeout=10)
+        return self._session
 
     async def headers(self):
         token = await self.token.get()
@@ -216,35 +223,72 @@ class IamCredentialsClient:
     def service_account_email(self):
         return self.token.service_data.get('client_email')
 
-    async def sign_blob(self, payload: str, service_account_email: Optional[str] = None,
+    async def get_public_key(self, key_id: Optional[str] = None, key: Optional[str] = None,
+                             service_account_email: Optional[str] = None,
+                             project: Optional[str] = None,
+                             session: aiohttp.ClientSession = None,
+                             timeout: int = 10) -> Dict[str, str]:
+        service_account_email = service_account_email or self.service_account_email()
+        project = project or await self.token.get_project()
+
+        if not key_id and not key:
+            raise ValueError('Must have either key_id or key')
+
+        if not key:
+            key = f'projects/{project}/serviceAccounts/{service_account_email}/keys/{key_id}'
+
+        url = (f'https://iam.googleapis.com/v1/{key}?publicKeyType=TYPE_X509_PEM_FILE')
+        headers = await self.headers()
+
+        session = session or self.session
+
+        resp = await session.get(url, headers=headers, timeout=timeout)
+        resp.raise_for_status()
+        return await resp.json()
+
+    async def get_public_key_names(self, service_account_email: Optional[str] = None,
+                                   project: Optional[str] = None,
+                                   session: aiohttp.ClientSession = None,
+                                   timeout: int = 10) -> List[Dict[str, str]]:
+        service_account_email = service_account_email or self.service_account_email()
+        project = project or await self.token.get_project()
+
+        url = (f'https://iam.googleapis.com/v1/projects/{project}/'
+               f'serviceAccounts/{service_account_email}/keys')
+
+        headers = await self.headers()
+        session = session or self.session
+        resp = await session.get(url, headers=headers, timeout=timeout)
+        resp.raise_for_status()
+        return (await resp.json()).get('keys', [])
+
+    # https://cloud.google.com/iam/credentials/reference/rest/v1/projects.serviceAccounts/signBlob
+    async def sign_blob(self, payload: Optional[Union[str, bytes]],
+                        service_account_email: Optional[str] = None,
                         delegates: Optional[list] = None,
                         session: aiohttp.ClientSession = None,
-                        timeout: int = 10) -> None:
-        if not valid_base64(payload):
-            raise TypeError('payload must be base64 encoded.')
+                        timeout: int = 10) -> Dict[str, str]:
+        payload = encode(payload)
 
         service_account_email = service_account_email or self.service_account_email()
 
         if not service_account_email:
-            raise ValueError('sign_blob must have a valid service_account_email')
+            raise TypeError('sign_blob must have a valid service_account_email')
 
         resource_name = f'projects/-/serviceAccounts/{service_account_email}'
         delegates = delegates or [resource_name]
 
         url = f'https://iamcredentials.googleapis.com/v1/{resource_name}:signBlob'
 
-        json_str = json.dumps({'delegates': delegates, 'payload': payload})
+        json_str = json.dumps({'delegates': delegates, 'payload': payload.decode('utf-8')})
+
         headers = await self.headers()
         headers.update({
             'Content-Length': str(len(json_str)),
             'Content-Type': 'application/json'
         })
 
-        if not self.session:
-            self.session = aiohttp.ClientSession(conn_timeout=10,
-                                                 read_timeout=10)
         session = session or self.session
-        resp = await session.post(url, data=json_str, headers=headers,
-                                  timeout=timeout)
+        resp = await session.post(url, data=json_str, headers=headers, timeout=timeout)
         resp.raise_for_status()
-        return (await resp.json()).get('signedBlob')
+        return await resp.json()
--- a/auth/gcloud/aio/auth/auth.py
+++ b/auth/gcloud/aio/auth/auth.py
@@ -247,4 +247,4 @@ class IamCredentialsClient:
         resp = await session.post(url, data=json_str, headers=headers,
                                   timeout=timeout)
         resp.raise_for_status()
-        return await resp.json().get('signedBlob')
+        return (await resp.json()).get('signedBlob')
--- a/auth/gcloud/aio/auth/auth.py
+++ b/auth/gcloud/aio/auth/auth.py
@@ -17,6 +17,7 @@ from urllib.parse import urlencode
 import aiohttp
 import backoff
 import jwt
+from gcloud.aio.auth.utils import valid_base64
 
 
 GCE_METADATA_BASE = 'http://metadata.google.internal/computeMetadata/v1'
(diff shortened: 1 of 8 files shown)


auth/gcloud/aio/auth/auth.py
```diff
@@ -191,3 +192,60 @@ def __init__(self, service_file: Optional[str] = None,
...
+    async def headers(self):
+        token = await self.token.get()
+        return {
+            'Authorization': f'Bearer {token}',
+        }
+
+    def service_account_email(self):
+        return self.token.service_data.get('client_email')
+
+    async def sign_blob(self, payload: str, service_account_email: Optional[str],
```
TheKevJames: Given you've marked `service_account_email` as `Optional`, should we give it a default `= None`?

auth/tests/unit/utils_test.py
```diff
@@ -0,0 +1,15 @@
+import gcloud.aio.auth.utils as utils  # pylint: disable=unused-import
+
+
+def test_importable():
```
TheKevJames: Feel free to nuke this importable test -- we've just been using that as a placeholder for files without actual tests.

auth/gcloud/aio/auth/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    encoded = base64.b64encode(payload)
+    return encoded.replace(b'+', b'-').replace(b'/', b'_').decode('utf-8')
+
+
+def valid_base64(data: str) -> bool:
+    """
+    Check that the data is made up of base64 valid chars.
+
+    """
+    return not bool(INVALID_BASE64_CHARS.search(data))
```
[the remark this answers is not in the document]
nickfirespotter: Since the sign_blob isn't doing the base64 encoded/decoding we really just need to make sure that the payload is valid, not that it is actually encded with the encode util function. So thats why I just checked that the chars were valid rather than making sure it was base64 encoded.

auth/gcloud/aio/auth/auth.py
```diff
@@ -191,3 +192,60 @@ def __init__(self, service_file: Optional[str] = None,
...
+            raise ValueError('sign_blob must have a valid service_account_email')
+
+        resource_name = f'projects/-/serviceAccounts/{service_account_email}'
+        delegates = delegates or [resource_name]
+
+        url = f'https://iamcredentials.googleapis.com/v1/{resource_name}:signBlob'
+
+        request = json.dumps({'delegates': delegates, 'payload': payload})
+
+        json_str = json.dumps(request)
```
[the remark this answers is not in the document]
nickfirespotter: good catch, looks like an artifact from the original common class I defined. Updating.

auth/gcloud/aio/auth/auth.py
```diff
@@ -191,3 +192,60 @@ def __init__(self, service_file: Optional[str] = None,
...
+
+
+class IamCredentialsClient:
+    def __init__(self, project: Optional[str] = None,
+                 service_file: Optional[str] = None,
+                 session: Optional[aiohttp.ClientSession] = None,
+                 token: Optional[Token] = None) -> None:
+        self.project = project
+        self.session = session
+        self.token = token or Token(service_file=service_file, session=session, scopes=SCOPES)
```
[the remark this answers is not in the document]
nickfirespotter: Sure, seems like a good safety check. I will implement.

auth/gcloud/aio/auth/auth.py
```diff
@@ -191,3 +192,60 @@ def __init__(self, service_file: Optional[str] = None,
...
+
+SCOPES = ['https://www.googleapis.com/auth/cloud-platform']
+
+
+class IamCredentialsClient:
+    def __init__(self, project: Optional[str] = None,
+                 service_file: Optional[str] = None,
+                 session: Optional[aiohttp.ClientSession] = None,
+                 token: Optional[Token] = None) -> None:
+        self.project = project
```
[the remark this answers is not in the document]
nickfirespotter: hmm, yeah I guess so. I will remove. I thought it was needed by the token but I guess not.

auth/gcloud/aio/auth/auth.py
```diff
@@ -191,3 +192,60 @@ def __init__(self, service_file: Optional[str] = None,
         self.access_token_duration = int(content['expires_in'])
         self.access_token_acquired_at = datetime.datetime.now()
         self.acquiring = None
+
+
+SCOPES = ['https://www.googleapis.com/auth/cloud-platform']
```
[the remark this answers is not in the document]
nickfirespotter: I swear the documentation said this was the scope required.  but it seems iam will work. Updating

auth/gcloud/aio/auth/auth.py
```diff
@@ -191,3 +192,60 @@ def __init__(self, service_file: Optional[str] = None,
...
+        self.token = token or Token(service_file=service_file, session=session, scopes=SCOPES)
+
+    async def headers(self):
+        token = await self.token.get()
+        return {
+            'Authorization': f'Bearer {token}',
+        }
+
+    def service_account_email(self):
+        return self.token.service_data.get('client_email')
```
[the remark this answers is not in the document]
nickfirespotter: its never null, it just might be an empty dict. If it is then the client_email will be None and we have a check for that. So I think we are covered.

auth/gcloud/aio/auth/auth.py
nickfirespotter: Done

auth/tests/unit/utils_test.py
nickfirespotter: Will do

auth/gcloud/aio/auth/auth.py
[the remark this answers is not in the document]
TheKevJames: Looks like the docs say either will work (which makes sense, since `cloud-platform` is the union of all other permissionsets).

auth/gcloud/aio/auth/auth.py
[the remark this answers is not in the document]
TheKevJames: Ah, good call, I missed that we'd changed to returning empty dicts. Betcha it was for exactly this reason!

auth/tests/integration/smoke_test.py
```diff
@@ -43,3 +45,16 @@
...
+@pytest.mark.asyncio  # type: ignore
+async def test_sign_blob(creds: str) -> None:
+    data = 'Testing Can be confidential!'
+    payload = encode(data)
+
+    async with aiohttp.ClientSession(conn_timeout=10, read_timeout=10) as s:
+
+        iam_credentials = IamCredentialsClient(service_file=creds, session=s)
+        signed_data = iam_credentials.sign_blob(payload)
+        assert signed_data  # TODO(nick): How do we verify data is signed?
```
TheKevJames: Any ideas on this? I don't see any reversible API call we could check with.

nickfirespotter: My best thought is to have the string presigned and checked in as test data. We could pre-sign it using the gcloud command line tool and check in the value and then confirm its the same?
So dont see anything about how to verify the data is signed by your service account in the docs.

TheKevJames: That seems reasonable? I ran your tests on CI, here's the signed blob:
```
Ye7i1BxgqYnHBC9mZxXU+uV0/r0zH4mvm7qWW166mmixSC00vC4qfrEnwipYG4QYA7A52KtBQcOWe9SnCYyw/shqlz9QJH8TW5I+f7/WaE+09GUBinp7bvHpolQ/G9hyYS0a4mUN4Yzh9UqsMfKbi/9chWKVrX8XlWrDxORv2a1F7M5rvr971AJ5jZDlmAVK5FHk48O3xBRGgIle36v+HgS7NLfoAMfyVrjYl1LFByd/uR+1GmGwTBi0LttpGfWm70TXxwUcijZ8BZcxkJGqtIslqceEukP1PVztAR8g8LTO6yfqoI5/tKlsffW5b7OdpJGQEYFv69OVHflvCFKPgQ==
```

auth/gcloud/aio/auth/auth.py
```diff
@@ -191,3 +192,59 @@ def __init__(self, service_file: Optional[str] = None,
...
+        })
+
+        if not self.session:
+            self.session = aiohttp.ClientSession(conn_timeout=10,
+                                                 read_timeout=10)
+        session = session or self.session
+        resp = await session.post(url, data=json_str, headers=headers,
+                                  timeout=timeout)
+        resp.raise_for_status()
+        return await resp.json().get('signedBlob')
```
TheKevJames: This should be `return (await resp.json()).get('signedBlob')` -- otherwise, this will try to call `.get` on the coroutine and `await` the result.

auth/tests/integration/smoke_test.py
```diff
@@ -43,3 +45,16 @@
...
+
+@pytest.mark.asyncio  # type: ignore
+async def test_sign_blob(creds: str) -> None:
+    data = 'Testing Can be confidential!'
+    payload = encode(data)
+
+    async with aiohttp.ClientSession(conn_timeout=10, read_timeout=10) as s:
+
+        iam_credentials = IamCredentialsClient(service_file=creds, session=s)
+        signed_data = iam_credentials.sign_blob(payload)
```
TheKevJames: This should be `signed_data = await iam_credentials...`

auth/tests/integration/smoke_test.py
TheKevJames: Well, either I did not reproduce the results correctly in the below command (@nickfirespotter could you please verify?) or the results change based on time. I'll re-run the below tomorrow to see if the later is the case.

```
$ gcloud iam service-accounts sign-blob \
    --iam-account=public-testing@voiceai-staging.iam.gserviceaccount.com \
    <(python3 -c "import base64; print(base64.b64encode('Testing Can be confidential!'.encode('utf-8')).replace(b'+', b'-').replace(b'/', b'_').decode('utf-8'), end='')") \
    ./foo  2>/dev/null
$ python3 -c "import base64; print(base64.b64encode(open('./foo', 'rb').read()).replace(b'+', b'-').replace(b'/', b'_').decode('utf-8'))"
FnmTO-oK-joLmPUgFgvSaSeBmabWen38W70-LlTSfCi1BfcKh24jTtLYnvTLR6qCcet7qyiFtwXoGHYbFgesWsA327NiO-X3EvXbakf9rW_c5wheQMFV4Hf0lmRJOYF7o7jGCgBY47DHlEo531stIXn80U_9zmnHcmEZn9o9lUKb7WccIvDLUS-gPcGCBL7UVjtrMtuM6sXOPe1dhhdKtLQhfar4oC4paCVWx4ksFIiPkAXyV5_uvrKplS2PC2Y5JVKUdRjBAq9Q3Cf_BWOMmpo5L93dGuPVM5bWmRjH6OPanLPr5JQg-j5MrRB-ENfMD8zwrhPwbN-eBRrFTjPbbQ==
```

auth/gcloud/aio/auth/auth.py
nickfirespotter: i wonder how this worked so far.... changing.

auth/tests/integration/smoke_test.py
nickfirespotter: done

auth/tests/integration/smoke_test.py
TheKevJames: The returned blob is the same -- but it still doesn't match the signed blob the tests output.

TheKevJames: Removing the bytes fuckery, I get:
```
$ gcloud iam service-accounts sign-blob \
    --iam-account=public-testing@voiceai-staging.iam.gserviceaccount.com \
    <(python3 -c "import base64; print(base64.b64encode(b'Testing Can be confidential!').replace(b'+', b'-').replace(b'/', b'_'), end='')") \
    ./foo  2>/dev/null
$ python3 -c "import base64; print(base64.b64encode(open('./foo', 'rb').read()).replace(b'+', b'-').replace(b'/', b'_'))"
b'JaZ9SMwrNJN2dE9WsTn8ds6isQwzCRngmWUrEu8-m8fTt0-zOyArUE1cc1pS_tyMgcx_MKd3eBbLHbKZ4gF5RAMWczujX--YlO09e4Ot_UM_1WDO3bwA0iKlg_14jYEay6i2gJ7zOS4XfoCG4aotVyshzKoLRpMB98wbFp1AZxZo35q7UZH98Z4mtU5BLV40RVk7lGhZdvgXEgvbECdEMXSN1fI4kayKS7ks_KLJ4liPugJpMQDwvf7YJkx8xuvtZBc0exo63bFin8NDA-PMk6XNF_7OoEiuubr48I-7UItqOjXrqmAPnngXTDAt9AiYDG88_4bjP2yFLta1DIdhbA=='
```

TheKevJames:
[Merged](https://github.com/talkiq/gcloud-aio/commit/713704fbbb34fabc0215bdb70891a385f5864765)

PR #76, opened 2019-05-21
Datastore export

nickfirespotter:
Hook up the Datastore export API into gcloud.aio.datastore

### DIFF
--- a/datastore/gcloud/aio/datastore/datastore.py
+++ b/datastore/gcloud/aio/datastore/datastore.py
@@ -11,6 +11,7 @@ from gcloud.aio.auth import Token  # pylint: disable=no-name-in-module
 from gcloud.aio.datastore.constants import Consistency
 from gcloud.aio.datastore.constants import Mode
 from gcloud.aio.datastore.constants import Operation
+from gcloud.aio.datastore.datastore_operation import DatastoreOperation
 from gcloud.aio.datastore.entity import EntityResult
 from gcloud.aio.datastore.key import Key
 from gcloud.aio.datastore.query import BaseQuery
@@ -348,18 +349,20 @@ class Datastore:
         mutation = self.make_mutation(operation, key, properties=properties)
         await self.commit([mutation], transaction=transaction, session=session)
 
-    async def export(self, output_bucket_name: str,
-                     kinds: Optional[List[str]] = None, namespaces: Optional[List[str]] = None,
+    async def export(self, output_bucket_prefix: str,
+                     kinds: Optional[List[str]] = None,
+                     namespaces: Optional[List[str]] = None,
                      labels: Optional[Dict[str, str]]= None,
-                     session: aiohttp.ClientSession = None, timeout: int = 10) -> Dict[str, str]:
+                     session: aiohttp.ClientSession = None,
+                     timeout: int = 10) -> DatastoreOperation:
         project = await self.project()
         url = f'{API_ROOT}/projects/{project}:export'
 
         export_request = {
-            'entityFilter': {'kinds': kinds or [], 'namespaceIds': namespaces or []},
-            'outputUrlPrefix': f'gs://{output_bucket_name}'}
-        if labels:
-            export_request['labels'] = labels
+            'entityFilter': {'kinds': kinds or [],
+                             'namespaceIds': namespaces or []},
+            'outputUrlPrefix': f'gs://{output_bucket_prefix}',
+            'labels': labels or {}}
 
         payload = json.dumps(export_request).encode('utf-8')
 
@@ -373,13 +376,15 @@ class Datastore:
             self.session = aiohttp.ClientSession(conn_timeout=10,
                                                  read_timeout=10)
         session = session or self.session
-        resp = await session.post(url, data=payload, headers=headers, timeout=timeout)
+        resp = await session.post(url, data=payload, headers=headers,
+                                  timeout=timeout)
         resp.raise_for_status()
 
-        return await resp.json()
+        return DatastoreOperation.from_repr(await resp.json())
 
-    async def get_operation(self, name: str, session: aiohttp.ClientSession = None,
-                            timeout: int = 10) -> Dict[str, str]:
+    async def get_datastore_operation(self, name: str,
+                                      session: aiohttp.ClientSession = None,
+                                      timeout: int = 10) -> DatastoreOperation:
         url = f'{API_ROOT}/{name}'
 
         headers = await self.headers()
@@ -394,4 +399,4 @@ class Datastore:
         resp = await session.get(url, headers=headers, timeout=timeout)
         resp.raise_for_status()
 
-        return await resp.json()
+        return DatastoreOperation.from_repr(await resp.json())
--- a/datastore/gcloud/aio/datastore/datastore.py
+++ b/datastore/gcloud/aio/datastore/datastore.py
@@ -23,10 +23,10 @@ except ModuleNotFoundError:
 
 
 try:
-    API_ROOT = f'http://{os.environ["DATASTORE_EMULATOR_HOST"]}/v1/projects'
+    API_ROOT = f'http://{os.environ["DATASTORE_EMULATOR_HOST"]}/v1'
     IS_DEV = True
 except KeyError:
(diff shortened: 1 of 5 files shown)


datastore/gcloud/aio/datastore/datastore.py
```diff
@@ -346,3 +346,51 @@ def make_mutation(operation: Operation, key: Key,
...
+                     labels: Optional[Dict[str, str]]= None,
+                     session: aiohttp.ClientSession = None, timeout: int = 10) -> Dict[str, str]:
+        project = await self.project()
+        url = f'{API_ROOT}/projects/{project}:export'
+
+        export_request = {
+            'entityFilter': {'kinds': kinds or [], 'namespaceIds': namespaces or []},
+            'outputUrlPrefix': f'gs://{output_bucket_name}'}
+        if labels:
+            export_request['labels'] = labels
```
TheKevJames: Would it be cleaner to do this the same way as the other values, eg. by unconditionally setting `'labels': labels or {}`, or does that break their API?

datastore/gcloud/aio/datastore/datastore.py
```diff
@@ -346,3 +346,51 @@ def make_mutation(operation: Operation, key: Key,
         transaction = await self.beginTransaction(session=session)
         mutation = self.make_mutation(operation, key, properties=properties)
         await self.commit([mutation], transaction=transaction, session=session)
+
+    async def export(self, output_bucket_name: str,
```
TheKevJames: Let's call this param `output_bucket_prefix`, since it supports arbitrary prefix-based folder nesting.

datastore/gcloud/aio/datastore/datastore.py
```diff
@@ -346,3 +346,51 @@ def make_mutation(operation: Operation, key: Key,
...
+        })
+
+        if not self.session:
+            self.session = aiohttp.ClientSession(conn_timeout=10,
+                                                 read_timeout=10)
+        session = session or self.session
+        resp = await session.post(url, data=payload, headers=headers, timeout=timeout)
+        resp.raise_for_status()
+
+        return await resp.json()
```
TheKevJames: Looks like a bunch of parsing on the user's side to handle the response of this. Should we create an [`Operation` class](https://cloud.google.com/datastore/docs/reference/admin/rest/Shared.Types/Operation) the same way as we have Key, Entity, etc?

datastore/gcloud/aio/datastore/datastore.py
nickfirespotter: updated

datastore/gcloud/aio/datastore/datastore.py
nickfirespotter: updated

datastore/gcloud/aio/datastore/datastore.py
nickfirespotter: added DatastoreOperation class

datastore/gcloud/aio/datastore/datastore_operation.py
```diff
@@ -0,0 +1,28 @@
...
+                 error=None, response: Optional[Dict[str, Any]] = None):
+        self.name = name
+        self.metadata = metadata
+        self.done = done
+        self.error = error
+        self.response = response
+
+    @classmethod
+    def from_repr(cls, data: Dict[str, Any]) -> 'DatastoreOperation':
+        return cls(data.get('name'), data.get('done'), data.get('metadata'),
```
TheKevJames: Let's do direct indexing instead of `get`s for any of the non-optional fields, to make errors more obvious upfront. Looks like that's `name` and `done`, judging by your `__init__` type hints.

datastore/gcloud/aio/datastore/datastore_operation.py
```diff
@@ -0,0 +1,28 @@
+from typing import Any
+from typing import Dict
+from typing import Optional
+
+
+class DatastoreOperation:
+    def __init__(self, name: str, done: bool,
+                 metadata: Optional[Dict[str, Any]] = None,
+                 error=None, response: Optional[Dict[str, Any]] = None):
```
TheKevJames: type hint for `error`, please, and a `-> None` at the end.

datastore/tests/integration/conftest.py
```diff
@@ -16,4 +16,8 @@ def kind() -> str:
 
 @pytest.fixture(scope='module')  # type: ignore
 def project() -> str:
-    return 'voiceai-staging'
+    return 'fststaging'
+
+@pytest.fixture(scope='module')  # type: ignore
+def export_bucket_name() -> str:
+    return os.environ['DATASTORE_EXPORT_BUCKET_NAME']
```
TheKevJames: You can set this to `voiceai-staging-public-test` (or a folder within it)

datastore/tests/integration/smoke_test.py
```diff
@@ -137,3 +141,41 @@
 
         after = await ds.runQuery(query, session=s)
         assert len(after.entity_results) == num_results + 3
+
+
+@pytest.mark.skipif('DATASTORE_EXPORT_BUCKET_NAME' not in os.environ,
```
TheKevJames: Not sure this will work, since the test fixture should throw a `KeyError` if this is unset before it gets to running tests.

datastore/tests/integration/smoke_test.py
```diff
@@ -137,3 +141,41 @@
...
+                    reason='DATASTORE_EXPORT_BUCKET_NAME not set')
+@pytest.mark.asyncio  # type: ignore
+async def test_datastore_export(creds: str, project: str,
+                                export_bucket_name: str):
+    kind = 'DatastoreExportTestModel'
```
TheKevJames: Use the `kind` fixture, please, or create a new fixture, but for now we want everything linked to the "public test" stuff.

datastore/tests/integration/smoke_test.py
```diff
@@ -137,3 +141,41 @@
...
+
+    rand_uuid = str(uuid.uuid4())
+
+    async with aiohttp.ClientSession(conn_timeout=10, read_timeout=10) as s:
+        ds = Datastore(project=project, service_file=creds, session=s)
+
+        mutations = [ds.make_mutation(Operation.INSERT,
+                                      Key(project, [PathElement(kind)]),
+                                      properties={'rand_str': rand_uuid})]
+        await ds.commit(mutations, session=s, mode=Mode.NON_TRANSACTIONAL)
```
TheKevJames: Using the `ds.insert` convenience method might be easier here.

datastore/noxfile.py
```diff
@@ -3,7 +3,7 @@
 import nox
 
 
-LOCAL_DEPS = ('../auth/', )
+LOCAL_DEPS = ('../auth/', '../storage/', )
```
TheKevJames: This is probably going to make independent versioning of `storage` and `datastore` difficult, for example when we've updated one but not the other to use incompatible dependencies. Let's leave this off for now, disable the new test by default, and mark the skip message with "requires gcloud-aio-storage" so we can run it adhoc for now.

datastore/tests/integration/conftest.py
nickfirespotter: i did this as a os enviorn so by default the test doesn't run since it can take 10s of seconds.

datastore/tests/integration/smoke_test.py
nickfirespotter: hmm, pretty sure it did work. it skipped when I didn't have the env var set. I can try again

datastore/tests/integration/smoke_test.py
nickfirespotter: okay, i though it would be better to use a new kind just for this test. Why use the same kind for all?

datastore/noxfile.py
nickfirespotter: k, well not sure the independent versioning is important since this dependency is solely for unit tests. If in the future datastore ever does depend on storage then it can be removed. But if you want it removed we can, will just make running the export test more complicated.

TheKevJames:
Merged separately so I could apply eg. `s/fststaging/voiceai-staging`. Releasing shortly!

PR #77, opened 2019-05-22
Storage signed url

nickfirespotter:
Create utility function on GCS Blob for a signed url for non-logged in access to the GCS blob.

### DIFF
--- a/storage/gcloud/aio/storage/blob.py
+++ b/storage/gcloud/aio/storage/blob.py
@@ -7,8 +7,9 @@ from typing import Optional
 from urllib.parse import quote
 
 import aiohttp
-from gcloud.aio import auth
-from gcloud.aio.auth import utils as auth_utils
+from gcloud.aio.auth import decode  # pylint: disable=no-name-in-module
+from gcloud.aio.auth import IamClient  # pylint: disable=no-name-in-module
+from gcloud.aio.auth import Token  # pylint: disable=no-name-in-module
 
 
 HOST = 'storage.googleapis.com'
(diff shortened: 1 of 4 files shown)


storage/gcloud/aio/storage/bucket.py
```diff
@@ -21,6 +21,15 @@ def __init__(self, storage, name: str) -> None:
 
         return Blob(self, blob_name, metadata)
 
+    async def blob_exists(self, blob_name: str,
```
TheKevJames: Looks like this method is just `bool(get_blob(..))`. I'm not sure its worth creating this explicitly since its just duplicating data and promotes a "check then get" usage, which would be a couple unnecessary API hits. Thoughts?

storage/tests/integration/conftest.py
```diff
@@ -5,7 +5,8 @@
 
 @pytest.fixture(scope='module')  # type: ignore
 def bucket_name() -> str:
-    return 'voiceai-staging-public-test'
+    return 'fststaging_nick_testing'
```
TheKevJames: Note to self: don't merge until this gets changed back.

storage/gcloud/aio/storage/blob.py
```diff
@@ -26,3 +38,75 @@ def chunk_size(self) -> int:
...
+            iam_client: Optional[IamClient] = None,
+            service_account_email: Optional[str] = None,
+            project: Optional[str] = None, service_file: Optional[str] = None,
+            token: Optional[Token] = None,
+            session: Optional[aiohttp.ClientSession] = None) -> str:
+        """ Create a temporary access URL for Storage Blob accessible by
+        anyone with the link.
+
+        # Adapted from Google Documentation:
+        https://cloud.google.com/storage/docs/access-control/signing-urls-manually#python-sample
```
TheKevJames: Looks like a pretty direct port to me. Only thing I saw missing was the `expiration > 604800` check. What's the reason behind removing it? I could go either way.

storage/tests/integration/conftest.py
nickfirespotter: doh, totally meant to revert this, will do now.

storage/gcloud/aio/storage/bucket.py
nickfirespotter: what about:
try:
  self.get_blob(blob_name, session)
  return True
except aiohttp.ClientResponseError:
  return False

I think it is nice to give a caller the condition check without having to worry about the ClientResponseError

storage/gcloud/aio/storage/blob.py
nickfirespotter: nope, I can add that in if you want.

storage/gcloud/aio/storage/bucket.py
TheKevJames: Wouldn't that potentially give some incorrect values? For example, a `ClientResponseError(408)` might just be a network blip -- that method would incorrectly return False when it might indeed be the case that the blob exists.

storage/gcloud/aio/storage/blob.py
TheKevJames: Yeah, sounds good to keep the block as a direct port, then.

storage/gcloud/aio/storage/bucket.py
nickfirespotter: well if there is a network outage then if I were to attempt the solution you suggested we would get an exception raised as well. We could check the response code is 404 as well with the hopes that they never return 404 unless the object really isn't there.

TheKevJames: Hmm, good point. Alright, if we're only returning a `False` when we're absolutely sure (eg. 404s, maybe others? we'll have to double-check their API responses) and re-raising the error in other cases, that seems reasonable to me.

nickfirespotter: I was thinking about this some more, do we really want to return True if the API request returns a 5XX error? Seems like a developer would use this as a precursor to proceeding further with a task knowing the file was available. If GCS API was having issues, I think I would want this to be False rather than True. So IMO I think the general check for the aiohttp.ClientResponseError seems correct to me. But if you disagree, I will build to what you think

TheKevJames: I think I could go either way on this -- I'm just a bit worried about returning incorrect results. Safest would probably be to just let/force the user to deal with situations differently; eg. if they want to retry on a 503 vs assume the system is down and move on is probably implementation-dependent, so I wouldn't want to prevent users from handling those differently.

Another thing to consider: a 404 might be a 403 in disguise rather than actually meaning the object doesn't exist. I think it'd be fine to return `False` there, but users would definitely need to account for that case if need be.

I think we should do:
- any 2xx or successful download counts as `exists == True`
- 404 or 410 counts as `exists == False`
- re-raise any other errors

PR #80, opened 2019-06-13
Add support for upload with metadata

ondrejSkowronek:


storage/gcloud/aio/storage/storage.py
```diff
@@ -102,9 +102,9 @@ def get_bucket(self, bucket_name: str) -> Bucket:
 
     async def upload(self, bucket: str, object_name: str, file_data: Any,
                      *, content_type: str = None, headers: dict = None,
+                     metadata: dict = None,
                      session: aiohttp.ClientSession = None, timeout: int = 30,
                      force_resumable_upload: bool = None) -> dict:
-        token = await self.token.get()
```
TheKevJames: Any reason you've changed this?

storage/gcloud/aio/storage/storage.py
```diff
@@ -241,24 +241,26 @@ def _split_content_type(content_type: str) -> Tuple[str, str]:
...
-                                                  session=session)
+                                                  metadata, session=session, )
         data: dict = await self._do_upload(session_uri, stream,
                                            headers=headers, session=session,
                                            timeout=timeout)
         return data
 
     async def _initiate_upload(self, url: str, object_name: str, headers: dict,
-                               *,
+                               metadata: dict, *,
```
TheKevJames: Should this also be a kwarg with default `= None`?

ondrejSkowronek: Previously I wanted to be just arg, but pylint was telling me that this function has too much arguments, so I changed it because of this

storage/gcloud/aio/storage/storage.py
ondrejSkowronek: Same reason as previous one, pylint was telling me that function has too many local variables, because I have added new arguments. And this variable was the only one which is not used twice, so I chose to this local variable to remove.

TheKevJames: Gotcha -- for now, let's leave this line as is and have the `metadata` arg later on be a kwarg as well; feel free to add some pylint rule disables to get that all working. Once this is in, I'll tune our pylint settings -- I suspect they're a bit too sensitive.

TheKevJames:
I went ahead and applied the PR changes myself for lack of response. Merging this as #87. Thanks for the contribution! :)

PR #81, opened 2019-06-14
feat(pubsub.subscription): kill old code and introduce new subscription module

kylepad:
### In this PR:
- kill all old pubsub code, it's broke and it dont work no more
- **new** Subscription and Utils code

### TODO: 
- [x] update README


pubsub/gcloud/aio/pubsub/subscription.py
```diff
@@ -1,66 +1,106 @@
...
-            await asyncio.sleep(1)
-
-        return await self.acknowledge(ack_ids, client=client,
-                                      retries=retries-1)
+    def __init__(self,
+                 subscription: str,
+                 *,
+                 loop: Optional[asyncio.AbstractEventLoop] = None
+                 ) -> None:
+        self.subscription = subscription
```
TheKevJames: Looks like the Google API has this value passed in to `create_subscription` and `subscribe` rather than setting it in `init` -- let's try to follow their interface here to make this as close to a drop-in as possible.

pubsub/gcloud/aio/pubsub/subscription.py
```diff
@@ -1,66 +1,106 @@
...
-        if self.exists(client=client):
-            return
+        if not loop:
+            loop = asyncio.get_event_loop()
+        self.loop = loop
 
+    def create_subscription(self,
+                            topic: str,
+                            *,
+                            ack_deadline_seconds: int = 5
```
TheKevJames: Tempted to just accept a `**kwargs` here and pass 'em directly to the Google client, so we don't have to keep this in sync with anything. Thoughts on that?

If so, let's link to [their docstring](https://github.com/googleapis/google-cloud-python/blob/11c72ade8b282ae1917fba19e7f4e0fe7176d12b/pubsub/google/cloud/pubsub_v1/gapic/subscriber_client.py#L236) rather than re-write our own.

TheKevJames:
Please also update the [README](https://github.com/talkiq/gcloud-aio/blob/master/pubsub/README.rst)

pubsub/gcloud/aio/pubsub/subscription.py
kylepad: yeah that sounds good to me

pubsub/gcloud/aio/pubsub/utils.py
```diff
@@ -0,0 +1,46 @@
...
+    Here are the gotchas is uses to do so:
+    - sets `future.__class__` so the `isinstance` check works
+    - sets `future._condition`, `future._state`, and `future._done_callbacks`
+      to their equivalent expected values (these are the attributes which
+      Google decided to avoid mirroring from `concurrent.futures`)
+    - spawns an infinite task which `await`s every second, which prevents the
+      Google future from occasionally getting stuck
+    """
+    # BEWARE: here be dragons
+    async def await_on_interval(interval: int) -> None:
```
kylepad: **COPIED FROM PREVIOUS PR**
**egalpin** 3 days ago
Is this function a good idea? What's the run-time story? The task wakes every 1 second? Can it be preempted if its underlying IO task completes before the next tick? Can a failure mode result in a task that just loops forever?

We've used it before in offline from the sounds of it. I just mean...how much faith do we have in this yielding consistent behavior?

kylepad: @TheKevJames do you have any additional incite here?

TheKevJames: To be honest, this is mostly magic and dragons to me, as well. Here's what I remember:
- Google's subscriber would never wake up after the first task unless I had _any_ other async task scheduled at all times -- eg. running a task which completes didn't do the trick.
- If the interval was sub-second, profiling suggested a huge amount of time (think >90%) was spent in that function.
- If the interval was large, subscriptions would get batched, ie. we'd grab `flow_control.max_messages` and process them, then do nothing until the next time this method woke up.

So basically my thought is that fixing Google subscriptions requires a noop async looping task which does nothing but kick the scheduler as often as possible but no more often than about a second. That... sounds idiotic to write out, so I suspect we could tune this a bit more.

pubsub/README.rst
```diff
@@ -13,54 +13,164 @@ Installation
 Usage
 -----
 
-This Pub/Sub implementation is a slim wrapper around
-``google-cloud-pubsub==0.25.0`` with a few key features:
+This Pub/Sub implementation is based on ``google-cloud-pubsub==0.41.0``
```
TheKevJames: Could we do anything around version ranges? What's the earliest `google-cloud-pubsub` version which this methodology would support? (I _think_ its 0.26.0, but we'll want to double-check).

pubsub/README.rst
```diff
@@ -13,54 +13,164 @@ Installation
...
+        max_lease_duration=10,
+    )
+
+    keep_alive = client.subscribe(
+        'subscription_name',
+        message_callback,
+        flow_control=example_flow_control
+    )
+
+Understand how modifying FlowControl affects how your pubsub runtime will operate can
```
TheKevJames: Grammar: `Understanding`, `can *be* confusing,`.

pubsub/requirements.txt
```diff
@@ -1,2 +1 @@
-google-auth >= 1.0.0, < 2.0.0
-google-cloud-pubsub == 0.25.0
+google-cloud-pubsub==0.41.0
```
TheKevJames: See version range comment.

pubsub/README.rst
kylepad: words, hard 
🦆

pubsub/README.rst
kylepad: Looks like we could do `google-cloud-pubsub >= 0.29.4`

PR #83, opened 2019-07-04
feat(storage): optional query parameters support in "upload"

DmytroDialpad:
Possibility to provide Optional query parameters to Storage.upload
This will work for simple and resumable uploads.
For instance, optional parameters can be used to set predefined
ACL on the object, provide kmsKeyName etc.

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -100,8 +100,10 @@ class Storage:
         data: dict = await resp.json()
         return data
 
+    # pylint: disable-msg=too-many-locals
     async def upload(self, bucket: str, object_name: str, file_data: Any,
-                     *, content_type: str = None, headers: dict = None,
+                     *, content_type: str = None, parameters: dict = None,
+                     headers: dict = None,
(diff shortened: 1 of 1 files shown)


storage/gcloud/aio/storage/storage.py
```diff
@@ -100,8 +100,10 @@ def get_bucket(self, bucket_name: str) -> Bucket:
         data: dict = await resp.json()
         return data
 
+    # pylint: disable-msg=too-many-locals
```
DmytroDialpad: We had 15 local variables in total with the limit of 15, after I've added "parameters" pylint started to complain. Decided to suppress warning for now as changing the interface or splitting the method can be a separate refactoring commit, but can include everything to this pull request if we want to

TheKevJames: Sounds good to me. nit: please use `disable` instead of `disable-msg` -- the latter has been deprecated as of pylint 0.21.0.

DmytroDialpad: changed in the new commit

PR #87, opened 2019-07-17
Add support for upload with metadata (re-home)

TheKevJames:


TheKevJames:
(Copying from #80, merging w/o secondary review since it had my review on the first ticket)

PR #88, opened 2019-07-22
feat(datastore) Add support for geoPointValue

KevinGDialpad:
The Python3-only and async version of https://github.com/talkiq/gcloud-rest/pull/40 with @TheKevJames' comments addressed!


datastore/gcloud/aio/datastore/value.py
```diff
@@ -82,5 +85,6 @@ def _infer_type(self, value: Any) -> TypeName:
     def _get_supported_types(cls):
         return {
             **TYPES,
+            LatLng: TypeName.GEOPOINT,
```
TheKevJames: This is already in `TYPES`, no need to re-add it here

KevinGDialpad: Aha, believe it or not, I have this change locally but it didn't make it into the commit! Amending now

PR #91, opened 2019-08-01
Make sure bigquery inserts actually return a result

kvdveer:
When google receives a bigtable insert and something goes wrong, this is not reflected by the HTTP status. Instead, the response contains a report of what did and what didn't work.

Prior this change this report was silently ignored by `gcloud.aio.bigquery.Table.insert`, and there was no way to notice data problems.

### DIFF
--- a/bigquery/gcloud/aio/bigquery/bigquery.py
+++ b/bigquery/gcloud/aio/bigquery/bigquery.py
@@ -65,7 +65,11 @@ class Table:
     async def insert(self, rows: List[Dict[str, Any]],
                      skip_invalid: bool = False, ignore_unknown: bool = True,
                      session: Optional[aiohttp.ClientSession] = None,
-                     timeout: int = 60) -> None:
+                     timeout: int = 60) -> List[Dict[str, Any]]:
+        """Streams data into BigQuery
+
+        Returns a list of insertion errors (if any)"""
+
         if not rows:
             return
 
@@ -90,3 +94,4 @@ class Table:
         resp = await session.post(url, data=payload, headers=headers,
                                   params=None, timeout=timeout)
         resp.raise_for_status()
+        return await resp.json()['insertErrors']


TheKevJames:
Wow, that's... absurd. What a weird way to build an API. I'd be inclined to think it might be nice to "fix this for them". What would you think about raising an exception if the API returns errors? I would imagine the code could look something like:
```py
resp.raise_for_status()
data = await resp.json()
if data.get('insertErrors'):
    raise aiohttp.ClientResponseError(
        resp.request_info, resp.history, status=400,
        message=[e['errors'] for e in data['insertErrors']],
        headers=resp.headers)
return data
```

kvdveer:
Problem is that you'd be raising partial exceptions. Of the 1000 rows sent, one might have an issue, where the other 999 would have been inserted properly. Raising an exception for a partial success is as counter-intuitive as not raising an exception for a partial failure.

Given that there's no right answer, I'd argue for not trying to change the API too much, as that would yield the least surprise. Users of bigquery are (or should be) aware of this quirk. They just need a way to correctly deal with it.

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -90,3 +90,4 @@ def _make_insert_body(rows: List[Dict[str, Any]],
         resp = await session.post(url, data=payload, headers=headers,
                                   params=None, timeout=timeout)
         resp.raise_for_status()
+        return await resp.json()
```
TheKevJames: Please also update this function's [type hint](https://github.com/talkiq/gcloud-aio/pull/91/files#diff-1e69c56cdf8bd8bf1742278093d488edR68): should now be `-> Dict[str, Any]`, I believe.

kvdveer: I've updated the signature to a list of insertion errors. I think that's the most useful approach - there's not much of interest in the list otherwise.

kvdveer:
> Ah, good call, I didn't realize this was a set of partial errors, but this makes sense. Maybe we should do a `log.warning` here?

That doesn't feel right to me. A user who writes code to correctly handle the insertion errors, would still be pestered with false warnings.

TheKevJames:
Good point on the logs. I think I'd rather return the entire response here and let the user drop the fields they don't care about, but we've been going back and forth enough -- happy to apply that patch myself.

We appreciate the contribution, thanks for helping us build a better library!

PR #94, opened 2019-08-09
Allow service data to be passed directly to a token or an API

### LINKED ISSUE #92: Token should accept the service data directly
In my case the JSON credentials are not in a file, they are in memory. It would be ideal if the token object also accepted credentials directly. Right now I have to write them to a temporary file.

rubik:
This PR adds the ability to call `Token` or any API with `service_data` instead of just `service_file`. It is useful when one already has the service data in memory as a dictionary and not in a file.

Fixes #92.

### DIFF
--- a/auth/gcloud/aio/auth/iam.py
+++ b/auth/gcloud/aio/auth/iam.py
@@ -1,4 +1,5 @@
 import json
+from typing import Any
 from typing import Dict
 from typing import List
 from typing import Optional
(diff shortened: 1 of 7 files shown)


TheKevJames:
First off: thanks for the contribution, we really appreciate it!

Could you please update this PR to use the `io` method described [on the issue](https://github.com/talkiq/gcloud-aio/issues/92#issuecomment-520069162)? I'd like to avoid changing the interface and plumbing this data through as much as possible. If you'd like to update the `gcloud-aio-auth` README with the usage of the new flow, that would be fantastic.

rubik:
@TheKevJames In that case, I think it's best to close this PR and open a new one.

Issue #53, opened 2018-12-05
ensure outside contributors can run integration tests


TheKevJames:

JensMadsen:
Any progess here :-) Now unless you hack the tests a bucket to which I have no credentials file is used.

TheKevJames:
Sorry @JensMadsen, haven't had a chance to get to this yet. Hopefully soon, but we've had a busy quarter here so far, so not as much time for OSS contributions as I'd like :)

PR #95, opened 2019-08-10
Allow service_file to be an io.IOBase instance

### LINKED ISSUE #92: Token should accept the service data directly

rubik:
This PR allows `service_file` to be any `io.IOBase` instance, like `io.StringIO`. In general, it allows any object with a `read()` method, but I think it's best to keep the type declaration stricter. A file-like object should inherit from `io.IOBase` anyway. The README was updated to reflect this change.

Fixes #92

### DIFF
--- a/auth/gcloud/aio/auth/token.py
+++ b/auth/gcloud/aio/auth/token.py
@@ -11,8 +11,8 @@ import time
 from typing import Any
 from typing import Dict
 from typing import List
-from typing import Union
 from typing import Optional
+from typing import Union
 from urllib.parse import quote_plus
 from urllib.parse import urlencode
 
--- a/auth/gcloud/aio/auth/token.py
+++ b/auth/gcloud/aio/auth/token.py
@@ -4,12 +4,14 @@ Google Cloud auth via service account file
 import asyncio
 import datetime
 import enum
+import io
 import json
 import os
 import time
 from typing import Any
 from typing import Dict
 from typing import List
+from typing import Union
 from typing import Optional
 from urllib.parse import quote_plus
 from urllib.parse import urlencode
@@ -34,7 +36,7 @@ class Type(enum.Enum):
     SERVICE_ACCOUNT = 'service_account'
 
 
-def get_service_data(service: Optional[str]) -> Dict[str, Any]:
+def get_service_data(service: Optional[Union[str, io.IOBase]]) -> Dict[str, Any]:
     service = service or os.environ.get('GOOGLE_APPLICATION_CREDENTIALS')
     if not service:
         cloudsdk_config = os.environ.get('CLOUDSDK_CONFIG')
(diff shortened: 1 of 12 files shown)


auth/gcloud/aio/auth/token.py
```diff
@@ -34,7 +36,7 @@ class Type(enum.Enum):
     SERVICE_ACCOUNT = 'service_account'
 
 
-def get_service_data(service: Optional[str]) -> Dict[str, Any]:
+def get_service_data(service: Optional[Union[str, io.IOBase]]) -> Dict[str, Any]:
```
TheKevJames: Lint fix:
```suggestion
def get_service_data(
        service: Optional[Union[str, io.IOBase]]) -> Dict[str, Any]:
```

auth/gcloud/aio/auth/token.py
```diff
@@ -56,13 +58,16 @@ def get_service_data(service: Optional[str]) -> Dict[str, Any]:
             raise
 
         return {}
+    except TypeError:
+        data: Dict[str, Any] = json.loads(service.read())
```
TheKevJames: In case this fails (eg. the in-memory `service_file` is not json-loadable, we'll want to maintain the old behavior of returning and empty dictionary. Perhaps something like:
```suggestion
        try:
            data: Dict[str, Any] = json.loads(service.read())
            return data
        except Exception:  # pylint: disable=broad-except
            return {}
```
Alternatively, maybe we could nest this within the first `try` in order to avoid duplicating the final global `Exception` block -- open to anything that maintains the behavior!

auth/tests/unit/token_test.py
```diff
@@ -1,5 +1,33 @@
...
+import pytest
 
 
 def test_importable():
     assert True
+
+
+@pytest.mark.asyncio
+async def test_service_as_io():
+    service_data = {
```
TheKevJames: You can go ahead and insert a `# pylint: disable=line-too-long` at the top of this method, since that's totally ok for test data; but please fix the other lint issue (4-space indents instead of 2-space)

rubik:
@TheKevJames I have fixed the issues but the CI pipeline seems to be failing for unrelated reasons.

TheKevJames:
@rubik yes, sorry about that, external users running the integration tests is a known issue right now (see #53) since we don't want to expose the GCP service account the tests require for billing and security reasons. I've re-run the tests with my permissions, and the results [look good](https://circleci.com/workflow-run/647cb11e-b734-449d-856d-2ba538b7dfb5).

Just a lint issue yet to be fixed: feel free to apply [this patch](https://github.com/talkiq/gcloud-aio/commit/9cf13d0fa1203588840cfa0f5e6776f8cc44d1c3) to fix it up.

rubik:
@TheKevJames I see, thanks for the explanation. The lint issues have been fixed.

TheKevJames:
Perfect, this LGTM. Thanks for the contribution! I'll get `gcloud-aio-auth` released with this new code right away.

rubik:
@TheKevJames Thanks for taking the time to review!

PR #101, opened 2019-09-04
feat(allow-rest-compatible-build): Refactors and changes to automate gcloud-rest builds

harsh-dialpad:
In this PR:
- Refactored REST API calls into a `Service` class object
- Abstracted `IO` calles in `Service` objects
- Updated CI to strip `aysnc/await` keywords
- Updated CI to use `python-future`'s `pasteurize` to make py3 code compatible wth py2
- Fix a few python 2 and python 3 incompatibilities like missing classes, inheritance trees, etc.
- Added linting for massaged code
    - Albeit with lots of checks disabled.

Update:
- Removed `pubsub` because it does not fit the design
- Enabled integration tests for `gcloud-rest`

Note:
- `noxfile.py` and `requirements.txt` need to updated upon approval to point to the correct package versions.


jonathan-johnston:
Nice, Harsh! Okay I'm going to have to strap in for a review session.

harsh-dialpad:
> Nice, Harsh! Okay I'm going to have to strap in for a review session.

Hey @jonathan-johnston, I thought of a way to make the changes transparent to the caller. I am pushing those in soon so if you want you can hold off on the review right now.

taskqueue/tests/integration/pushqueue_test.py
```diff
@@ -13,7 +13,8 @@
     schedule_time = datetime.utcnow() + timedelta(days=1)
 
     task = {
-        'scheduleTime': f'{schedule_time.isoformat("T")}Z',
+        'scheduleTime': '{}Z'.format(
+            schedule_time.isoformat(text_to_native_str('T'))),
```
TheKevJames: Why do we need the `text_to_native_str`?

.circleci/config.yml
```diff
@@ -23,14 +41,28 @@ jobs:
...
           command: |
             cd $(echo "${CIRCLE_TAG}" | sed 's/-.*//')
             python setup.py sdist bdist_wheel
             twine upload dist/*
 
+  pypi-rest:
+    docker:
+      - image: python:3.7.4-slim
+    steps:
+      - run: pip install pyopenssl twine
```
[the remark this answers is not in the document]
jonathan-johnston: Ditto for other places that do this in this file -- I know some of that code already existed, but I think it's worth doing anyway.

auth/gcloud/aio/auth/iam.py
```diff
@@ -5,23 +5,29 @@
...
 from .token import Token
 from .token import Type
 from .utils import encode
 
+# Selectively load libraries based on the package
+# TODO: Can we somehow just pick up the pacakge name instead of this
+if BUILD_GCLOUD_REST:
+    from requests import Session
+else:
+    from aiohttp import ClientSession as Session
```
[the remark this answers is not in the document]
jonathan-johnston: Hmm, I do like the explicitness of this block. You're going to get async unless you explicitly tell the library to be built as synchronous.

auth/gcloud/aio/auth/iam.py
```diff
@@ -5,23 +5,29 @@
 from typing import Optional
 from typing import Union
 
-import aiohttp
-
+from .build_constants import BUILD_GCLOUD_REST
+from .session import AioSession as RestSession
```
jonathan-johnston: This tripped me up the first time I read it: `RestSession` is also async, but building with `BUILD_GCLOUD_REST` builds as synchronous? The naming is overloaded.

auth/gcloud/aio/auth/session.py
```diff
@@ -0,0 +1,157 @@
...
+from typing import Dict
+
+import requests
+
+from .build_constants import BUILD_GCLOUD_REST
+
+class BaseSession():
+    __metaclass__ = ABCMeta
+
+    def __init__(self, session=None, conn_timeout: int = 10,
```
jonathan-johnston: Untyped `session` in typed context?

jonathan-johnston: Also no return types?

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -24,13 +34,13 @@ class Table:
     def __init__(self, dataset_name: str, table_name: str,
                  project: Optional[str] = None,
                  service_file: Optional[Union[str, io.IOBase]] = None,
-                 session: Optional[aiohttp.ClientSession] = None,
+                 session: Optional[Session] = None,
```
jonathan-johnston: Hm, why is `session` optional in this class, but not in the `IamClient` class?

datastore/gcloud/aio/datastore/value.py
```diff
@@ -48,8 +48,8 @@ def from_repr(cls, data: Dict[str, Any]) -> 'Value':
         else:
             supported = [name.value for name in supported_types.values()]
             raise NotImplementedError(
-                f'{data.keys()} does not contain a supported value type'
-                f' (any of: {supported})')
+                '{} does not contain a supported value type'
+                ' (any of: {})'.format(data.keys(), supported))
```
jonathan-johnston: I find format strings to be a lot more readable when you use references instead of positional interpolation:
```python
'{vals} does not contain a supported value type'
' (any of: {supported})'.format(vals=data.keys(), supported=supported))
```

storage/gcloud/aio/storage/storage.py
```diff
@@ -59,31 +79,34 @@ def get_bucket(self, bucket_name: str) -> Bucket:
         }
 
         if not self.session:
-            self.session = aiohttp.ClientSession(conn_timeout=10,
-                                                 read_timeout=10)
-        session = session or self.session
-        resp = await session.delete(url, headers=headers, params=params or {},
-                                    timeout=timeout)
-        resp.raise_for_status()
```
jonathan-johnston: Why no more `raise_for_status()`?

taskqueue/gcloud/aio/taskqueue/queue.py
```diff
@@ -44,25 +51,24 @@ def __init__(self, project: str, taskqueue: str,
...
+            self.session = RestSession()
+        s = RestSession(session) if session else self.session
         headers = await self.headers()
 
         resp = await s.request(method, url, headers=headers, **kwargs)
         # N.B. This is awaited early to give an extra helping hand to various
         # debug tools, which tend to be able to capture assigned variables but
         # not un-awaited data.
         data = await resp.json()
-        resp.raise_for_status()
```
jonathan-johnston: Why don't we want to `raise_for_status` anymore? When are we receiving exceptions? This seems like a change in behaviour of the library.

auth/gcloud/aio/auth/iam.py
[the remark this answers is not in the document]
TheKevJames: Fair enough -- I added a bunch of comments throughout this PR about ways we could avoid the `BUILD_GCLOUD_REST` variable, since a build var which gets magically changed by a `sed` command might be a bit hard to grok; not all places are necessarily better without this block... but I think it might be worth it to remove that key; the fewer places we need to think through building multiple code paths to add a feature to `gcloud-aio` / `gcloud-rest`, the better; otherwise we're not going to get too much of a win from merging the codebases, since we'll be building things twice anyway.

That said, maybe there are some places where the `if` makes more sense! I could go either way in some cases.

datastore/gcloud/aio/datastore/value.py
TheKevJames: Good call -- if we can't have f-strings, having references would be the next best thing.

storage/gcloud/aio/storage/storage.py
TheKevJames: @jonathan-johnston looks like `raise_for_status()` is now called within the `session.{get,post,etc}()` methods. I could go either way on whether its better to have this be implicit -- I do kind of like the code reduction and guarantee of handling errors.

taskqueue/gcloud/aio/taskqueue/queue.py
TheKevJames: https://github.com/talkiq/gcloud-aio/pull/101#discussion_r321904133

We're still calling `raise_for_status`, but its now called within the `resp = await s.request` call above.

Note that this breaks the `N.B.` comment above this line, though! Maybe we should emulate the `auto_raise_for_status` arg in requests here, so we can selectively not have the Session lib call this for us?

.circleci/config.yml
```diff
@@ -23,14 +41,28 @@ jobs:
       - image: python:3.7.4-slim
     steps:
       - run: pip install pyopenssl twine
-      - checkout
```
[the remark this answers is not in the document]
harsh-dialpad: So that is for `aio` code only. Since the workflow lints and packages both `aio` and `rest` code in a single iteration, we need a `nox` with `checkout` for `aio` and a `nox` with an attached `workspace` containing massaged code - `nox-rest`:   https://github.com/talkiq/gcloud-aio/blob/c06802930b12de9485daa0e5649827d6a40bab4e/.circleci/config.yml#L138

I will move `nox-rest` up to `nox` so its side by side for future reference.

.circleci/config.yml
```diff
@@ -4,6 +4,24 @@ orbs:
...
+        type: string
+    docker:
+      - image: python:2.7.15
+    steps:
+      - attach_workspace:
+          at: rest
+      - run:
+          command: |
+            cd rest/
+            git init && git add -A
```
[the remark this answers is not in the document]
harsh-dialpad: `pre-commit` won't run without a repository and I need to use only the massaged code for linting.
I can theoretically massage and lint in the same step (combine `py3to2` and `lint-rest`) but I foresee it becoming too unwieldy.
This way, ignoring this workaround, we have more or less two conceptually distinct steps that are easier to understand.
Let me have your thoughts.

taskqueue/tests/integration/pushqueue_test.py
harsh-dialpad: `future-fstrings-show` in `py3to2` adds `unicode_literals` which renders all strings as `unicode` in python 2. This function helps translate it back to string.
Ref: https://python-future.org/reference.html#future.utils.native_str_to_bytes

storage/gcloud/aio/storage/storage.py
harsh-dialpad: Anyone has more thoughts on this? We can go either way on this...

kms/gcloud/aio/kms/kms.py
```diff
@@ -40,36 +46,32 @@ def __init__(self, keyproject: str, keyring: str, keyname: str,
...
         url = f'{self.api_root}:decrypt'
         body = {
             'ciphertext': ciphertext,
         }
 
         if not self.session:
-            self.session = aiohttp.ClientSession(conn_timeout=10,
-                                                 read_timeout=10)
-        s = session or self.session
+            self.session = RestSession()
```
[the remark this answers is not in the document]
harsh-dialpad: Yeah, it is more or less redundant now. Let me find all occurrences of this and fix it.

bigquery/gcloud/aio/bigquery/bigquery.py
harsh-dialpad: From what I can tell, the it is probably something to do with the way the libraries are expected to be used. `auth` package seems to be a linked dependency for every other `gcloud` package and probably the original design had it as a required dependency to force the calling package to define a `session` so that the same `session` can also be used by the caller.

Which now makes me think if we should actually allow it as an optional parameter now in the updated code because this may create 2 sessions.

We can alternately update the code to pass the new `session` created in this line:
```self.session = RestSession(session) if session else RestSession()``` to other classes like `Token` like ```Token(..., session=self.session.session```

Thoughts?

File: bigquery/gcloud/aio/bigquery/bigquery.py [hunk 13]
```diff
@@ -6,13 +6,23 @@
...
-import aiohttp
+from gcloud.aio.auth import AioSession as RestSession  # pylint: disable=no-name-in-module
+from gcloud.aio.auth import BUILD_GCLOUD_REST  # pylint: disable=no-name-in-module
 from gcloud.aio.auth import Token  # pylint: disable=no-name-in-module
+
 try:
     import ujson as json
-except ModuleNotFoundError:
+except ImportError:
+     # HACK: Using `ImportError` instead of `ModuleNotFoundError` for python2
```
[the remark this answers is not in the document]
harsh-dialpad: Yeah, it was probably for me during the iteration 😄

auth/tests/unit/token_test.py
```diff
@@ -24,7 +24,7 @@ def test_importable():
         'auth_provider_x509_cert_url': 'https://www.googleapis.com/oauth2/v1/certs',
         'client_x509_cert_url': 'https://www.googleapis.com/robot/v1/metadata/x509/gcloud-aio%40random-project-123.iam.gserviceaccount.com'
     }
-    service_file = io.StringIO(json.dumps(service_data))
+    service_file = io.StringIO('{}'.format(json.dumps(service_data)))
```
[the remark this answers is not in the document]
harsh-dialpad: I can't even recall doing this, lol, thanks for noticing!

auth/noxfile.rest.py
```diff
@@ -0,0 +1,41 @@
+import os
```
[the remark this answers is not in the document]
harsh-dialpad: Yes, it is most likely possible. Let me add it.

auth/gcloud/aio/auth/utils.py
```diff
@@ -9,6 +10,14 @@ def decode(payload: str) -> bytes:
 
     See https://en.wikipedia.org/wiki/Base64#URL_applications
     """
+
+    if sys.version_info[0] < 3:
+        # Base64 encode/decode does not accept `str` as input in python2
```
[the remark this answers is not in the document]
harsh-dialpad: Sorry the comment is incorrect. By doing `future-fstrings-show`, it adds `unicode_literals` that redefines some classes so the default behaviour changes:

```
>>> import base64
>>> from __future__ import unicode_literals
>>> s = 'dGVzdA=='
>>> type(s)
<type 'unicode'>
>>> base64.b64decode(s, altchars=b'-_')
  File "/usr/local/lib/python2.7/base64.py", line 73, in b64decode
    s = s.translate(string.maketrans(altchars[:2], '+/'))
TypeError: character mapping must return integer, None or unicode
```

.circleci/config.yml
```diff
@@ -55,6 +87,70 @@ jobs:
...
+      - image: python:3.7.0-slim
+    steps:
+      - checkout
+      # Install some requirements
+      - run: pip install pytest future strip-hints future-fstrings[rewrite]
+
+      # Rename files and paths to `rest` and make strings py2 compatible
+      - run: find . -exec dirname {} \; | grep "aio" | sed 's/aio/rest/g' | xargs -L1 mkdir -p
+      - run: for file in $(find . -type f | grep "aio"); do cp -r $file $(echo "$file" | sed 's/aio/rest/g'); done
+      # For some files with conflicts due to unicode encoding in py2, do not use
```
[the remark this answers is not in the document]
harsh-dialpad: The comment is probably misleading and it is about `future-fstring-show` redifining strings and bytes. Some examples of places where this is an issue are:
- https://github.com/talkiq/gcloud-aio/blob/aa2436759b6cd926cdb405f9e46d7ea38c832a3e/datastore/tests/unit/value_test.py#L18
- https://github.com/talkiq/gcloud-aio/blob/aa2436759b6cd926cdb405f9e46d7ea38c832a3e/datastore/tests/unit/value_test.py#L13 (hence the need for the `if...else` in https://github.com/talkiq/gcloud-aio/blob/c06802930b12de9485daa0e5649827d6a40bab4e/datastore/tests/unit/value_test.py#L15

Most of these issues are in `datastore` package because we determine the message type based on the variable's `type` and that gets modified with the class redefinitions.

.circleci/config.yml
```diff
@@ -55,6 +87,70 @@ jobs:
...
+      - run: for file in $(find . -type f | grep '\.py$' | grep -Ev "datastore.*(value|constants|value_types_test|value_test)\.py$" | sort | uniq); do new_file=$(echo "$file" | sed 's/aio/rest/g'); temp_file="${new_file}.bkp"; future-fstrings-show $file > $temp_file; strip-hints $temp_file > "${temp_file}.tmp"; mv "${temp_file}.tmp" $new_file; rm $temp_file; echo "Formatting ${new_file}"; done
+
+      # Remove the files with `aio` in file names (we copied all required files
+      # already)
+      - run: find . -type f | grep "aio" | xargs -L1 rm -r
+
+      # Replace build metadata
+      - run: find . -type f | xargs -L1 sed -Ei 's/gcloud\.aio/gcloud\.rest/g'
+      - run: find . -type f | xargs -L1 sed -Ei 's/gcloud-aio/gcloud-rest/g'
+      - run: find . -type f | grep "setup\.py$" | xargs -L1 sed -Ei "s/python_requires='>= 3\.6'/python_requires='>= 2\.7\.15'/g"
```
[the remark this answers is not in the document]
harsh-dialpad: I am not sure,(and I am not sure how to test -- maybe a nox env??) but aren't there some major differences in 2.7.x revisions?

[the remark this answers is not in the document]
TheKevJames: Not really, no, its generally just been bugfixes for years and years. Maybe do a nox test as a one-off (eg. do `2.7.0` and `2.7.15` and if both work I'm good to just call it `>= 2.7`.

storage/gcloud/aio/storage/storage.py
TheKevJames: If no one else has comments: I think how you've refactored this is lovely :+1:

auth/tests/unit/token_test.py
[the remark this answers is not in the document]
harsh-dialpad: Okay, I remember this now. So when we run `py3to2` in CI, `future-fstrings-show` redefines `str` literals to `unicode`. That works great in this case because `io.StringIO` does not like `str` inputs in `python2`. So this noop operation actually helps out with that.

I think a comment explain this will be worthwhile?

auth/gcloud/aio/auth/iam.py
harsh-dialpad: Seems fair, I was stuck on the naming as well. I think I have a better way to handle this now.

harsh-dialpad: Let me know if this seems better. Now there is no aliasing of this class in code.

auth/gcloud/aio/auth/iam.py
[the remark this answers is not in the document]
harsh-dialpad: So, I think we can keep this while we are importing modules. Because depending on the client's `pyenv` it might have `aiohttp` etc. if they are using `gcloud-rest` in `python3`. Then we can get weird issues when wrong classes are being loaded.

However, I agree, we can remove the `if`s that are inline in code that are handling incompatibilities in package APIs.

.circleci/config.yml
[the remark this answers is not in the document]
jonathan-johnston: That sounds about right to me!

auth/gcloud/aio/auth/iam.py
```diff
@@ -5,23 +5,29 @@
...
 
 API_ROOT_IAM = 'https://iam.googleapis.com/v1'
 API_ROOT_IAM_CREDENTIALS = 'https://iamcredentials.googleapis.com/v1'
 SCOPES = ['https://www.googleapis.com/auth/iam']
 
 
 class IamClient:
-    def __init__(self, service_file: Optional[Union[str, io.IOBase]] = None,
-                 session: Optional[aiohttp.ClientSession] = None,
+    def __init__(self, session: Session,
```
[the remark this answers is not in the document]
harsh-dialpad: Ah, my bad, this seems to be a blooper 👍

auth/gcloud/aio/auth/session.py
jonathan-johnston: Hm, I would have thought the base class would work, right? Unless `mypy` doesn't work properly (I wouldn't be surprised).

bigquery/gcloud/aio/bigquery/bigquery.py
jonathan-johnston: Interesting dependency... And I thought based on your comment above that the `IamClient` would have an optional `session` input, but it's now required (but set to be `Optional`) as a positional argument.

auth/gcloud/aio/auth/session.py
harsh-dialpad: Well, to be honest, I don't know, but I specifically did not put a typing hint here because of differing types of `session` in `AioSession` and `SyncSession`. I can put something but I am not sure if it will actually help.

Do you have a suggestion on how we might improve this?

bigquery/gcloud/aio/bigquery/bigquery.py
harsh-dialpad: Crap, my bad, I am making it optional now.

taskqueue/gcloud/aio/taskqueue/queue.py
harsh-dialpad: Done!

harsh-dialpad:
PR updated and all reviewers' comments have been addressed. Feel free to re-review.

auth/gcloud/aio/auth/session.py
jonathan-johnston: Ah, for some reason I thought this was related to the session objects you were defining, but I misread the code. I think it's fine as is for now!

auth/gcloud/aio/auth/utils.py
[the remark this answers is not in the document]
TheKevJames: Makes sense -- let's make sure this ends up as a comment here.

auth/requirements.txt
```diff
@@ -1,4 +1,8 @@
 aiohttp >= 2.0.0, < 4.0.0
 backoff >= 1.0.0, < 2.0.0
 cryptography >= 2.0.0, < 3.0.0
+future >= 0.17.0, < 0.18.0
 pyjwt >= 1.5.3, < 2.0.0
+requests >= 2.20.0, < 2.23.0
```
TheKevJames: I think this `requirements.txt` should align with the usage of file-on-disks, eg. `-aio`, so that we don't break being able to do a local `pip install -e ./` and get things running -- anywhere we add/remove dependencies to do `aio` -> `rest` should be outside of this file.

Another option would be to comment it out in the file-on-disk and uncomment it in the py3to2 script (in the same way as we comment out aiohttp we could uncomment `requests`)

.circleci/config.yml
```diff
@@ -94,6 +142,64 @@ jobs:
             ./bin/linux/amd64/github-release release -t "${CIRCLE_TAG}"
             cat release-description.md | grep ${PROJECT} | ./bin/linux/amd64/github-release edit -t ${CIRCLE_TAG} -d -
 
+  py3to2:
+    docker:
+      - image: python:3.7.0-slim
+    steps:
+      - checkout
+      # Install some requirements
+      - run: pip install pytest==2.3.1 future==0.17.1 strip-hints==0.1.7 future-fstrings[rewrite]==1.2.0
```
TheKevJames: If we're pinning a bunch of stuff, let's put this in a `requirements.txt` for easier management. Maybe `requirements.py3to3.txt` in the root of the repo, or something like that?

auth/gcloud/aio/auth/session.py
```diff
@@ -0,0 +1,153 @@
...
+                                              **kwargs)
+            if auto_raise_for_status:
+                resp.raise_for_status()
+            return resp
+
+if BUILD_GCLOUD_REST:
+    import requests
+
+    class SyncSession(BaseSession):
+        google_api_lock = threading.RLock()
```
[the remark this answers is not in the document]
harsh-dialpad: Yes, that's true. Sorry must have missed this earlier, will definitely get this updated.

auth/gcloud/aio/auth/token.py
```diff
@@ -116,6 +141,11 @@ def __init__(self, service_file: Optional[Union[str, io.IOBase]] = None,
         return self.access_token
 
     async def ensure_token(self) -> None:
+        if BUILD_GCLOUD_REST:
```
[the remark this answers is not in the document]
harsh-dialpad: Updated!

auth/gcloud/aio/auth/iam.py
```diff
@@ -5,25 +5,30 @@
...
 
 
 class IamClient:
-    def __init__(self, service_file: Optional[Union[str, io.IOBase]] = None,
-                 session: Optional[aiohttp.ClientSession] = None,
+    def __init__(self, session: Optional[Session] = None,
+                 service_file: Optional[Union[str, io.IOBase]] = None,
                  token: Optional[Token] = None) -> None:
-        self.session = session
+        self.session = AioSession(session) if session else AioSession()
```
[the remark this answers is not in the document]
harsh-dialpad: Yes, that is true! Thanks for noticing, I didn't see that the provision was already there.

storage/gcloud/aio/storage/bucket.py
```diff
@@ -15,24 +23,24 @@ def __init__(self, storage, name: str) -> None:
...
 
     async def blob_exists(self, blob_name: str,
-                          session: aiohttp.ClientSession = None) -> bool:
+                          session: Optional[Session] = None) -> bool:
         try:
             await self.get_blob(blob_name, session=session)
             return True
-        except aiohttp.ClientResponseError as e:
+        except ResponseError as e:
             if e.status in {404, 410}:
```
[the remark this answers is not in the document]
harsh-dialpad: Thanks, I have updated this section

auth/requirements.txt
harsh-dialpad: Makes sense, updating it.

auth/gcloud/aio/auth/session.py
```diff
@@ -0,0 +1,157 @@
...
+                                              **kwargs)
+            resp.raise_for_status()
+            return resp
+
+class SyncSession(BaseSession):
+    google_api_lock = threading.RLock()
+
+    @property
+    def session(self) -> requests.Session:
+        self._session = self._session or requests.Session()
```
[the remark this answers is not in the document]
harsh-dialpad: Not necessarily: https://2.python-requests.org/en/master/api/#request-sessions

auth/gcloud/aio/auth/token.py
```diff
@@ -83,7 +103,7 @@ def __init__(self, service_file: Optional[Union[str, io.IOBase]] = None,
             self.token_type = Type.GCE_METADATA
             self.token_uri = GCE_ENDPOINT_TOKEN
 
-        self.session = session
+        self.session = RestSession(session) if session else RestSession()
```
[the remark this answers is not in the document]
harsh-dialpad: Ahhh, great point. I didn't even notice that I had already made the provision for this. Updating this.

auth/gcloud/aio/auth/session.py
```diff
@@ -0,0 +1,157 @@
...
+
+        async def request(self, method: str, url: str, headers: Dict[str, str],
+                          **kwargs: Any) -> aiohttp.ClientResponse:
+            resp = await self.session.request(method, url, headers=headers,
+                                              **kwargs)
+            resp.raise_for_status()
+            return resp
+
+class SyncSession(BaseSession):
+    google_api_lock = threading.RLock()
```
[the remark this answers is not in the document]
harsh-dialpad: I have updated the code to create a static class variable for `_google_api_lock`. Let me know if that looks okay.

auth/gcloud/aio/auth/session.py
```diff
@@ -0,0 +1,157 @@
...
+
+if BUILD_GCLOUD_REST:
+    import requests
+
+    class SyncSession(BaseSession):
+        _google_api_lock = threading.RLock()
+
+        @property
+        def google_api_lock(self) -> threading.RLock:
+            return SyncSession._google_api_lock  # pylint: disable=protected-access
```
TheKevJames: I'm about 80% convinced this will let us maintain the locking semantics `gcloud-rest` needs. Sanity check: here's the case we need to cover:

```python
class Token(object):
    def __init__(self, ..., google_api_lock=None, ...):
        self.google_api_lock = google_api_lock or threading.RLock()

class KMS(object):
    def __init__(self, ..., google_api_lock=None, ...):
        self.google_api_lock = google_api_lock or threading.RLock()
        self.access_token = Token(..., google_api_lock=self.google_api_lock)

class Storage(object):
    def __init__(self, ..., google_api_lock=None, ...):
        self.google_api_lock = google_api_lock or threading.RLock()
        self.access_token = Token(..., google_api_lock=self.google_api_lock)
```

In the above example, if a user is making use of both the `KMS` and `Storage` classes, they need to be able to pass in a single `RLock` which is shared amongst *all requests made to Google*, ie. the lock used in `Token`, `KMS`, and `Storage` instances all need to be the same instance.

I _think_ what you've done here is ensure that users can't / don't need to pass in the lock, but we will use the same one everywhere anyway. Is that the case? And, further sanity check: is that the case in py2 as well as py3?

harsh-dialpad: > I think what you've done here is ensure that users can't / don't need to pass in the lock, but we will use the same one everywhere anyway. Is that the case?
Yes

> And, further sanity check: is that the case in py2 as well as py3?
Yes, the "static" class variable behaves the same way in both versions.

TheKevJames: Awesome, thanks for the sanity check!

PR #103, opened 2019-09-17
initial commit: adding credentials to constructor

atezerjani:
I changed the constructor and added a credentials to let us make subscriber from the service key file, exactly the same as publisher

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -3,6 +3,7 @@ import concurrent.futures
 import signal
 from typing import Callable
 from typing import Optional
+from typing import Any
 
 from google.api_core import exceptions
 from google.cloud import pubsub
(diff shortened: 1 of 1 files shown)


pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -17,9 +17,13 @@ class SubscriberClient:
 
     def __init__(self,
                  *,
+                 credentials: Optional = None,
                  loop: Optional[asyncio.AbstractEventLoop] = None
                  ) -> None:
-        self._subscriber = pubsub.SubscriberClient()
+        if credentials:
+            self._subscriber = pubsub.SubscriberClient(credentials=credentials)
```
TheKevJames: Looks like this client [actually accepts kwargs](https://github.com/googleapis/google-cloud-python/blob/b4ea85cd6de240956b44ed25a4d449a9c3f1649d/pubsub/google/cloud/pubsub_v1/subscriber/client.py#L73). Rather than whitelisting the args we're aware of, might be easiest to just pass through everything -- it'd be nice to be able to avoid having to make a PR like this for each new feature that we aren't yet exposing.

TheKevJames:
Thanks for the contribution, and sorry this took so long blocked behind #101! I've fixed up the merge conflict and [merged this in](https://github.com/talkiq/gcloud-aio/commit/d5386551e00788a3c93b2f126823b435269d3634), you can expect this in `gcloud-aio-pubsub>1.0.2`.

Issue #84, opened 2019-07-09
Working example for cloud storage


adamshamsudeen:
Can you please share a working example for cloud storage? 
I'm got an error when I ran the smoke test code.
`aiohttp.client_exceptions.ClientResponseError: 404, message='Not Found'`

TheKevJames:
Writing up some more complete docs is definitely on our TODO list -- marking this one as "help wanted" to hopefully get some community help :)

In the meantime, just a sanity check: the exact bucket names etc in the smoke test would be referencing our internal test buckets which you would likely not have access to; make sure that, and the object name, and the project name, etc, is all configured for your use-case.

If you have a code snippet for the 404 you're experiencing, I'd be happy to help debug!

adamshamsudeen:
I got it working. How can I start contributing to the Docs?

TheKevJames:
For now, a PR modifying the "Usage" section of the relevant README would be awesome. I'd be happy to review any docs improvements like that!

adamshamsudeen:
I have submitted a PR #111 , please let me know id I should make any changes.

TheKevJames:
#111 has (long since) been merged, closing the ticket!

PR #111, opened 2019-10-16
Update readme: added an example to Usage

adamshamsudeen:
I have added a simple example to Readme.
With respect to the issue #84.


TheKevJames:
Ah, sorry, this PR slipped completely past my radar.

Thanks for the contribution, it'll be nice to get some docs here :)

PR #118, opened 2019-11-22
Set _waiters attribute to make it *appear* as concurrent.futures.Future

meetmangukiya:


TheKevJames:
@meetmangukiya thanks for the contribution! Looks good to me!

PR #121, opened 2019-12-02
feat(storage): implement copy operation

jonathan-johnston:
According to the GCP Storage API documentation, the copy operation can be achieved either by using the `copyTo` endpoint or the `rewriteTo` endpoint. Google recommends using `rewriteTo`, since `copyTo` simply uses a single call to the behaviour of `rewriteTo` whereas multiple may be needed for large files or other multi-part copy causes.

`rewriteTo` exposes a `done` field in the response payload; if this is false then the value of the `rewriteToken` field must be used in subsequent calls until the copy operation is completed.

https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite

Testing:

- Added an extra flow to the smoke test to (1) copy the originally uploaded object, (2) verify that its contents are expected, (3) delete it on cleanup, and (4) ensure the deletion was successful

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -62,15 +62,28 @@ class Storage:
                    headers: dict = None, params: dict = None,
                    timeout: int = 10,
                    session: Optional[Session] = None) -> bytes:
+
+        """
+        When files are too large, multiple calls to `rewriteTo` are made. We
+        refer to the same copy job by using the `rewriteToken` from the
+        previous return payload in subsequent `rewriteTo` calls.
+
+        Using the `rewriteTo` GCS API is preferred in part because it is able
+        to make multiple calls to fully copy an object whereas the `copyTo` GCS
+        API only calls `rewriteTo` once under the hood, and thus may fail if
+        files are large.
+
+        In the rare case you need to resume a copy operation, include the
+        `rewriteToken` in the `params` dictionary. Once you begin a multi-part
+        copy operation, you then have 1 week to complete the copy job.
+
+        https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite
+        """
         token = await self.token.get()
 
         if not new_name:
             new_name = object_name
 
-        # Using `rewriteTo` is preferred in part because it is able to make
-        # multiple calls to fully copy an object.
-        #
-        # https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite
         url = (f"{API_ROOT}/{bucket}/o/{quote(object_name, safe='')}/rewriteTo"
                f"/b/{destination_bucket}/o/{quote(new_name, safe='')}")
 
@@ -89,10 +102,6 @@ class Storage:
 
         data: dict = await resp.json()
 
-        # When files are too large, multiple calls to `rewriteTo` need to be
-        # made. To refer to the same copy job, the `rewriteToken` from the
-        # previous return payload is used in subsequent `rewriteTo` calls. The
-        # user then has 1 week to complete the copy job.
         while not data.get('done') and data.get('rewriteToken'):
             params['rewriteToken'] = data['rewriteToken']
             resp = await s.post(url, headers=headers, params=params,
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -87,6 +87,12 @@ class Storage:
         url = (f"{API_ROOT}/{bucket}/o/{quote(object_name, safe='')}/rewriteTo"
                f"/b/{destination_bucket}/o/{quote(new_name, safe='')}")
 
+        # We may optionally supply metadata* to apply to the rewritten
+        # object, which explains why `rewriteTo` is a POST endpoint; however,
+        # we don't expose that here so we have to send an empty body. Therefore
+        # the `Content-Length` and `Content-Type` indicate an empty body.
+        #
+        # * https://cloud.google.com/storage/docs/json_api/v1/objects#resource
         headers = headers or {}
         headers.update({
             'Authorization': f'Bearer {token}',
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -57,6 +57,40 @@ class Storage:
     def get_bucket(self, bucket_name: str) -> Bucket:
         return Bucket(self, bucket_name)
 
+    async def copy(self, bucket: str, object_name: str,
+                   destination_bucket: str, *, new_name: str = None,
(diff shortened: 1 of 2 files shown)


storage/gcloud/aio/storage/storage.py
```diff
@@ -57,6 +57,50 @@ def __init__(self, *, service_file: Optional[Union[str, io.IOBase]] = None,
...
+        resp = await s.post(url, headers=headers, params=params,
+                            timeout=timeout)
+
+        data: dict = await resp.json()
+
+        # When files are too large, multiple calls to `rewriteTo` need to be
+        # made. To refer to the same copy job, the `rewriteToken` from the
+        # previous return payload is used in subsequent `rewriteTo` calls. The
+        # user then has 1 week to complete the copy job.
+        while not data.get('done') and data.get('rewriteToken'):
```
allandialpad: would it be useful to wrap this while in something like:

`try:
    while:
        retry code
except Exception as Ex:
    logger.error(f`Exception {Ex} interrupted copy operation')
    raise Ex`

for posterity

maybe include some stats if they are available from the API or generate your own stats like `rewrite attempts` My concern comes from the concurrency here as exceptions in other parts of the event loop would interrupt this code 

thoughts?

storage/gcloud/aio/storage/storage.py
```diff
@@ -57,6 +57,50 @@ def __init__(self, *, service_file: Optional[Union[str, io.IOBase]] = None,
     def get_bucket(self, bucket_name: str) -> Bucket:
         return Bucket(self, bucket_name)
 
+    async def copy(self, bucket: str, object_name: str,
+                   destination_bucket: str, *, new_name: str = None,
+                   headers: dict = None, params: dict = None,
+                   timeout: int = 10,
+                   session: Optional[Session] = None) -> bytes:
+        token = await self.token.get()
+
```
allandialpad: few nits:

- I like docstrings, everywhere (do you guys purposefully avoid them to promote self documenting code?) Here I'd put a docstring explaining the variables, the logic of rewriteto, and that link to gcloud docs
- I like like variable pairs like `bucket` and `destination_bucket` to have explicit pair name like `source_bucket` and `destination_bucket`

storage/gcloud/aio/storage/storage.py
jonathan-johnston: I'll address each of your points:

1. The `rewriteToken` is not a retry operation, but a mechanic exposed by GCP to handle large copies. If the `rewriteTo` operation is incomplete, GCP requests that we keep following the API chain until it is complete. In that sense, it doesn't throw and we don't have to use a `try-except` block here.
2. To the best of my knowledge, exceptions in the event loop as a whole don't stop other coroutines from running unless the exception is not caught and crashes the program. In that case, the `rewriteTo` operation will expire in 7 days if not resumed and we're back to our initial state, so there's nothing to worry about. If the call is made again, then to the best of my knowledge a new `rewriteTo` operation is created and again after 7 days the previously unfinished one expired.

Does that change the context of your comments now?

jonathan-johnston: In addition, the user can always restart the copy if they have the `rewriteToken` by modifying the `params`. I can add that to the docstring.

storage/gcloud/aio/storage/storage.py
jonathan-johnston: The function signature is relatively straightforward and I'm following convention in this repo, but I could add the docstring to describe some of the function semantics.

storage/gcloud/aio/storage/storage.py
```diff
@@ -57,6 +57,59 @@ def __init__(self, *, service_file: Optional[Union[str, io.IOBase]] = None,
     def get_bucket(self, bucket_name: str) -> Bucket:
         return Bucket(self, bucket_name)
 
+    async def copy(self, bucket: str, object_name: str,
```
kylepad: I like Allan's suggestion about renaming `bucket` to `source_bucket`, I think that would make this slightly clearer, as well as the code below where it is used.

storage/gcloud/aio/storage/storage.py
```diff
@@ -57,6 +57,59 @@ def __init__(self, *, service_file: Optional[Union[str, io.IOBase]] = None,
...
+                   destination_bucket: str, *, new_name: str = None,
```
kylepad: optional nit: I find it's easier to parse function arguments when the default and keyword arguments are separated, so I'd move `*, new_name: str = None,` to the next line

storage/gcloud/aio/storage/storage.py
TheKevJames: > do you guys purposefully avoid them to promote self documenting code?

No, but that's the best excuse for our laziness that I've heard yet :D That said -- yes, I would argue that docstrings are great but the more you can write code which doesn't need documentation, the better. One big problem with docstrings (and any docs) is how easy it is for them to fall out-of-sync with code; especially in situations like this where all the arguments are convention through the whole repo, having every function have the same docstring re: those copied values could quickly get out of hand!

Note that the above isn't meant to be an argument against writing docs/comments/docstrings, more of a warning of things to watch out for!

storage/gcloud/aio/storage/storage.py
TheKevJames: I think keeping with the style is important as well, and we use `bucket` throughout this file. If we really want the variable names to pair, maybe change `destination_bucket` instead? I suppose it could just be `destination`.

storage/gcloud/aio/storage/storage.py
TheKevJames: I disagree here -- I think @jonathan-johnston did the right thing in sticking with the same coding conventions used throughout this repo. If we want to change convention like that, we should do it everywhere so that each chunk of this library doesn't have a different style. Lets leave this as-is for now and create a follow-up PR touching everything if we think that's worth it.

storage/gcloud/aio/storage/storage.py
```diff
@@ -57,6 +57,50 @@ def __init__(self, *, service_file: Optional[Union[str, io.IOBase]] = None,
...
+        # multiple calls to fully copy an object.
+        #
+        # https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite
+        url = (f"{API_ROOT}/{bucket}/o/{quote(object_name, safe='')}/rewriteTo"
+               f"/b/{destination_bucket}/o/{quote(new_name, safe='')}")
+
+        headers = headers or {}
+        headers.update({
+            'Authorization': f'Bearer {token}',
+            'Content-Length': '0',
```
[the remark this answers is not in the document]
jonathan-johnston: Definitely!

PR #124, opened 2019-12-11
Adding load, copy, get, delete APIs in bigquery table class.

nitishdialpad:


TheKevJames:
Awesome! This all looks good to me -- I'll get it out with some release notes as v3.2.0!

Issue #90, opened 2019-07-31
 Add support for arrayValue 


KevinGDialpad:
It's the last type missing.

https://github.com/talkiq/gcloud-aio/blob/ed062beca0a62b914b26946ff851224b4ed827ff/datastore/gcloud/aio/datastore/constants.py#L74-L84

We need to implement the [ArrayValue](https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/runQuery#ArrayValue) type - also see #88 for a similar task

TheKevJames:
Solved by @wosc and released in `gcloud-aio-datastore` v5.2.0~

PR #138, opened 2020-02-05
Add support for arrayValue

wosc:
See https://github.com/talkiq/gcloud-aio/issues/90

### DIFF
--- a/datastore/gcloud/aio/datastore/array.py
+++ b/datastore/gcloud/aio/datastore/array.py
@@ -1,0 +1,38 @@
+try:
+    from collections.abc import Sequence
+except ImportError:
+    from collections import Sequence
+from typing import Any
+from typing import Dict
+from typing import List
+
+from gcloud.aio.datastore import value
+
+
+# https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/runQuery#ArrayValue
+class Array(Sequence):
+
+    def __init__(self, items: List[Any]) -> None:
+        super(Array, self).__init__()
+        self.items = items
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Array):
+            return False
+        return self.items == other.items
+
+    def __repr__(self) -> str:
+        return str(self.to_repr())
+
+    @classmethod
+    def from_repr(cls, data: Dict[str, Any]) -> 'Array':
+        return cls([value.Value.from_repr(x) for x in data['values']])
+
+    def to_repr(self) -> Dict[str, Any]:
+        return {'values': [x.to_repr() for x in self]}
+
+    def __getitem__(self, index: Any) -> Any:
+        return self.items[index]
+
+    def __len__(self) -> int:
+        return len(self.items)
--- a/datastore/gcloud/aio/datastore/value.py
+++ b/datastore/gcloud/aio/datastore/value.py
@@ -2,6 +2,7 @@ from datetime import datetime
 from typing import Any
 from typing import Dict
 
+from gcloud.aio.datastore.array import Array
 from gcloud.aio.datastore.constants import TypeName
 from gcloud.aio.datastore.constants import TYPES
 from gcloud.aio.datastore.key import Key
@@ -35,6 +36,8 @@ class Value:  # pylint:disable=useless-object-inheritance
             if json_key in data:
                 if json_key == 'nullValue':
                     value = None
+                elif value_type == Array:
+                    value = Array.from_repr(data[json_key])
                 elif value_type == datetime:
                     value = datetime.strptime(data[json_key],
                                               '%Y-%m-%dT%H:%M:%S.%f000Z')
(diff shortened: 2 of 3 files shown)


datastore/gcloud/aio/datastore/value.py
```diff
@@ -35,6 +36,8 @@ def from_repr(cls, data: Dict[str, Any]) -> 'Value':
             if json_key in data:
                 if json_key == 'nullValue':
                     value = None
+                elif json_key == 'arrayValue':
```
TheKevJames: Looks like you've added this to `TypeName`, so you should be able to change this check to `elif value_type == Array:` (which would be great, since it would help us avoid copying that string around outside of our constants).

datastore/gcloud/aio/datastore/value.py
```diff
@@ -74,6 +77,10 @@ def to_repr(self) -> Dict[str, Any]:
     def _infer_type(self, value: Any) -> TypeName:
         odd_strings = {'newstr', 'unicode'}
         kind = str if value.__class__.__name__ in odd_strings else type(value)
+        # We should instead use Array as the key in TYPES directly, but that
```
TheKevJames: My [comment on your other PR](https://github.com/talkiq/gcloud-aio/pull/139#discussion_r377963227) actually gave me a thought for a way to avoid this hack -- rather than adding `Array` to `TYPES`, adding it directly in the `_get_supported_types()` method will allow you to nuke this hack and avoid the import cycle issue you ran into.

wosc:
Hi, thanks here as well for your thoughtful and detailed comments, I've updated the code accordingly.

datastore/gcloud/aio/datastore/array.py
```diff
@@ -0,0 +1,35 @@
+from typing import Any, Dict, Iterator, List
+import collections.abc
+import gcloud.aio.datastore
```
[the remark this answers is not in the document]
wosc: Again, unfortunately not possible due to import cycles. value.py already from-imports Array, so we cannot have array.py from-import Value in turn.

datastore/gcloud/aio/datastore/array.py
```diff
@@ -0,0 +1,43 @@
+try:
+    from collections.abc import Sequence
+except ImportError:
+    from collections import Sequence
+from typing import Any
+from typing import Dict
+from typing import Iterator
+from typing import List
+
+import gcloud.aio.datastore.value
```
TheKevJames: Lets go with `from gcloud.aio.datastore import value` here, please (see [my comment on the other PR](https://github.com/talkiq/gcloud-aio/pull/139#discussion_r380926182))

datastore/gcloud/aio/datastore/array.py
```diff
@@ -0,0 +1,43 @@
...
+        super(Array, self).__init__()
+        self.items = items
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Array):
+            return False
+        return self.items == other.items
+
+    def __repr__(self) -> str:
+        return repr(self.items)
```
TheKevJames: Could we define this as `return str(self.to_repr())` to match the `__repr__`s we have defined on our other classes?

datastore/gcloud/aio/datastore/array.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @classmethod
+    def from_repr(cls, data: Dict[str, Any]) -> 'Array':
+        return cls([gcloud.aio.datastore.Value.from_repr(x)
+                    for x in data['values']])
+
+    def to_repr(self) -> Dict[str, Any]:
+        return {'values': [x.to_repr() for x in self]}
+
+    def __iter__(self) -> Iterator:
```
TheKevJames: Do you need to define this? I believe the `Sequence` class will already do this for you (since you've defined `__getitem__`) -- see [the parent implementation](https://github.com/python/cpython/blob/3.8/Lib/_collections_abc.py#L870)

datastore/gcloud/aio/datastore/value.py
```diff
@@ -35,6 +36,8 @@ def from_repr(cls, data: Dict[str, Any]) -> 'Value':
             if json_key in data:
                 if json_key == 'nullValue':
                     value = None
+                elif value_type == Array:
```
TheKevJames: I like what you did in your other PR for merging the `value_type.from_repr` cases -- no need to change this, since it'll annoy us with merge conflicts, this comment is just a note to myself to merge these too approaches when I merge both your PRs.

wosc:
Updated thanks. I must have read the collections.abc docs wrong.

TheKevJames:
Awesome -- much appreciated! Will get this released shortly.

I've merged this along with a couple merge conflict fixes [here](https://github.com/talkiq/gcloud-aio/commit/6b7d2411146b0baa7371eefe9ab6a672a67c7501)

PR #139, opened 2020-02-05
Add support for embedded entity values

wosc:
See https://cloud.google.com/datastore/docs/concepts/entities#embedded_entity

### DIFF
--- a/datastore/gcloud/aio/datastore/constants.py
+++ b/datastore/gcloud/aio/datastore/constants.py
@@ -63,6 +63,7 @@ class TypeName(enum.Enum):
     BLOB = 'blobValue'
     BOOLEAN = 'booleanValue'
     DOUBLE = 'doubleValue'
+    ENTITY = 'entityValue'
     INTEGER = 'integerValue'
     GEOPOINT = 'geoPointValue'
     KEY = 'keyValue'
--- a/datastore/gcloud/aio/datastore/entity.py
+++ b/datastore/gcloud/aio/datastore/entity.py
@@ -26,12 +26,19 @@ class Entity:
 
     @classmethod
     def from_repr(cls, data: Dict[str, Any]) -> 'Entity':
-        return cls(cls.key_kind.from_repr(data['key']), data.get('properties'))
+        # https://cloud.google.com/datastore/docs/reference/data/rest/v1/Entity
+        # "for example, an entity in Value.entity_value may have no key"
+        if 'key' in data:
+            key = cls.key_kind.from_repr(data['key'])
+        else:
+            key = None
+        return cls(key, data.get('properties'))
 
     def to_repr(self) -> Dict[str, Any]:
         return {
-            'key': self.key.to_repr(),
-            'properties': self.properties,
+            'key': self.key.to_repr() if self.key else None,
+            'properties': {k: self.value_kind(v).to_repr()
+                           for k, v in self.properties.items()},
         }
 
 
--- a/datastore/gcloud/aio/datastore/value.py
+++ b/datastore/gcloud/aio/datastore/value.py
@@ -2,6 +2,7 @@ from datetime import datetime
 from typing import Any
 from typing import Dict
 
+from gcloud.aio.datastore import entity
 from gcloud.aio.datastore.constants import TypeName
 from gcloud.aio.datastore.constants import TYPES
 from gcloud.aio.datastore.key import Key
@@ -35,6 +36,8 @@ class Value:  # pylint:disable=useless-object-inheritance
             if json_key in data:
                 if json_key == 'nullValue':
                     value = None
+                elif value_type == entity.Entity:
+                    value = value_type.from_repr(data[json_key])
                 elif value_type == datetime:
                     value = datetime.strptime(data[json_key],
                                               '%Y-%m-%dT%H:%M:%S.%f000Z')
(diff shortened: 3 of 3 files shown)


datastore/gcloud/aio/datastore/value.py
```diff
@@ -60,7 +64,7 @@ def from_repr(cls, data: Dict[str, Any]) -> 'Value':
 
     def to_repr(self) -> Dict[str, Any]:
         value_type = self._infer_type(self.value)
-        if value_type in {TypeName.GEOPOINT, TypeName.KEY}:
+        if value_type in {TypeName.ENTITY, TypeName.GEOPOINT, TypeName.KEY}:
```
TheKevJames: Could you explain this change? I think you _might_ have meant to add `TypeName.ARRAY` here rather than `ENTITY`?

datastore/gcloud/aio/datastore/constants.py
```diff
@@ -81,4 +83,7 @@ class TypeName(enum.Enum):
     LatLng: TypeName.GEOPOINT,
     str: TypeName.STRING,
     type(None): TypeName.NULL,
+    # We should use .entity.Entity here, but that would lead to a lot of
+    # unsolveable import cycles.
+    Any: TypeName.ENTITY,
```
TheKevJames: Hmm... I see the problem here, but I'm a bit worried about adding `Any` here as a key, as a few bad type hints could lead us to mapping something as an `ENTITY` incorrectly down the line. Instead of adding anything to this `TYPES` dictionary, you could solve this issue by adding it as a late-binding in `Value._get_supported_types()`, eg. by doing:
```python
    @classmethod
    def _get_supported_types(cls):
        supported_types = TYPES
        supported_types.update({
            cls.key_kind: TypeName.KEY,
+           Entity: TypeName.ENTITY,
        })
        return supported_types
```

datastore/gcloud/aio/datastore/entity.py
```diff
@@ -26,12 +26,19 @@ def __repr__(self) -> str:
...
+        else:
+            key = None
+        return cls(key, data.get('properties'))
 
     def to_repr(self) -> Dict[str, Any]:
         return {
-            'key': self.key.to_repr(),
-            'properties': self.properties,
+            'key': self.key.to_repr() if self.key else None,
+            'properties': {k: self.value_kind(v).to_repr()
```
TheKevJames: Nice catch -- I completely missed this bug. Note to self: looks like we'll need to mark this as a major version bump, as this will be a non-backwards-compatible fix.

datastore/gcloud/aio/datastore/value.py
```diff
@@ -2,6 +2,7 @@
 from typing import Any
 from typing import Dict
 
+import gcloud.aio.datastore.entity
```
TheKevJames: Please write this as `from gcloud.aio.datastore.entity import Entity` (and my bad -- our linter should have caught this but seems to be mis-configured).

datastore/gcloud/aio/datastore/value.py
```diff
@@ -74,6 +78,10 @@ def to_repr(self) -> Dict[str, Any]:
     def _infer_type(self, value: Any) -> TypeName:
         odd_strings = {'newstr', 'unicode'}
         kind = str if value.__class__.__name__ in odd_strings else type(value)
+        # We should instead use Entity as the key in TYPES directly, but that
```
TheKevJames: As mentioned above, applying this change to `_get_supported_types()` directly would be a bit nicer -- I believe that should eliminate the need for this check.

datastore/gcloud/aio/datastore/constants.py
wosc: Hi, thanks for your thoughtful comments, I've adjusted this per your suggestion.

datastore/gcloud/aio/datastore/value.py
wosc: I'm afraid that's not possible, because it would create yet another import cycle: entity.py already has `from .value import Value`, so we cannot put `from .entity import Entity` into value.py.

TheKevJames: Ah, gotcha, I missed that. Let's go with `from gcloud.aio.datastore import entity` then -- we're mostly trying to avoid having the `google.aio.datastore` namespace scattered through code here. Tested that and it appears to run with no import cycle!

datastore/gcloud/aio/datastore/value.py
```diff
@@ -38,10 +39,10 @@ def from_repr(cls, data: Dict[str, Any]) -> 'Value':
                 elif value_type == datetime:
                     value = datetime.strptime(data[json_key],
                                               '%Y-%m-%dT%H:%M:%S.%f000Z')
-                elif value_type == cls.key_kind:
-                    value = cls.key_kind.from_repr(data[json_key])
-                elif value_type == LatLng:
-                    value = LatLng.from_repr(data[json_key])
+                elif value_type in [cls.key_kind,
```
TheKevJames: Please use a set here rather than a list to match style with the rest of this project.

wosc:
Ah, from-importing the module does indeed work just fine. Updated.

TheKevJames:
Awesome, this looks pretty great to me! I think that new import style I asked you to invoke is causing some issues on Python 2.7.. I'll merge this in and then figure out a fix.

Really appreciate the contribution(s)! I'll get this and your other PR all merged in and released shortly.

PR #147, opened 2020-03-17
Add parametrization to load job, add query method

DaniilAnichin:
1) Make bigquery.Table.load method accept source format, autodetect & write disposition parameters
2) Add bigquery.Table.query method to create table from a query result

### DIFF
--- a/bigquery/gcloud/aio/bigquery/bigquery.py
+++ b/bigquery/gcloud/aio/bigquery/bigquery.py
@@ -1,6 +1,7 @@
 import io
 import json
 import uuid
+from enum import Enum
 from typing import Any
 from typing import Callable
 from typing import Dict
@@ -24,6 +25,21 @@ SCOPES = [
     'https://www.googleapis.com/auth/bigquery.insertdata',
     'https://www.googleapis.com/auth/bigquery',
 ]
+
+
+class SourceFormat(Enum):
+    AVRO = 'AVRO'
+    CSV = 'CSV'
+    DATASTORE_BACKUP = 'DATASTORE_BACKUP'
+    NEWLINE_DELIMITED_JSON = 'NEWLINE_DELIMITED_JSON'
+    ORC = 'ORC'
+    PARQUET = 'PARQUET'
+
+
+class Disposition(Enum):
+    WRITE_APPEND = 'WRITE_APPEND'
+    WRITE_EMPTY = 'WRITE_EMPTY'
+    WRITE_TRUNCATE = 'WRITE_TRUNCATE'
 
 
 class Table:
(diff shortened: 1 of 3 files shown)


DaniilAnichin:
Has pubsub key expired? Seems like I changed nothing in the way that test works

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -233,16 +268,21 @@ def _make_load_body(
...
+            source_uris, project, autodetect, source_format, write_disposition,
+        )
+        return await self._post_json(url, body, session, timeout)
 
-        headers = await self.headers()
-        headers.update({
-            'Content-Length': str(len(payload)),
-            'Content-Type': 'application/json',
-        })
+    async def query(
```
TheKevJames: I think this might get a bit confusing, since this is not a "bigquery query" but rather a "bigquery insert via query" -- looks like I didn't RTF very well when picking names. Could you please rename `load` to `insert_via_load`, `copy` to `insert_via_copy`, and this method to `insert_via_query`? Open to other naming schemes, of course, just want to make sure this doesn't confuse folks who are looking for the ability to actually run a query!

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -233,16 +268,21 @@ def _make_load_body(
...
+            self, query: str, session: Optional[Session] = None,
+            write_disposition: Disposition = Disposition.WRITE_TRUNCATE,
```
TheKevJames: Looks like Google sets the default to `WRITE_EMPTY` -- let's make sure our default matches theirs to avoid any confusion.

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -209,22 +252,14 @@ def _make_load_body(
...
-        resp = await s.post(url, data=payload, headers=headers, params=None,
-                            timeout=timeout)
-        return await resp.json()
+        return await self._post_json(url, body, session, timeout)
 
     # https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/insert
     async def load(
             self, source_uris: List[str], session: Optional[Session] = None,
+            autodetect: bool = True,
+            source_format: SourceFormat = SourceFormat.DATASTORE_BACKUP,
```
TheKevJames: Please set the default to `CSV` to match the Google API.

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -209,22 +252,14 @@ def _make_load_body(
...
+            write_disposition: Disposition = Disposition.WRITE_TRUNCATE,
```
TheKevJames: Default to `WRITE_EMPTY` please, to match Google's API.

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -209,22 +252,14 @@ def _make_load_body(
...
-        s = AioSession(session) if session else self.session
-        resp = await s.post(url, data=payload, headers=headers, params=None,
-                            timeout=timeout)
-        return await resp.json()
+        return await self._post_json(url, body, session, timeout)
 
     # https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/insert
     async def load(
             self, source_uris: List[str], session: Optional[Session] = None,
+            autodetect: bool = True,
```
TheKevJames: I can't tell for sure from Google's docs, but it seems this defaults to False in the API -- let's match that setting.

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -209,22 +252,14 @@ def _make_load_body(
...
-            'Content-Type': 'application/json',
-        })
-
-        s = AioSession(session) if session else self.session
-        resp = await s.post(url, data=payload, headers=headers, params=None,
-                            timeout=timeout)
-        return await resp.json()
+        return await self._post_json(url, body, session, timeout)
 
     # https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/insert
```
TheKevJames: Would you please add a comment linking to `https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfigurationLoad` on the next line?

bigquery/gcloud/aio/bigquery/bigquery.py
TheKevJames: If you could please also add the following comment above this method definition, that'd help us track things down in the future:
```
# https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/insert
# https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfigurationQuery
```

DaniilAnichin:
Sorry for late reply, I tried to fix all comments you made & rebase to the current master; seems like one test is still failing, though.

PR #149, opened 2020-03-23
fix(datastore): allow microsecond resolution timestamp values

snstanton:
Slice the date string to drop the nanoseconds and trailing 'Z'.

### DIFF
--- a/datastore/gcloud/aio/datastore/value.py
+++ b/datastore/gcloud/aio/datastore/value.py
@@ -35,8 +35,8 @@ class Value:  # pylint:disable=useless-object-inheritance
                 if json_key == 'nullValue':
                     value = None
                 elif value_type == datetime:
-                    value = datetime.strptime(data[json_key],
-                                              '%Y-%m-%dT%H:%M:%S.%f000Z')
+                    value = datetime.strptime(data[json_key][:-4],
+                                              '%Y-%m-%dT%H:%M:%S.%f')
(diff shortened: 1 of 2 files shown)


snstanton:
I didn't change anything in test-pubsub, so I'm reasonably confident this failure has nothing to do with my change.

snstanton:
Hmm.  I'm finding more corner cases, so I'm going to close this request until I get more test coverage.

TheKevJames:
@snstanton looking forward to seeing this, if you figure out those edge-cases :) It would be great to get that hooked into the integration tests, as well, since it looks like this will depend heavily on the Google API responses!

wrt the Pubsub tests, it looks like the guard against running that test for external contributors is still having issues -- you can go ahead and ignore that. Theoretically, [this block](https://github.com/talkiq/gcloud-aio/blob/6a120024e8bca0f18ad8c404b89a637705e3fdac/pubsub/noxfile.py#L7-L31) is supposed to be `session.skip`ing those tests (since we don't yet have GCP creds that we can share with external contributors), but it looks like I'll need to rework that...

snstanton:
Ok, I think I've got the additional cases sorted out.  It turns out all of msec, usec, and nsec were used at one time or another.  I guess that's what happens when you've been using datastore since the earliest days of appengine.

I'm not sure how to run the integration tests, so I'm hesitant to add anything there.

PR #152, opened 2020-03-30
Add close of session

elysium31:


TheKevJames:
Oh -- and could you also please add type annotations to each of these new methods? Looks like many of them return nothing (eg. `-> None`) and the enter method returns an instance of the class (eg. `-> 'Storage'`).

Issue #140, opened 2020-02-06
Add support for pagination in Storage APIs


MarkEEaton:
Currently, `list_blobs()` returns a maximum of 1000 blobs from a bucket.

Desired behavior: that `list_blobs()` iterate through subsequent pages of blobs to return all of the blobs in a bucket, even when there are more than 1000.

Maybe there is some way to paginate when I call `list_blobs()` that I am missing?

Here is the code that I am running to produce this issue:

    import settings
    import asyncio
    from gcloud.aio.storage import Storage
    from aiohttp import ClientSession as Session

    async def test():
        async with Session() as session:
            storage = Storage(session=session)
            bucket = storage.get_bucket(settings.bucket_name)
            blobs = await bucket.list_blobs()
            return len(blobs)  # 1000

    asyncio.run(test())

MarkEEaton:
I've provisionally solved this for myself by implementing pagination in `list_blobs()`. It's not the best solution because I haven't accounted for corner cases, the code is ugly, and the pagination  behavior is hardcoded, rather than customizable via params, but it works for me for now. 

For guidance, I followed the docs at https://cloud.google.com/storage/docs/json_api/v1/objects/list.

TheKevJames:
Unfortunately, the current `gcloud-aio-storage` does not support pagination. We would absolutely love to get this added, though -- if you (or anyone else reading this) would like to submit a pull request, I'd be happy to help get it merged and released. Even hardcoded and ugly code could be a great start in getting this added :)

(sorry for the late response on this -- I seem to have disabled notifications on new issues on this repo; should be fixed now!)

MarkEEaton:
Sure. I will try to clean up my code and submit a PR (hopefully tomorrow). I'm glad you think this would be a good addition!

PR #153, opened 2020-03-31
feat(storage) add pagination to list_blobs()

MarkEEaton:
This is a solution for #140 

I know that the code in this PR is not as nice as the rest of the gcloud-aio repository, but I appreciate you encouraging my submission and having patience with a newbie. I value your feedback.

### DIFF
--- a/storage/gcloud/aio/storage/bucket.py
+++ b/storage/gcloud/aio/storage/bucket.py
@@ -46,20 +46,19 @@ class Bucket:
 
     async def list_blobs(self, prefix: str = '',
                          session: Optional[Session] = None) -> List[str]:
-        params = {'prefix': prefix}
-        content = await self.storage.list_objects(self.name,
-                                                  params=params,
-                                                  session=session)
-        next_content = {}
-        next_content['nextPageToken'] = content.get('nextPageToken')
-        while next_content.get('nextPageToken'):
-            params['pageToken'] = next_content['nextPageToken']
-            next_content = await self.storage.list_objects(self.name,
-                                                           params=params,
-                                                           session=session)
-            content['items'] = content['items']\
-                    + next_content.get('items', list())
-        return [x['name'] for x in content.get('items', list())]
+        params = {'prefix': prefix, 'pageToken': ''}
+        items = []
+        while True:
+            content = await self.storage.list_objects(self.name,
+                                                      params=params,
+                                                      session=session)
+            items.extend([x['name'] for x in content.get('items', list())])
+
+            params['pageToken'] = content.get('nextPageToken')
+            if not params['pageToken']:
+                break
+
+        return items
 
 
     def new_blob(self, blob_name: str) -> Blob:
--- a/storage/gcloud/aio/storage/bucket.py
+++ b/storage/gcloud/aio/storage/bucket.py
@@ -47,10 +47,20 @@ class Bucket:
     async def list_blobs(self, prefix: str = '',
                          session: Optional[Session] = None) -> List[str]:
         params = {'prefix': prefix}
-        content = await self.storage.list_objects(self.name, params=params,
+        content = await self.storage.list_objects(self.name,
+                                                  params=params,
                                                   session=session)
-
+        next_content = {}
+        next_content['nextPageToken'] = content.get('nextPageToken')
+        while next_content.get('nextPageToken'):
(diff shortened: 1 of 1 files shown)


storage/gcloud/aio/storage/bucket.py
```diff
@@ -47,10 +47,20 @@ def __init__(self, storage, name: str) -> None:
...
                                                   session=session)
-
+        next_content = {}
+        next_content['nextPageToken'] = content.get('nextPageToken')
+        while next_content.get('nextPageToken'):
+            params['pageToken'] = next_content['nextPageToken']
+            next_content = await self.storage.list_objects(self.name,
+                                                           params=params,
+                                                           session=session)
+            content['items'] = content['items']\
```
TheKevJames, line 60: General pro-tip here: `lists` support `append` and `extend` methods, which are generally preferred as being a bit more readable then using the arithmetic operators. In this case, what you've written here could instead be formulated as:
```python
content['items'].extend(next_content.get('items', list()))
```

storage/gcloud/aio/storage/bucket.py
```diff
@@ -47,10 +47,20 @@ def __init__(self, storage, name: str) -> None:
     async def list_blobs(self, prefix: str = '',
                          session: Optional[Session] = None) -> List[str]:
         params = {'prefix': prefix}
-        content = await self.storage.list_objects(self.name, params=params,
+        content = await self.storage.list_objects(self.name,
```
TheKevJames, line 50: I find that whenever you find yourself writing a loop like this where the first iteration needs to be treated specially, it's often the case that restructuring your code to avoid the duplication can lead to a much cleaner result. In this case, there's a couple potential cleanups you can make which will come together into a nice overall result:
* passing a param set to an empty string will act as if you hadn't sent that param at all. So on your first iteration you could have `'pageToken': ''`, which means you don't have to worry about selective in passing in / modifying your params only on the second and future iterations
* you do some juggling to make sure `content['items']` contains the full list -- in this case, it's easier to just pull the data out of `content` and into a collector after each API request (which could also be something you could do for all iterations rather than just the ones after the first!)

Pulling that all together, you could end up with something like this:
```python
params = {'prefix': prefix, 'pageToken': ''}
items = []
while True:
    content = await self.storage.list_objects(self.name,
                                              params=params,
                                              session=session)
    items.extend([x['name'] for x in content.get('items', list())])

    params['pageToken'] = content.get('nextPageToken')
    if not params['pageToken']:
        break

return items
```

storage/gcloud/aio/storage/bucket.py
```diff
@@ -47,10 +47,20 @@ def __init__(self, storage, name: str) -> None:
...
+                                                  params=params,
                                                   session=session)
-
+        next_content = {}
```
TheKevJames, line 53: You do a lot of juggling here with `content` and `new_content` to make sure they're formulated the right way with the right data at the right times -- as a general rule, when you find yourself mutating the same values over and over, you may be able to find a better datatype for the job. I've added a comment above which describes how you could eliminate this `next_content` variable entirely.

For interest's sake, you could also just use a simple value here; if you instead just did:
```python
next_token = content.get('nextPageToken')
while next_token:
    params['pageToken'] = next_token
    # SNIP ...
    next_token = next_content['nextPageToken']
```
then you would have a bit simpler scope -- the `next_token` value would be explicitly used as your list iterator, but `next_content` can stay as only in scope within each iteration of your list: the interesting thing there is that fewer variables needed to be juggled means your code is easier to understand; no one outside of the loop needs to care what `next_content['foobar']` is set to.

The other big benefit is that mini-refactors like this can pave the way for even bigger&better changes; in the above code snippet, now that we've replaced `next_content` with `next_token`, it might become clear that this value only ever interacts with `params['pageToken']` -- in fact, you can get away entirely without having `next_token` at all and just using `params['pageToken']` entirely! You can see how I used that trick in my above full code block to shrink the solution a bit :)

storage/gcloud/aio/storage/bucket.py
MarkEEaton: This makes a lot of sense! It looks much nicer than what I had written. Thank you for taking the time to give all of this feedback!

MarkEEaton:
Thanks again for all of your guidance on this!

envil:
This feature comes just in time, working good for me. Thanks! 👍

PR #154, opened 2020-04-02
Method for bucket meta

elysium31:


storage/gcloud/aio/storage/storage.py
```diff
@@ -365,6 +365,21 @@ def _split_content_type(content_type: str) -> Tuple[str, str]:
         data: dict = await resp.json()
         return data
 
+    async def get_bucket_meta(self, bucket: str, *, params: dict = None,
```
TheKevJames, line 368: nit: could you please name this `get_bucket_metadata`? We use the full `metadata` name elsewhere on this class and it would be great to be consistent.

PR #157, opened 2020-04-14
fix(datastore): handle additional legacy date formats

snstanton:
I found one more corner case: datetimes with second resolution.


snstanton:
Can you bump the datastore version and push a release with this fix?

TheKevJames:
@snstanton done and released as v6.1.0 -- sorry, this completely slipped my mind!

Issue #162, opened 2020-05-12
Include offset and limit in Query repr even if they are 0


KevinGDialpad:
These should check for None instead:
https://github.com/talkiq/gcloud-aio/blob/abe2ea915c342df06b69d439b2138b270e574eeb/datastore/gcloud/aio/datastore/query.py#L87-L90

TheKevJames:
Solved in #164 and released as `gcloud-{aio,rest}-datastore==7.0.0`!

PR #164, opened 2020-05-13
fix(datastore): avoid mangling None queryvals

TheKevJames:
The Datastore API treats `None` and `0` differently for `offset` and
`limit` and sets a default of `None` -- this changeset ensures we aren't
overriding their defaults and can support both values as intended.

### DIFF
--- a/datastore/gcloud/aio/datastore/query.py
+++ b/datastore/gcloud/aio/datastore/query.py
@@ -1,6 +1,7 @@
 from typing import Any
 from typing import Dict
 from typing import List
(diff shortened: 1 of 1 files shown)


TheKevJames:
@KevinGDialpad this should solve #162 -- could you please verify it works for the use-case you were running into?

danadialpad:
> @KevinGDialpad this should solve #162 -- could you please verify it works for the use-case you were running into?

I verified it in our tests. Thanks for fixing!

Issue #184, opened 2020-07-30
get_distribution('gcloud-aio-storage').version call breaks cloud function


unovongalixor:
i've been using gcloud-aio-storage in a cloud function to handle parallel uploads with great success, but recently tried deploying and found the cloud function wouldn't deploy with gcloud-aio-storage as a dependency in requirements.txt.

you can verify this by attempting to deploy a hello world example with gcloud-aio-storage in the requirements file.

in my project i've vendorized gcloud-aio-storage and gcloud-aio-auth and hardcoded the versions in both of their __init__.py files

eddiedialpad:
Hello! We had the same problem -- there's a shared dependency between gcloud-aio-storage, and the internal google cloud function worker code -- we both require aiohttp -- who then requires yarl.

The yarl package got an update four days ago that's introduces a change in a function signature -- I'm in the process of making a PR to pin a good version of yarl here -- and also to message aiohttp and yarl about the incompatibility.

For now -- add this to your `requirements.txt` --

`yarl==1.4.2`

TheKevJames:
@unovongalixor the fix @eddiedialpad described should now be released in `gcloud-aio-auth==3.4.1`; updating should do the trick. Let us know if you find any other problems!

eddiedialpad:
@unovongalixor -- I'm planning to undo the yarl version pin in gcloud-aio shortly, since further reading revealed that it's Google Cloud Functions that has a version pin to 1.4.2 -- their doc here lists their version pins: https://cloud.google.com/functions/docs/writing/specifying-dependencies-python

PR #189, opened 2020-08-07
Revert "chore(deps): pin yarl for google cloud function compat (#185)"

eddiedialpad:
This reverts commit 099cb35ed1c074ac4b226ae73e923814ddc6c927.

We determined that the yarl incompatibility only affects Google Cloud Functions (GCF), and is known to be pinned there at 1.4.2 -- we recommend users who use GCF to pin according to their docs here: https://cloud.google.com/functions/docs/writing/specifying-dependencies-python

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -26,6 +26,20 @@ Installation
 
     $ pip install --upgrade gcloud-{aio,rest}-{client_name}
 
+Compatibility
+-------------
+
+These libraries are used with Google projects. Here are notes on compatibility
+issues. While we cannot offer specific support for issues originating from
+other projects, we can point toward known resolutions.
+
+- Google Cloud Functions pins ``yarl``; ``gcloud-aio-*`` indirectly requires
+  ``yarl`` via ``aiohttp`` and an unpinned version of ``yarl`` can cause your
+  cloud functions to stop building. Please pin your requirements as described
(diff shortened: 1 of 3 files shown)


eddiedialpad:
Left a comment in https://github.com/talkiq/gcloud-aio/issues/184 -- to refer author to google cloud function docs re manual pinning yarl.

eddiedialpad:
I just added a brief blurb in the README.rst -- pls let me know if this belongs here + any editorial notes you have.

README.rst
```diff
@@ -26,6 +26,20 @@ Installation
 
     $ pip install --upgrade gcloud-{aio,rest}-{client_name}
 
+Compatibility
+-------------
+
+These libraries are used with Google projects. Here are notes on compatibility
```
harsh-dialpad, line 32: nit: `These libraries are used with Google projects.`
I don't think there is any need for this statement ~ its implied that these libraries work with GCP projects.

The blurb otherwise looks okay to me.

eddiedialpad:
Hello @TheKevJames -- it looks like upgrading the `python_version` for the `tester/pipcheck` steps in `circleci` allows the checks to run. Pls lemme know if the pinned version is OK (3.7.8) -- I based this version on the `tester` orb.

Issue #195, opened 2020-08-21
GCS download range


domZippilli:
GCS supports a Range header on object downloads. You can see this in the Google Cloud SDK's GCS download implementation [here](https://github.com/googleapis/python-storage/blob/master/google/cloud/storage/blob.py#L893), namely the `start` and `end` arguments.

This is useful for surgically downloading only a piece of a structured file, or more commonly doing multiple TCP streams for a download of a single large file.

It would be great to have this capability in gcloud-aio's download.

PR #200, opened 2020-08-29
Add header overrides to download function

domZippilli:
Addresses #195 and other cases (toggling gzip transcoding, for example) by allowing the caller to supply a dictionary of headers. These headers are applied over the default session headers.

With this update, users can download ranges as follows:

```
await client.download(bucket,
                      blob,
                      headers={"Range": f"bytes={start}-{end}"})
```

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -149,9 +149,11 @@ class Storage:
         return data
 
     async def download(self, bucket: str, object_name: str, *,
+                       headers: dict = {},
                        timeout: int = 10,
                        session: Optional[Session] = None) -> bytes:
-        return await self._download(bucket, object_name, timeout=timeout,
+        return await self._download(bucket, object_name, 
+                                    override_headers=headers, timeout=timeout,
                                     params={'alt': 'media'}, session=session)
 
     async def download_to_filename(self, bucket: str, object_name: str,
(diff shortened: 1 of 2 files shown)


storage/gcloud/aio/storage/storage.py
```diff
@@ -149,9 +149,11 @@ async def delete(self, bucket: str, object_name: str, *,
         return data
 
     async def download(self, bucket: str, object_name: str, *,
+                       headers: dict = {},
```
TheKevJames, line 152: Please have the default here be `None`. Using a mutable object as a default value can lead to some [unexpected behaviour in Python](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) and should generally be avoided.

TheKevJames (changes requested):
Thanks for the contribution! Overall, this looks good -- it'll be great to allow for custom header usage here :)

TheKevJames:
Fixed up and merged as 640a2dd3448294bb3ab5801f5328e7c2e7a64854. Will be included in `gcloud-*-storage==5.7.0`

domZippilli:
Thank you @TheKevJames! Shortly after making this PR, we had a baby that arrived a bit early and I've been on leave a while... I lost this in the inbox deluge while I was away.

TheKevJames:
No problem at all -- and congratulations!

PR #201, opened 2020-09-03
feat(gcloud-aio-pubsub): start migrate subscriber to aio, wrap Message, FlowControl, nix Scheduler

eddiedialpad:
Add a layer of indirection for Google's Message class (`google.cloud.pubsub_v1.subscriber.message.Message`) with (`gcloud.aio.pubsub.subscriber_message.Message`);

Provide our own FlowControl class

Try removing scheduler --

This allows us to change the implementation details of our Message without the user having to change their usage of it, so that migration to full gcloud-aio can proceed underneath.

FlowControl is just a named tuple, so implementing it here means the user can stop importing it from google-cloud.

Finally, the Scheduler argument doesn't really make sense if we're migrating to fully asyncio since it allows the user to submit a custom queue for retrieving messages -- something that really makes more sense in a multi-thread context.

Steps ...
• ✅ evaluate if this is ergonomic
• ✅ stage this lib so we can tell if I've gotten wrapping the callbacks correct
• ✅ stage this by injecting it into an app that uses the named Subscriber features.
• ✅ copy changes learned from injection staging, and stage here one more time using example code.

### DIFF
--- a/pubsub/gcloud/aio/pubsub/__init__.py
+++ b/pubsub/gcloud/aio/pubsub/__init__.py
@@ -4,7 +4,7 @@ __version__ = get_distribution('gcloud-aio-pubsub').version
 from gcloud.aio.pubsub.publisher_client import PublisherClient
 from gcloud.aio.pubsub.subscriber_client import SubscriberClient
 from gcloud.aio.pubsub.utils import PubsubMessage
-from gcloud.aio.pubsub.subscriber_message import FlowControl
+from gcloud.aio.pubsub.subscriber_client import FlowControl
 from gcloud.aio.pubsub.subscriber_message import SubscriberMessage
 
 
--- a/pubsub/gcloud/aio/pubsub/__init__.py
+++ b/pubsub/gcloud/aio/pubsub/__init__.py
@@ -4,8 +4,9 @@ __version__ = get_distribution('gcloud-aio-pubsub').version
 from gcloud.aio.pubsub.publisher_client import PublisherClient
 from gcloud.aio.pubsub.subscriber_client import SubscriberClient
 from gcloud.aio.pubsub.utils import PubsubMessage
-from gcloud.aio.pubsub.subscriber_message import Message
+from gcloud.aio.pubsub.subscriber_message import FlowControl
+from gcloud.aio.pubsub.subscriber_message import SubscriberMessage
 
 
-__all__ = ['__version__', 'PublisherClient', 'Message', 'PubsubMessage',
-           'SubscriberClient']
+__all__ = ['__version__', 'FlowControl', 'PublisherClient', 'PubsubMessage',
+           'SubscriberClient', 'SubscriberMessage']
--- a/pubsub/gcloud/aio/pubsub/__init__.py
+++ b/pubsub/gcloud/aio/pubsub/__init__.py
@@ -4,7 +4,8 @@ __version__ = get_distribution('gcloud-aio-pubsub').version
 from gcloud.aio.pubsub.publisher_client import PublisherClient
 from gcloud.aio.pubsub.subscriber_client import SubscriberClient
 from gcloud.aio.pubsub.utils import PubsubMessage
+from gcloud.aio.pubsub.subscriber_message import Message
 
 
-__all__ = ['__version__', 'PublisherClient', 'PubsubMessage',
+__all__ = ['__version__', 'PublisherClient', 'Message', 'PubsubMessage',
            'SubscriberClient']
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -14,10 +14,12 @@ else:
 
     from google.api_core import exceptions
     from google.cloud import pubsub
-    from google.cloud.pubsub_v1.subscriber.message import Message
+    from google.cloud.pubsub_v1.subscriber.message import Message \
+        as GoogleMessage
     from google.cloud.pubsub_v1.subscriber.scheduler import Scheduler
     from google.cloud.pubsub_v1.types import FlowControl
 
+    from .message import Message
     from .utils import convert_google_future_to_concurrent_future
 
 
@@ -101,9 +103,10 @@ else:
 
         def _wrap_callback(self,
                            callback: Callable[[Message], None]
-                           ) -> Callable[[Message], None]:
+                           ) -> Callable[[GoogleMessage], None]:
             """Schedule callback to be called from the event loop"""
             def _callback_wrapper(message: Message) -> None:
-                asyncio.run_coroutine_threadsafe(callback(message), self.loop)
+                asyncio.run_coroutine_threadsafe(
+                    callback(message.google_message), self.loop)
 
             return _callback_wrapper
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -25,12 +25,13 @@ else:
     from .utils import convert_google_future_to_concurrent_future
 
 
-    class FlowControl:
+    class FlowControl(tuple):
         def __init__(self,
                      max_bytes: int = 100 * 1024 * 1024,
                      max_messages: int = 1000,
                      max_lease_duration: int = 1 * 60 * 60,
                      max_duration_per_lease_extension: int = 0) -> None:
+            super().__init__()
             self.max_bytes = max_bytes
             self.max_messages = max_messages
             self.max_lease_duration = max_lease_duration
@@ -118,8 +119,9 @@ else:
                            callback: Callable[[SubscriberMessage], None]
                            ) -> Callable[[Message], None]:
             """Schedule callback to be called from the event loop"""
-            def _callback_wrapper(message: SubscriberMessage) -> None:
+            def _callback_wrapper(message: Message) -> None:
                 asyncio.run_coroutine_threadsafe(
-                    callback(message.google_message), self.loop)
+                    callback(SubscriberMessage.from_google_cloud(message)),
+                    self.loop)
 
             return _callback_wrapper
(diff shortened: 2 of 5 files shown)


pubsub/gcloud/aio/pubsub/__init__.py
```diff
@@ -4,7 +4,8 @@
 from gcloud.aio.pubsub.publisher_client import PublisherClient
 from gcloud.aio.pubsub.subscriber_client import SubscriberClient
 from gcloud.aio.pubsub.utils import PubsubMessage
+from gcloud.aio.pubsub.subscriber_message import Message
 
 
-__all__ = ['__version__', 'PublisherClient', 'PubsubMessage',
+__all__ = ['__version__', 'PublisherClient', 'Message', 'PubsubMessage',
```
TheKevJames, line 10: I wonder if we could reuse/augment the existing `PubsubMessage`? Or maybe that's getting ahead of ourselves at this point?

Perhaps `SubscriberMessage` might be more clear for now?

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -61,8 +61,7 @@ def subscribe(self,
             sub_keepalive: asyncio.Future = self._subscriber.subscribe(
                 subscription,
                 self._wrap_callback(callback),
-                flow_control=flow_control,
-                scheduler=scheduler
+                flow_control=flow_control
```
TheKevJames, line 64: Why the removed arg?

eddiedialpad: If we're migrating to a fully asyncio version of this, that means that the scheduler doesn't make sense here anymore. At its core, this creates a thread which manages a queue to process messages; we'll be using the asyncio event loop for that functionality instead. The user would thus choose a different event loop long before they'd touch our constructors.

pubsub/gcloud/aio/pubsub/__init__.py
eddiedialpad: Note to self: check if this is semantically the same message class as per google cloud API.

eddiedialpad:
Okay, I staged this using our staging environment with a temporary pubsub subscription -- I discovered my wrapping for the Google Cloud Message type was inside-out [fixed], and that FlowControl needs to be iterable [fixed].

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -15,12 +20,30 @@ def __init__(self, **kwargs) -> None:
...
+        def __init__(self,
+                     max_bytes: int = 100 * 1024 * 1024,
+                     max_messages: int = 1000,
+                     max_lease_duration: int = 1 * 60 * 60,
+                     max_duration_per_lease_extension: int = 0) -> None:
+            self.max_bytes = max_bytes
+            self.max_messages = max_messages
+            self.max_lease_duration = max_lease_duration
+            self.max_duration_per_lease_extension = (
+                max_duration_per_lease_extension)
```
jonathan-johnston, line 38: Could you add a bit of documentation to describe each of the fields briefly?

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -15,12 +20,30 @@ def __init__(self, **kwargs) -> None:
...
     from google.cloud import pubsub
     from google.cloud.pubsub_v1.subscriber.message import Message
-    from google.cloud.pubsub_v1.subscriber.scheduler import Scheduler
-    from google.cloud.pubsub_v1.types import FlowControl
 
+    from .subscriber_message import SubscriberMessage
     from .utils import convert_google_future_to_concurrent_future
 
 
+    class FlowControl():
```
jonathan-johnston, line 28: You don't need the `()` here.

jonathan-johnston (approved):
Overall this looks great, @eddiedialpad! I would like to see if @TheKevJames has any other comments on this as well as see how this integrates into our workflow. I noticed that you've already done some testing on this front that I'd love to see!

pubsub/gcloud/aio/pubsub/subscriber_client.py
TheKevJames: Agreed -- might be nice to link to the Google code for this class, at least, since IIRC they have some light docs on the meaning of each of these fields inline.

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -15,12 +20,30 @@ def __init__(self, **kwargs) -> None:
...
+                     max_messages: int = 1000,
+                     max_lease_duration: int = 1 * 60 * 60,
+                     max_duration_per_lease_extension: int = 0) -> None:
+            self.max_bytes = max_bytes
+            self.max_messages = max_messages
+            self.max_lease_duration = max_lease_duration
+            self.max_duration_per_lease_extension = (
+                max_duration_per_lease_extension)
+
+        def __getitem__(self, index: int):
```
TheKevJames, line 40: Is this a feature we're using and/or want to expose? This seems a bit instructable to me, eg. perhaps it's something which the Google FlowControl construct accidentally exposes rather than an intentional part of the API?

Fine to keep it in as a transition, if we need it to help migrate towards doing this ourselves, but otherwise probably best to stick with a more clear access-by-name interface.

TheKevJames (approved):
Also agreed with Jon's outstanding comments!

eddiedialpad:
> Overall this looks great, @eddiedialpad! I would like to see if @TheKevJames has any other comments on this as well as see how this integrates into our workflow. I noticed that you've already done some testing on this front that I'd love to see!

Hai Jon! Okay -- so I _think_ the easiest way to do this is for me to make another draft branch that contains my test driver code? Lemme try that.

eddiedialpad:
Okay, so this was my test code -- 

```
import json
import random
from typing import Any
from typing import Dict
from typing import Optional
from typing import List
from typing import Tuple

from gcloud.aio.pubsub.subscriber_client import SubscriberClient
from gcloud.aio.pubsub.subscriber_client import FlowControl
from gcloud.aio.pubsub.subscriber_message import SubscriberMessage


def main():
    subscriber = SubscriberClient()
    async def super_awesome_callback(message: SubscriberMessage):
        print(
            message.publish_time.timestamp(),
            message.message_id,
            message.attributes,
            message.ordering_key,
            message.size,
            message.ack_id,
            message.delivery_attempt,
        )

        # What's up with this next block?!
        # - started as only message.ack()
        # - then added other pass thrus one-by-one
        # - so what you're seeing is a cumulation of a few iterations
        # - verified that nack() retrying a given message
        # - this ended up behaving the same as modify_ack_deadline(0)
        # - messages that go through .ack() adn .drop() do not reappear

        if random.random() < .1:
            message.modify_ack_deadline(0)
        message.ack() if random.random() < .8 else (
            message.nack() if random.random() < .5 else message.drop())

    future = subscriber.subscribe(
        STAGE_SUB,
        super_awesome_callback,
        flow_control=FlowControl(),
    )

    subscriber.run_forever(future)

if __name__ == '__main__':
    main()
```

eddiedialpad:
Okay -- injection into a real app succeeded. Need to wrap up here before merging and take care of the comments here.

pubsub/gcloud/aio/pubsub/__init__.py
eddiedialpad: Okay, so our implementation may eventually make these the same type.

However, a cursory look suggests that Google Cloud Pubsub _started off_ with publish-side and subscribe-side messages being the same type [from the protocol buffers], and eventually differentiated into different messages.

From the Publisher Client, the PubsubMessage protobuf wraps a binary blob with all the metadata needed to route it via gapic+gRPC -- https://github.com/googleapis/python-pubsub/blob/master/google/cloud/pubsub_v1/publisher/client.py#L357

From the Subscriber Client, the Subscriber.Message python object represents a parsed message, and makes it ready for user code to consume natively (blob -> attributes, data, meta) -- https://github.com/googleapis/python-pubsub/blob/master/google/cloud/pubsub_v1/subscriber/message.py#L53

Okay -- so, having said all of that -- _can we make both messages the same class?_ -- Maybe. But the similarity between the two may be an illusion. Really, they're symmetrical but have opposite orientations as the directionality converting either {python representation [or binary blob]} to|from {form compatible w gapic+gRPC|aiohttp+REST} matters.

Hmm ... however, having a symmetrical message class could also mean we would be able to use all the utilities to construct+parse messages into+outof either Python or API form ... which I suppose could serve some _retrying_ usecases [eg publish failure? np, poke inside message object, proofread+edit, publish again] and is _more functionality than is supported by native Google Cloud API_ ...

(However going from subscriber message blob -> blob for resubmission back into a topic would ofc be an anti-pattern heeh.)

Okay -- answer _for this PR_ -- no, let's not do that right now.

In future: I think _we literally have a deadlettering usecase_ that _might benefit_ from a symmetrical message class (I say "might" because we're nacking subscriber messages due to a downstream failure -- rather than some failure within the current app, so we're actually on the wrong side of the retry here). Let's take a look at how much more we can do with the existing deadlettering facilities, as well as bearing symmetrical messages in mind, as we work through our stack (deferred decision).

pubsub/gcloud/aio/pubsub/subscriber_client.py
eddiedialpad: Okay -- so -- I think _not wrapping_ Google's class is more problematic than just wrapping it -- so my solution here is twofold: (1) wrap the class with a local passthru; (2) link to the docs for the valid fields; (3) as a convenience, list the valid fields and default values;

pubsub/gcloud/aio/pubsub/subscriber_client.py
eddiedialpad: Hah, sorry -- this is to obey the python - collections - sequence - interface _just enough_ that all expectations for what constitutes a tuple can be satisfied.

TheKevJames (approved):

eddiedialpad:
Okay, I'll leave it to y'all to deploy this changeset then :D

jonathan-johnston:
Will do @eddiedialpad! I'll do the rollout + release notes

Issue #206, opened 2020-09-17
AsyncGenerator support in storage upload


langma:
It would be great if the upload could consume an AsyncGenerator as the `stream` for upload body. It seems that in https://github.com/talkiq/gcloud-aio/blob/master/storage/gcloud/aio/storage/storage.py#L250 there is explicit check that the uploaded body (stream) must be of type `io.IOBase`. 

At least `aiohttp` will accept an AsyncGenerator out of the box, so the change would not be that big, but there might be an issue with `requests` though. Could this be circumvented with something?

My actual use case is that I'm reading a byte stream from a socket and I need to upload that stream as a file to GCS. The current implementation would require either a tempfile, or `BytesIO` in memory. It would be nice if the buffering would be handled by the network stack.

TheKevJames:
We'd definitely be open to accepting a PR implementing this suggestion! I suspect we'd need to ensure that the `requests` version of this library strips out the relevant check -- might require a change to the `py3to2` code generation.

PR #207, opened 2020-09-18
feat(storage): add support for generators for upload body

langma:
Add explicit support for generators, async and sync, to support other than
existing IOBase or file upload sources.

Fixes https://github.com/talkiq/gcloud-aio/issues/206


langma:
I tried it locally and can say that the `async` version is behaving like I thought, allowing me to remove the boilerplate buffering code. However, I haven't tried with the REST version (although requests should also accept a Generator body). Also, I'm using python 3.8 and cannot easily test older versions.

PR #208, opened 2020-09-22
chore(pubsub): bump version

jonathan-johnston:
Major version bump for backwards incompatible change https://github.com/talkiq/gcloud-aio/pull/201.


egalpin (approved):
boom goes the dynamite, as they say.

PR #209, opened 2020-09-22
chore(pubsub): update docs and FlowControl

jonathan-johnston:
I noticed that our docs on the `SubscriberClient` and `FlowControl` were out of date with the version of `google-cloud-pubsub` we are now using. I updated them to reflect the state of internals of `google-cloud-pubsub==1.7.0` (the version we are on) after digging through its source.

In addition, I noticed that the examples use `f = FlowControl(); f.max_messages` and such types of property lookups which our version of `FlowControl` doesn't have so I updated it to make the docs easy to follow.

### DIFF
--- a/pubsub/README.rst
+++ b/pubsub/README.rst
@@ -76,7 +76,7 @@ find in the `official Google documentation`_.
 Since the underlying Google implementation of ``Scheduler`` only allows for the
 concrete ``ThreadScheduler`` which is also the default, we have opted not to
 expose this configuration option. Additionally, we would like to fully
-deprecate said Google implementation in favour of a fully `asyncio`
+deprecate said Google implementation in favour of a fully ``asyncio``
 implementation which uses the event loop as the scheduler.
 
 When subscribing to a subscription you can optionally pass in a ``FlowControl``
--- a/pubsub/README.rst
+++ b/pubsub/README.rst
@@ -38,13 +38,15 @@ Here's the rough usage pattern for subscribing:
 .. code-block:: python
 
     from gcloud.aio.pubsub import SubscriberClient
-    from google.cloud.pubsub_v1.subscriber.message import Message
+    from gcloud.aio.pubsub import SubscriberMessage
 
     client = SubscriberClient()
     # create subscription if it doesn't already exist
-    client.create_subscription('subscription_name', 'topic_name')
+    client.create_subscription(
+        'projects/<project_name>/subscriptions/<subscription_name>',
+        'projects/<project_name>/topics/<topic_name>')
 
-    async def message_callback(message: Message) -> None:
+    async def message_callback(message: SubscriberMessage) -> None:
         try:
             # just an example: process the message however you need to here...
             result = handle(message)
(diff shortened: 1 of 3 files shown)


pubsub/README.rst
```diff
@@ -65,102 +69,96 @@ Here's the rough usage pattern for subscribing:
...
-configuration arguments from the official pubsub client which you can find in
-the `official Google documentation`_.
+Our create_subscription method is a thin wrapper and thus supports almost all
+keyword configuration arguments from the official pubsub client which you can
+find in the `official Google documentation`_.
+
+Since the underlying Google implementation of ``Scheduler`` only allows for the
+concrete ``ThreadScheduler`` which is also the default, we have opted not to
+expose this configuration option. Additionally, we would like to fully
+deprecate said Google implementation in favour of a fully `asyncio`
```
TheKevJames, line 79: two backticks around `asyncio` (one backtick implies link/ref, two backticks is shorthand for `:code:`)

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -43,9 +43,28 @@ def __init__(self, *args: List[Any], **kwargs: Dict[str, Any]) -> None:
             """
             self._flow_control = _FlowControl(*args, **kwargs)
 
+        def __repr__(self) -> str:
+            return self._flow_control.__repr__()
+
         def __getitem__(self, index: int) -> int:
             return self._flow_control[index]
 
+        @property
```
TheKevJames, line 52: Alternative option: dynamically generate these with `__getattr__`. 
```python
>>> class Foo:
...     def __init__(self, x):
...           self._x = x
...     def __getattr__(self, y):
...          return self._x[y]
...
>>> Foo({'a':42}).a
42
```
(or in this case, the body of `__getattr__` would `return getattr(self._flow_control, y)`)

TheKevJames (approved):
LGTM after backtick fix

pubsub/README.rst
jonathan-johnston: Oops, this was just force of habit in Slack! Good catch.

pubsub/gcloud/aio/pubsub/subscriber_client.py
jonathan-johnston: I thought about doing this but (1) RIP type hints and (2) we are pinned to <2.0.0 so our interface shouldn't change in a breaking way. I'm happy doing it like `__getattr__` regardless though. Thoughts?

jonathan-johnston: I made the changes here to use `__getattr__` and added a unit test, but I hadn't realized that with this setup we get `pylint` errors on `FlowControl` in the form of `Instance of 'FlowControl' has no 'max_messages' member (no-member)` for each attribute. But with a toy example which _should_ be identical this error doesn't crop up. Okay this is very weird.

jonathan-johnston: Okay, this is just because `pylint` was only looking at the first definition of `FlowControl` which resides within `if BUILD_GCLOUD_REST:` which of course is unimplemented and has no `__getattr__` method.

jonathan-johnston: LMK if this looks reasonable!

TheKevJames: Yeah, I really like this approach -- delegating absolutely everything to the parent given this is a pass through class makes a lot of sense to me!

PR #210, opened 2020-09-22
chore(pubsub) Add rest variant of pubsub package that uses grpc package under the hood

amartsinovich-dialpad:
Okayyy, so we need `gcloud-rest-pubsub`. If I got it right at the moment `aio` uses `google-cloud-pubsub` under the hood. In that case I don't see why we can't do the same for `rest`! This PR is exactly for that.

I tested this locally and it seems to be working.

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -68,9 +68,7 @@ if BUILD_GCLOUD_REST:
                       flow_control: Union[FlowControl, Tuple[int, ...]] = ()
                       ) -> StreamingPullFuture:
             """
-            Create subscription through pubsub client, hijack the returned
-            "non-concurrent Future" and coerce it into being a "concurrent
-            Future", wrap it into a asyncio Future and return it.
+            Pass call to the google-cloud-pubsub SubscriberClient class
             """
             sub_keepalive: StreamingPullFuture = (
                 self._subscriber.subscribe(
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -68,7 +68,11 @@ if BUILD_GCLOUD_REST:
                       flow_control: Union[FlowControl, Tuple[int, ...]] = ()
                       ) -> StreamingPullFuture:
             """
-            Pass call to the google-cloud-pubsub SubscriberClient class
+            Pass call to the google-cloud-pubsub SubscriberClient class.
+            This method will most likely be deprecated once gcloud-rest-pubsub
+            stop using google-cloud-pubsub under the hood. If this is
+            what you need we strongly recommend using official library.
+
             """
             sub_keepalive: StreamingPullFuture = (
                 self._subscriber.subscribe(
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -1,63 +1,97 @@
+from typing import Any
+from typing import Callable
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Tuple
+from typing import Union
+
 from gcloud.aio.auth import BUILD_GCLOUD_REST  # pylint: disable=no-name-in-module
+from google.cloud.pubsub_v1.types import FlowControl as _FlowControl
+
+class FlowControl:
+    def __init__(self, *args: List[Any], **kwargs: Dict[str, Any]) -> None:
+        """
+        FlowControl transitional wrapper.
+        (FlowControl fields docs)[https://github.com/googleapis/python-pubsub/blob/v1.7.0/google/cloud/pubsub_v1/types.py#L124-L166]  # pylint: disable=line-too-long
+        Google uses a named tuple; here are the fields, defaults:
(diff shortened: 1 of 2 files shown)


pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -1,52 +1,90 @@
...
+        def subscribe(self,
+                      subscription: str,
+                      callback: Callable[[SubscriberMessage], Any],
+                      *,
+                      flow_control: Union[FlowControl, Tuple[int, ...]] = ()
+                      ) -> StreamingPullFuture:  # type: ignore
+            """
+            Create subscription through pubsub client, hijack the returned
+            "non-concurrent Future" and coerce it into being a "concurrent
+            Future", wrap it into a asyncio Future and return it.
```
TheKevJames, line 64: Docstring needs an update -- no asyncio in `gcloud-rest`

TheKevJames (changes requested):
Makes sense to me -- the only reason we hadn't built a `gcloud-rest-pubsub` before was that is was non-sensical (eg. since it's just a wrapper to google-cloud-pubsub and wouldn't add anything special [eg. asyncio compatibility], why bother?). Now that there's a use-case totally makes sense!

You'll also need to update the `pubsub` README with these changes! It calls out in a couple places that it "only builds for aio" and such.

TheKevJames (approved):

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -1,9 +1,39 @@
...
+
+    def __getattr__(self, attr: str) -> Any:
+        return getattr(self._flow_control, attr)
 
 if BUILD_GCLOUD_REST:
-    class FlowControl:
-        def __init__(self, **kwargs) -> None:
-            raise NotImplementedError('this class is only implemented in aio')
+    from google.cloud import pubsub_v1 as pubsub
+    from google.cloud.pubsub_v1.subscriber.futures import StreamingPullFuture
```
jonathan-johnston, line 36: This is the only thing that I'm worried about since it re-introduces `google-cloud-pubsub` as an explicit dependency in this client. In order to deprecate the gRPC-based `google-cloud-pubsub` we will have to encapsulate this in `gcloud-rest-pubsub` anyway (in the same way as we did for `Message` and `FlowControl`), so can we do it earlier instead?

jonathan-johnston (changes requested):
Overall LGTM, but one blocker!

pubsub/gcloud/aio/pubsub/subscriber_client.py
amartsinovich-dialpad: What is explicit dependency in this context? I thought `FlowControl` and `Message` were ones, because _clients_ had to import those from `google-cloud-pubsub`, create instances of them and then pass to `gcloud-*-pubsub`. Here, however, we don't expect clients to work with `StreamingPullFuture` directly? If we change it to another future with the same interface it shouldn't break anything?

Looping back to our discussion about `subscribe` method. I think deprecating `google-cloud-pubsub` would mean `gcloud-rest-pubsub` gonna lose `subscribe` method at the first place or at least have some breaking changes. WDYT?

TheKevJames: Maybe I'm misunderstanding, but I did not think we were working to remove grpc from pubsub _yet_. ie. Goal 1 is to have all our apps go through `gcloud-*-pubsub` instead of hit `google-cloud-pubsub` directly for everything without actually changing anything about behaviour. Goal 2 will be to modify `gcloud-*-pubsub` in place to do it all via REST endpoints instead of using the official lib.

jonathan-johnston: We aren't trying to remove grpc from pubsub right now. The interface shouldn't include references to `google-cloud-pubsub` objects so that we can modify the underlying implementation to be non-grpc. If we modified the underlying interface _now_ then we would still have to remove references to `google-cloud-pubsub`. Instead we are laying the groundwork to make that implementation swap easy, and this is what I mean by removing any "explicit" dependencies (i.e. references) to `google-cloud-pubsub` objects.

@amartsinovich-dialpad I do agree that this is a different case because users of this lib don't have to create this `StreamingPullFuture` but instead just receive it and use it. So in that sense it's not the worst thing in the world to leave as-is. That being said, it would be nicer to return another future with the same interface or perhaps a simplified interface. Like you said re: `subscribe` deprecation, you're convincing me that this should actually be fine because we're going to nuke it anyway.

Okay, this is fine! Thanks for the discussion y'all. Can you include a comment in the `subscribe` code that includes the gist of our conversation about deprecation & using a synchronous get-based approach to replace `subscribe`?

jonathan-johnston (approved):

pubsub/gcloud/aio/pubsub/subscriber_client.py
amartsinovich-dialpad: Done!

amartsinovich-dialpad:
@jonathan-johnston @TheKevJames thank you folks for the review. What happens now? 🙄

TheKevJames:
@amartsinovich-dialpad now you can (fix merge conflicts and) merge -- @jonathan-johnston should be able to help out in getting release notes etc set up so we can do a new version release.

Issue #212, opened 2020-09-25
Typing


cblegare:
It would be nice to have type hints or stubs for this library!

Mypy error: 

```
error: Skipping analyzing 'gcloud.aio.pubsub': found module but no type hints or library stubs
```

TheKevJames:
Huh, that's odd -- our pubsub library (and the others) is [pretty much fully type hinted](https://github.com/talkiq/gcloud-aio/blob/master/pubsub/gcloud/aio/pubsub/subscriber_client.py) (though we are not currently running `mypy` as a linter due to reasons which should have recently become unblocked, so I'd be 100% in favour of starting to do that and fixing any issues which might pop up!!). I wonder if this has to do with how we've set up the shared namespaces across the multiple packages -- it could be the case we got something wrong there, since I've never really tried to do this before!

That said, we make use of this package internally and haven't run into this issue -- mypy seems to be able to properly trace type hints through the library in our case. Can you share anything about your configuration which might help us debug what's going wrong here?

cblegare:
Maybe simply including an empty `py.typed` file would to the trick! 

see https://www.python.org/dev/peps/pep-0561/#packaging-type-information

TheKevJames:
This seems to be working now with the above `py.typed` changes :)

PR #213, opened 2020-09-29
style(lint): run mypy, fix typing issues

TheKevJames:
Adds mypy linting and fixes hints across all packages. See #212 for a bit more info. Looks like we might also need to play around with the `py.typed` file in order to ensure clients can make use of these hints.

### DIFF
--- a/auth/gcloud/aio/auth/iam.py
+++ b/auth/gcloud/aio/auth/iam.py
@@ -1,4 +1,3 @@
-import io
 import json
 from typing import Any
 from typing import AnyStr
--- a/auth/gcloud/aio/auth/iam.py
+++ b/auth/gcloud/aio/auth/iam.py
@@ -1,6 +1,9 @@
 import io
 import json
+from typing import Any
+from typing import AnyStr
 from typing import Dict
+from typing import IO
 from typing import List
 from typing import Optional
 from typing import Union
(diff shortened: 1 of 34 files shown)


jonathan-johnston (approved):

egalpin (approved):

auth/gcloud/aio/auth/iam.py
```diff
@@ -70,7 +72,8 @@ async def get_public_key(self, key_id: Optional[str] = None,
 
         resp = await s.get(url=url, headers=headers, timeout=timeout)
 
-        return await resp.json()
+        data: Dict[str, str] = await resp.json()
```
amartsinovich-dialpad, line 75: That's not the first time I see parsed json as `Dict[str, str]`. Why isn't it `Dict[str, Any]`?

amartsinovich-dialpad (approved):

auth/gcloud/aio/auth/iam.py
TheKevJames: According to the [API docs](https://cloud.google.com/iam/docs/reference/rest/v1/projects.serviceAccounts.keys/get), the response body should look like [this](https://cloud.google.com/iam/docs/reference/rest/v1/projects.serviceAccounts.keys#ServiceAccountKey), which just so happens to be `Dict[str, str]`. I figured more specific is better where possible!

PR #217, opened 2020-10-02
fix(pubsub) Cast Message to SubscriberMessage in a callback passed to subscribe()

amartsinovich-dialpad:
so @jonathan-johnston draw my attention to the fact that things like that:

```
if BUILD_GCLOUD_REST:
    class SubscriberMessage:
        def __init__(self, **kwargs: Any) -> None:
            raise NotImplementedError('this class is only implemented in aio')

        @staticmethod
        def from_google_cloud(message: Any) -> 'SubscriberMessage':
            raise NotImplementedError('this class is only implemented in aio')
```

Should probably not exist in a working `gcloud-rest-*` library.

In this PR:
1. Remove `if BUILD_GCLOUD_REST` and define `SubscriberMessage` for both `-aio-` and `-rest-` variants
2. Cast `Message` to `SubscriberMessage` in  `callback` provided to `SubscriberClient.subscribe` method

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -7,7 +7,10 @@ from typing import Tuple
 from typing import Union
 
 from gcloud.aio.auth import BUILD_GCLOUD_REST  # pylint: disable=no-name-in-module
+from google.cloud.pubsub_v1.subscriber.message import Message
 from google.cloud.pubsub_v1.types import FlowControl as _FlowControl
+
+from .subscriber_message import SubscriberMessage
 
 class FlowControl:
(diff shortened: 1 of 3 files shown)


jonathan-johnston (approved):
LGTM! Thanks for doing this Alex!

jonathan-johnston:
@amartsinovich-dialpad Did you run any tests on this outside of the unit tests? (ex. as a package in one of our services, even if just running `mypy` on our service code with this package installed to catch type errors)

amartsinovich-dialpad:
@jonathan-johnston I ran realtime-runner e2e test using this package. FYI, I secretly put dev versions [here](https://manage.fury.io/dashboard/talkiq/package/pkg_i5UNS/versions)

PR #219, opened 2020-10-06
hotfix(pubsub) Fix imports in subscriber_client.py

amartsinovich-dialpad:
If there are ways to test code under `if REST:` condition please let me know >.<


jonathan-johnston (approved):

jonathan-johnston:
@amartsinovich-dialpad AFAIK you can only test by creating a dev package -- the `aio -> rest` munging happens in the CI steps and involves a ton of string manipulation to make two separate code bases.

PR #223, opened 2020-10-09
feat(pubsub): proper subscriber cleanup

shaundialpad:
Currently calling [sub_keepalive.cancel()](https://github.com/talkiq/gcloud-aio/blob/pubsub-2.1.2/pubsub/gcloud/aio/pubsub/subscriber_client.py#L181) on the `asyncio.Future` returned by the subscriber does not properly invoke the original `cancel` method of Google's `StreamingPullFuture`. This fact was due to modifying and wrapping the future with the intention of getting Pub/Sub to correctly run callbacks on the asyncio event loop.

After some investigation, it was discovered that wrapping the future was not needed, and in fact only running the callback via `asyncio.run_coroutine_threadsafe`, which was added _after_ these changes, is needed to have Pub/Sub and asyncio play nicely together.

Now, `cancel`ling the returned future invokes the behaviour [here](https://github.com/googleapis/python-pubsub/blob/v1.7.0/google/cloud/pubsub_v1/subscriber/futures.py#L46) and so we cleanup properly as desired (i.e. stop pulling messages).

More details: [VAE-2169](https://switchcomm.atlassian.net/browse/VAE-2169)

NOTE: This change _probably_ requires a major version bump, since `cancel` now has new behavior, and typing has changed.

- [x] This change has been tested in realtime-egress and offline runner. (apps run properly, termination invokes `cancel` which correctly stops pulling messages)

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -8,6 +8,7 @@ from typing import Union
 
 from gcloud.aio.auth import BUILD_GCLOUD_REST  # pylint: disable=no-name-in-module
 from google.api_core import exceptions
+from google.cloud.pubsub_v1.subscriber.futures import StreamingPullFuture
 from google.cloud.pubsub_v1.subscriber.message import Message
 from google.cloud.pubsub_v1.types import FlowControl as _FlowControl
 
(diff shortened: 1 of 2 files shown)


eddiedialpad:
Hai Shaun! How were you able to test this?

TheKevJames:
@shaundialpad coould we get a docs update as well, which explains how this method will need to be used (eg. with `asyncio.run_coroutine_threadsafe`)?

shaundialpad:
> @shaundialpad coould we get a docs update as well, which explains how this method will need to be used (eg. with `asyncio.run_coroutine_threadsafe`)?

Sorry which documentation are you referring to? Tasks are automatically scheduled this way in our `SubscriberClient` so there is no extra work needed from the user.

TheKevJames (approved):
@shaundialpad 🤦 I definitely knew that and totally wasn't thinking that code lived client-side.

PR #224, opened 2020-10-10
feat(bigquery) Add some of the missing params to insert_via_load

amartsinovich-dialpad:
bq_publisher uses some parameters that are not supported in bigquery atm.

### DIFF
--- a/bigquery/gcloud/aio/bigquery/bigquery.py
+++ b/bigquery/gcloud/aio/bigquery/bigquery.py
@@ -301,15 +301,12 @@ class Table:
         if not source_uris:
             return {}
 
-        if not schema_update_options:
-            schema_update_options = []
-
         project = await self.project()
         url = f'{API_ROOT}/projects/{project}/jobs'
 
         body = self._make_load_body(
             source_uris, project, autodetect, source_format, write_disposition,
-            ignore_unknown_values, schema_update_options
+            ignore_unknown_values, schema_update_options or []
         )
         return await self._post_json(url, body, session, timeout)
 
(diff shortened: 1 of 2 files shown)


bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -282,16 +293,23 @@ async def insert_via_load(
...
-            timeout: int = 60) -> Dict[str, Any]:
+            timeout: int = 60,
+            ignore_unknown_values: bool = False,
+            schema_update_options: Optional[List[SchemaUpdateOption]] = None
+        ) -> Dict[str, Any]:
         """Loads entities from storage to BigQuery."""
         if not source_uris:
             return {}
 
+        if not schema_update_options:
```
TheKevJames, line 304: pro-tip: you can also do default values like this with `or`, eg:
```python
body = self._make_load_body(
    source_uris, project, autodetect, source_format, write_disposition,
    ignore_unknown_values, schema_update_options or [],
)
```

TheKevJames (approved):

bigquery/gcloud/aio/bigquery/bigquery.py
amartsinovich-dialpad: Yeah, it's okay here, but I think it's generally not the best pattern as it won't allow user to pass a falsey value.

amartsinovich-dialpad: ... I could say if I actually wrote `if schema_update_options is None` 🤦

jonathan-johnston (approved):
LGTM!

TheKevJames (approved):

jonathan-johnston:
@amartsinovich-dialpad LMK when you're all set 🙂

PR #225, opened 2020-10-10
feat(storage): add support for generators for upload body

### LINKED ISSUE #206: AsyncGenerator support in storage upload

langma:
Add explicit support for generators, async and sync, to support other than
existing IOBase or file upload sources.

Fixes #206

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -4,9 +4,11 @@ import json
 import logging
 import mimetypes
 import os
+import sys
 from typing import Any
 from typing import AnyStr
 from typing import Dict
+from typing import Generator
 from typing import IO
 from typing import Optional
 from typing import Tuple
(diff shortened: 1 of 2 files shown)


langma:
Closing as seemingly irrelevant.

hasier:
@TheKevJames sorry for the direct ping. The work @langma has done here seems spot on to fix #206 and it would greatly help to avoid boilerplate code to buffer generators (and the additional step within `aiohttp` to [run the sync `read`s in the eventloop executor](https://github.com/aio-libs/aiohttp/blob/62173bea1e4e3538d46e6e9b94d727d870d86879/aiohttp/payload.py#L298-L306)). `requests` seems to also [accept generators out of the box](https://requests.readthedocs.io/en/latest/user/advanced/#chunk-encoded-requests). Would you be open to taking a look into this and potentially reopening?

PR #227, opened 2020-10-13
feat(bq) Add method for v2/jobs/get

amartsinovich-dialpad:
I don't like this PR. There is no reason `get_job` should be a method of `Table` class, yet there is no other place for it. Unless we want to redesign `gcloud-aio-bigquery` :(

Let me know what you think

### DIFF
--- a/bigquery/gcloud/aio/bigquery/bigquery.py
+++ b/bigquery/gcloud/aio/bigquery/bigquery.py
@@ -51,12 +51,13 @@ class SchemaUpdateOption(Enum):
     ALLOW_FIELD_RELAXATION = 'ALLOW_FIELD_RELAXATION'
 
 
-class BQResource:
-
-    def __init__(self, project: Optional[str] = None,
+class Job:
+    def __init__(self, job_id: str,
+                 project: Optional[str] = None,
                  service_file: Optional[Union[str, io.IOBase]] = None,
                  session: Optional[Session] = None,
                  token: Optional[Token] = None) -> None:
+        self.job_id = job_id
         self._project = project
         self.session = AioSession(session)
         self.token = token or Token(service_file=service_file, scopes=SCOPES,
@@ -84,17 +85,6 @@ class BQResource:
         return {
             'Authorization': f'Bearer {token}',
         }
-
-
-class Job(BQResource):
-    def __init__(self, job_id: str,
-                 project: Optional[str] = None,
-                 service_file: Optional[Union[str, io.IOBase]] = None,
-                 session: Optional[Session] = None,
-                 token: Optional[Token] = None) -> None:
-        self.job_id = job_id
-        super().__init__(project=project, service_file=service_file,
-                         session=session, token=token)
 
     # https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/get
     async def get_job(self, session: Optional[Session] = None,
@@ -142,7 +132,7 @@ class Job(BQResource):
         return data
 
 
-class Table(BQResource):
+class Table:
     def __init__(self, dataset_name: str, table_name: str,
                  project: Optional[str] = None,
                  service_file: Optional[Union[str, io.IOBase]] = None,
@@ -150,9 +140,33 @@ class Table(BQResource):
                  token: Optional[Token] = None) -> None:
         self.dataset_name = dataset_name
         self.table_name = table_name
-        super().__init__(project=project,
-                         service_file=service_file,
-                         session=session, token=token)
+        self._project = project
+        self.session = AioSession(session)
+        self.token = token or Token(service_file=service_file, scopes=SCOPES,
+                                    session=self.session.session)
+
+    async def project(self) -> str:
+        if self._project:
+            return self._project
+
+        if BIGQUERY_EMULATOR_HOST:
+            self._project = str(os.environ.get('BIGQUERY_PROJECT_ID', 'dev'))
+            return self._project
+
+        self._project = await self.token.get_project()
+        if self._project:
+            return self._project
+
+        raise Exception('could not determine project, please set it manually')
+
+    async def headers(self) -> Dict[str, str]:
+        if BIGQUERY_EMULATOR_HOST:
+            return {}
+
+        token = await self.token.get()
+        return {
+            'Authorization': f'Bearer {token}',
+        }
 
     @staticmethod
     def _mk_unique_insert_id(row: Dict[str, Any]) -> str:
--- a/bigquery/gcloud/aio/bigquery/bigquery.py
+++ b/bigquery/gcloud/aio/bigquery/bigquery.py
@@ -150,7 +150,8 @@ class Table(BQResource):
                  token: Optional[Token] = None) -> None:
         self.dataset_name = dataset_name
         self.table_name = table_name
-        super().__init__(project=project, service_file=service_file,
+        super().__init__(project=project,
+                         service_file=service_file,
                          session=session, token=token)
 
     @staticmethod
@@ -341,7 +342,7 @@ class Table(BQResource):
     async def insert_via_copy(
             self, destination_project: str, destination_dataset: str,
             destination_table: str, session: Optional[Session] = None,
-            timeout: int = 60) -> Dict[str, Any]:
+            timeout: int = 60) -> Job:
         """Copy BQ table to another table in BQ"""
         project = await self.project()
         url = f'{API_ROOT}/projects/{project}/jobs'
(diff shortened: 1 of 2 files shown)


amartsinovich-dialpad:
@jonathan-johnston 
> Are there other operations on jobs that would fit in a Job class?
For a thin client library it seems excessive to break down REST resource into several other entities, so we'll most likely want to put them all into hypothetical `Job` class without asking this question. I see several paths here:
1. Go for `Job` class, rewire code shared between `Table` and `Job`
2. Go for a function, inevitably rewire shared code
3. Leave it as is and live with it
4. Step back, ditch `Table` and `Job` classes and create `Client` instead. Also break backward compatibility 👌 

# 1 seems like the most reasonable to me except for it normalizes the approach of having class-per-resource which I don't really understand in terms of what value it brings to the table

TheKevJames:
I also like the idea of #1 -- that feels like an incremental step towards doing this a bit cleaner without needing to break too much client code. Looks like `-bigquery` never got the same treatment we've applied to other packages; eg. that classes and methods should match the REST resources described in the Google docs, more or less. For maximal support, we'd want a class for each of these things (both types and groups) like we do in eg. `gcloud-aio-datastore` ([source](https://cloud.google.com/bigquery/docs/reference/rest/v2)):
image

Obviously, that's *way* out of scope. But if there's a decent middle ground which heads us in that direction without being a major breaking change? That sounds pretty good to me.

amartsinovich-dialpad:
@TheKevJames 
how about this naive thingy?
```
class BQResource:

    def __init__(self, project: Optional[str] = None,
                 service_file: Optional[Union[str, io.IOBase]] = None,
                 token: Optional[Token] = None) -> None:
        self._project = project
        self.session = AioSession(session)
        self.token = token or Token(service_file=service_file, scopes=SCOPES,
                                    session=self.session.session)

    async def project(self) -> str:
        if self._project:
            return self._project

        if BIGQUERY_EMULATOR_HOST:
            self._project = str(os.environ.get('BIGQUERY_PROJECT_ID', 'dev'))
            return self._project

        self._project = await self.token.get_project()
        if self._project:
            return self._project

        raise Exception('could not determine project, please set it manually')


class Table(BQResource):
    def __init__(self, dataset_name: str, table_name: str,
                 **kwargs) -> None:
        self.dataset_name = dataset_name
        self.table_name = table_name
        super().__init__(**kwargs)
```

TheKevJames:
@amartsinovich-dialpad oh, for sharing code between a `Table` and `Job` class, you mean? Hmm, depends what it means for client usage -- ideally, we'd avoid forcing users to handle too many moving pieces themselves. Does it make sense to construct those classes from each other like we do in [gcloud-aio-storage](https://github.com/talkiq/gcloud-aio/blob/00f4f4dde8daf1284b02317e7052108a3920dcf2/storage/gcloud/aio/storage/storage.py#L76) or something like that?

Just one nit on your psuedo-code: prefer avoiding `kwargs` in favor of duplicating the arguments; better to be explicit than to mask any argument mismatches between classes and lose the type hints.

amartsinovich-dialpad:
@TheKevJames 
>Does it make sense to construct those classes from each other like we do in `gcloud-aio-storage` or something like that?

It's possible, I guess, but, like I mentioned, I don't see what value this additional layer of python objects brings us. Some methods will return raw http response, others will return objects-resources. Do we have a vision for the package that we can derive those design decisions from?

amartsinovich-dialpad:
> Just one nit on your psuedo-code: prefer avoiding kwargs in favor of duplicating the arguments; better to be explicit than to mask any argument mismatches between classes and lose the type hints.

image

TheKevJames:
But Python *is* psuedo-code! :P 

In `gcloud-aio-datastore`, we serialize/deserialize all responses with custom objects, which seems to be pretty ergonomic. So you can call a method which returns a `Query` object... and then immediately run that query directly rather than needing to juggle the response into some other method.

The datastore style has been pretty nice to work with, and I like the idea of having all the `gcloud-aio-*` packages be reasonably similar in usage, so that's my first thought for a good place to head towards. That said, maybe there's a better way to do things? Following the `google-cloud-bigquery` structure might be reasonable, or maybe there's some other design which would be best?

amartsinovich-dialpad:
@TheKevJames @jonathan-johnston 
I don't know what would be the best way to put it, but here's a version with `Job` class maybe it's better

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -296,10 +361,8 @@ async def insert_via_load(
             timeout: int = 60,
             ignore_unknown_values: bool = False,
             schema_update_options: Optional[List[SchemaUpdateOption]] = None
-        ) -> Dict[str, Any]:
+        ) -> Job:
         """Loads entities from storage to BigQuery."""
-        if not source_uris:
```
TheKevJames, line 301: Hmm, I'm not sure we can remove this -- BQ errors when an empty list is passed in here. Maybe we have to early-return a faked job or have this method return `Optional[Job]`?

TheKevJames (changes requested):
Do the other `Table.insert_*` methods return Jobs as well? Probably best to do 'em all in one go.

Overall, I like the `BQResource` design!

bigquery/gcloud/aio/bigquery/bigquery.py
amartsinovich-dialpad: Why isn't it ok to raise on empty list? That seems to be an API requirement, I don't see why the client should pretend the call was successful. If we return fake Job the client will probably fail on fetching the job's status, if we return `None` it's gonna trigger `KeyError` downstream. Both cases will just make debugging process longer.

TheKevJames: I guess it's mostly to "fix" an oddity of their API -- if you write a function which requests we "load zero rows of data", and we do nothing, should we succeed or fail? I'd argue that's a success; the API, though, throws an error since that case does not have validation on the Google side of things.

Basically, if we remove this than users need to do the additional work of capturing errors and introspecting into them to make sure the error wasn't actually a success.

TheKevJames: Agreed 100% on the fake job, though. I'm leaning towards the `Optional[Job]` case -- that shouldn't cause KeyErrors since that's what the type hints are for in the first place! Linters should prevent folks from attempting to use an `Optional[Job]` as if it were a `Job`.

amartsinovich-dialpad: > I guess it's mostly to "fix" an oddity of their API -- if you write a function which requests we "load zero rows of data", and we do nothing, should we succeed or fail? I'd argue that's a success; the API, though, throws an error since that case does not have validation on the Google side of things.

But it does have validation:
```
{
  "error": {
    "code": 400,
    "message": "Load configuration must specify at least one source URI",
    "errors": [
      {
        "message": "Load configuration must specify at least one source URI",
        "domain": "global",
        "reason": "invalid"
      }
    ],
    "status": "INVALID_ARGUMENT"
  }
}
```
That's just how google defined their API. 

> Basically, if we remove this than users need to do the additional work of capturing errors and introspecting into them to make sure the error wasn't actually a success.

Well that sounds bad, but wouldn't it be a single `insert_via_load(source_uris) if source_uris` check?
I see a problem though: a user is required to have enough awareness to know that the list cannot be empty and we have no way to reflect this requirement in type hints. At the same time, `Optional[Job]` is not the best in explaining the situation either.

I dunno. Let me know if that changed you mind, if not I won't fight over google's schema.

TheKevJames: Hmm, yeah, I guess it's on the client either way; they'll either need to write:
```python
if source_uris:
    job = insert_via_load(source_uris)
    # 
```
or 
```python
job = insert_via_load(source_uris)
if job:
    # 
```

I guess it doesn't really matter as much as I thought haha. I think I'm still slightly more inclined towards `Optional[Job]` (just personally find it a bit clearer / more understandable here /shrug) but I don't have any strong opinions either way. If you do keep this check out, please make sure that this change is both highlighted in release notes and we make sure none of our own apps will run into issues here!

amartsinovich-dialpad: Ok, I'll stick to schema then. Well that's gonna be fun to verify all our projects!

TheKevJames (approved):

jonathan-johnston (approved):

amartsinovich-dialpad:
Ok, inconcistency with other `gcloud-*` libraries is bad, so here's good old copypasta solution

jonathan-johnston (approved):
Eggsellent 🍳

PR #232, opened 2020-10-20
feat(bigquery) Move insert_via_query to Job class

amartsinovich-dialpad:
Here we go again. So turned out `insert_via_query` can be totally executed against `destinationTable: null`. That kind of defeats the purpose of having this method in the `Table` class. So in this PR we move the method to `Job` (making `job_id` optional along the way) and mark previous one as deprecated.

### DIFF
--- a/bigquery/gcloud/aio/bigquery/bigquery.py
+++ b/bigquery/gcloud/aio/bigquery/bigquery.py
@@ -2,7 +2,6 @@ import io
 import json
 import os
 import uuid
-import warnings
 from enum import Enum
 from typing import Any
 from typing import Callable
@@ -468,6 +467,7 @@ class Table:
             timeout: int = 60, use_query_cache: bool = True,
             dry_run: bool = False, use_legacy_sql: bool = True) -> Job:
         """Create table as a result of the query"""
+        import warnings
         warnings.warn('using Table#insert_via_query is deprecated.'
                       'use Job#insert_via_query instead', DeprecationWarning)
         project = await self.project()
--- a/bigquery/gcloud/aio/bigquery/bigquery.py
+++ b/bigquery/gcloud/aio/bigquery/bigquery.py
@@ -1,4 +1,3 @@
-# pylint disable=reimported
 import io
 import json
 import os
--- a/bigquery/gcloud/aio/bigquery/bigquery.py
+++ b/bigquery/gcloud/aio/bigquery/bigquery.py
@@ -2,6 +2,7 @@ import io
 import json
 import os
 import uuid
+import warnings
 from enum import Enum
 from typing import Any
 from typing import Callable
@@ -52,7 +53,7 @@ class SchemaUpdateOption(Enum):
 
 
 class Job:
-    def __init__(self, job_id: str,
+    def __init__(self, job_id: Optional[str] = None,
                  project: Optional[str] = None,
                  service_file: Optional[Union[str, io.IOBase]] = None,
                  session: Optional[Session] = None,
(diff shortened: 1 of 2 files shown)


bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -86,6 +87,43 @@ async def headers(self) -> Dict[str, str]:
             'Authorization': f'Bearer {token}',
         }
 
+    @staticmethod
+    def _make_query_body(
+            query: str,
+            write_disposition: Disposition,
+            use_query_cache: bool,
+            dry_run: bool, use_legacy_sql: bool,
+            destination_table: Optional[Dict[str, object]]) -> Dict[str, Any]:
```
TheKevJames, line 96: Might make more sense to have this method accept an `Optional[Table]` directly, since this method seems to be responsible for doing all the jsonification -- as you currently have it, you need the `insert_via_query` method to have some understanding of the payload format, which is non-ideal since that spreads this info across multiple places.

TheKevJames (commented):
(had to interrupt my code review, will take a full look soon)

bigquery/gcloud/aio/bigquery/bigquery.py
amartsinovich-dialpad: yes, overthought myself. fixed now

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -401,15 +467,18 @@ async def insert_via_query(
             self, query: str, session: Optional[Session] = None,
             write_disposition: Disposition = Disposition.WRITE_EMPTY,
             timeout: int = 60, use_query_cache: bool = True,
-            dry_run: bool = False) -> Job:
+            dry_run: bool = False, use_legacy_sql: bool = True) -> Job:
```
TheKevJames, line 470: Given we're deprecating this method, let's avoid adding the new `use_legacy_sql` functionality -- rather stop maintaining it entirely than have it be partially maintained, since that might be a bit confusing to users.

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -52,7 +53,7 @@ class SchemaUpdateOption(Enum):
 
 
 class Job:
-    def __init__(self, job_id: str,
+    def __init__(self, job_id: Optional[str] = None,
```
TheKevJames, line 56: non-actionable comment: something about making this `job_id` optional feels off to me. Like, `Job.insert_via_query` totally makes sense to me, since that's how Google has it defined, but the fact that we might have `Job()` instances which aren't actually pointing to real jobs... I dunno, I wonder if it'd be hard to reason about?

Might point to aiming towards future refactoring efforts -- eg. it feels to me like perhaps this is a instance method vs class method thing? ie. doing `Job.insert_via_query(...)` makes sense, rather than having to first initialize the `Job` class. That messes with the `self.project()` calls, so it definitely doesn't fit the current system, but I wonder how best to make it work. Maybe the official Google libs have a good answer here?

TheKevJames, line 56: Actionable comment: given `job_id` is now `Optional`, we need to reflect that elsewhere. eg. calling `get_job()` on a `Job` with a null ID is gonna do weird string interpolation things and try accessing the job with ID `"None"`. We might need to add some guards to the various methods which check for `if self.job_id is None: raise Exception('cannot get job with no specified ID')` or something? Thoughts?

amartsinovich-dialpad: Well, this is in fact a sneak peak onto the abyss of complexity that google library is: https://github.com/googleapis/python-bigquery/blob/master/google/cloud/bigquery/client.py#L2778-L2809

In order to run query `google-cloud-bigquery` _generates_ job_id locally, creates a `JobReference` object that references nothing real at this point, then `QueryJob` object, links them together and only after this a call to the server is made.

amartsinovich-dialpad: I think I have an old tester's trauma that make me dislike restrictive clients. Oftentimes I _wanted_ to send `resource/None` to a server but a normal production client won't let me. 

Yeah we can raise on that, although that would require a custom exception I believe. Server error is not that bad too `ClientResponseError: 404, message='Not Found: {\n  "error": {\n    "code": 404,\n    "message": "Not found: Job project:None",\n    "errors": [\n      {\n        "message": "Not found: Job project:None",\n        "domain": "global",\n        "reason": "notFound"\n      }\n    ],\n    "status": "NOT_FOUND"\n  }\n}\n', url=URL('https://www.googleapis.com/bigquery/v2/projects/project/jobs/None')` so I'm good either way.

TheKevJames: Yeah, that's a reasonable enough error message. Alrighty, the whole null-job-id thing looks good to me then!

TheKevJames (approved):

jonathan-johnston (approved):

PR #233, opened 2020-10-21
chore(bq): bump bigquery from 5.0.0 -> 5.1.0

jonathan-johnston:
Releases new feature https://github.com/talkiq/gcloud-aio/pull/231

### DIFF
--- a/bigquery/setup.py
+++ b/bigquery/setup.py
@@ -13,7 +13,7 @@ with open(os.path.join(PACKAGE_ROOT, 'requirements.txt')) as f:
 
 setuptools.setup(
     name='gcloud-aio-bigquery',
-    version='5.0.0',
+    version='5.1.0',
     description='Python Client for Google Cloud BigQuery',
     long_description=README,
     namespace_packages=[


amartsinovich-dialpad (approved):
Could be bundled with https://github.com/talkiq/gcloud-aio/pull/232 But won't hurt releasing it earlier 👍

TheKevJames (approved):

jonathan-johnston:
@amartsinovich-dialpad I can hold off releasing this until #232 is in and can rebase off `master`. But isn't #232 an interface-breaking change?

amartsinovich-dialpad:
# 232:
1. `Job(job_id: str)` -> `Job(job_id: Optional[str]`
2. NEW `Job#insert_via_query`
3. DEPRECATED (but still working!) `Table#insert_via_query`
4. Added `use_legacy_sql: bool = True` to `Table#insert_via_query` as optional last argument

I have to confess I don't know the formal definition of breaking changes, but all those seem backwards compatible?

TheKevJames:
Nah, I don't think any of those are interface-breaking. (1) is strictly additive (ie. adding the `None` case), so all should be well there.

Mind, that doesn't mean we need to hold off here, necessarily. #232 can always just become `v5.2.0` instead /shrug

jonathan-johnston:
My bad! I didn't see that the function was deprecated, I thought it was removed entirely. Definitely not breaking changes then!

jonathan-johnston:
Rebased on `master` to pick up #232

PR #234, opened 2020-10-27
Add helper function to wait for job result

DaniilAnichin:
As for now, using any of insert_* jobs will result in implementing such waiting on caller side, otherwise one cannot guarantee loading success.


DaniilAnichin:
Lint error says `TimeoutError` is not defined; However, it's built-in since version 3.3, and gcloud-aio-bigquery states that minimum required python is 3.6; Tests also fail on python 2.7, which is seemingly not supported (though mentioned in readme & has set up tests). What should be the solution? Tune sources to be 2.7 compatible, or turn over 2.7 support in tests too?

TheKevJames:
@DaniilAnichin this repository builds multiple sets of packages: gcloud-aio and gcloud-rest. Though the former is indeed only meant to support 3.6+, the latter is fully compatible with Python 2.7.

You'll need to make some of your code conditional to account for this, eg. see [this example in gcloud-aio-storage](https://github.com/talkiq/gcloud-aio/blob/4215aa755db431f0b5b779723053acb0240b9cbf/storage/gcloud/aio/storage/storage.py#L21-L29) where we selectively load `asyncio.sleep` or `time.sleep` based on which package we're building.

As for the TimeoutError... looks like it was broken out from OSError in 3.3, so making use of OSError directly in the case of Py2.7 might be best!

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -131,6 +133,21 @@ async def cancel(self, session: Optional[Session] = None,
...
+        for i in range(backoff_count):
+            job_data = await self.get_job(session)
+            status = job_data['status']
+            if status['state'] == 'DONE':
+                if 'errorResult' in status:
+                    raise Exception(
+                        'Job finished with errors: ', status['errors'],
+                    )
+                return job_data
+            await sleep(i * 0.5)
```
TheKevJames, line 147: I'm not sure we want to be setting the exact backoff parameters on behalf of our users -- everyone has different use cases and we as a library should not be enforcing our assumptions on applications.

We probably want to expose the consistently useful bits of this (eg. "check if a job is done") but allow users to do their own backoff handling however they'd like -- eg. at TalkIQ we tend to make use of the [backoff](https://pypi.org/project/backoff/) library with very specific configurations for various use-cases.

What do you think about doing something similar here? If you remove the backoff handler, this method can then be configured by clients however they'd like, eg. something like:
```python
# gcloud-aio code:
    async def result(self, session: Optional[Session] = None) -> Dict[str, Any]:
        job_data = await self.get_job(session)
        status = job_data['status']
        if status['state'] == 'DONE':
            if 'errorResult' in status:
                raise Exception('Job finished with errors: ', status['errors'])
            return job_data
        raise OSError('Job results are still pending')

# client code:
@backoff.on_exception(backoff.expo, OSError, max_tries=8)  # or whatever the client wants here!
async def get_job_result():
    return await client.result()
```

TheKevJames (changes requested):

bigquery/gcloud/aio/bigquery/bigquery.py
DaniilAnichin: Yeah, looks nice; Will update soon

TheKevJames (approved):
Thanks! Looks great.

PR #239, opened 2020-11-04
feat(auth): add json argument to session.post and session.put

kylepad:
**NEW** add `json` argument to `Session.post` and `Session.put`
This feature is already supported by the the `aiohttp.ClientSession` and `request.Session` but as we do not allow `**kwargs` passthrough it makes sense to allow json as a specific argument for `POST` and `PUT` methods.

### DIFF
--- a/auth/gcloud/aio/auth/session.py
+++ b/auth/gcloud/aio/auth/session.py
@@ -37,7 +37,7 @@ class BaseSession:
 
     @abstractmethod
     async def post(self, url: str, headers: Dict[str, str],
-                   data: Optional[str], timeout: int,
+                   data: Optional[str], json: Optional[Any], timeout: int,
                    params: Optional[Dict[str, str]]) -> Response:
         pass
 
@@ -48,7 +48,7 @@ class BaseSession:
 
     @abstractmethod
     async def put(self, url: str, headers: Dict[str, str], data: IO[Any],
-                  timeout: int) -> Response:
+                  json: Optional[Any], timeout: int) -> Response:
         pass
 
     @abstractmethod
@@ -108,11 +108,13 @@ if not BUILD_GCLOUD_REST:
             return self._session
 
         async def post(self, url: str, headers: Dict[str, str],
-                       data: Optional[str] = None, timeout: int = 10,
+                       data: Optional[str] = None, json: Optional[Any] = None,
+                       timeout: int = 10,
                        params: Optional[Dict[str, str]] = None
                        ) -> aiohttp.ClientResponse:
-            resp = await self.session.post(url, data=data, headers=headers,
-                                           timeout=timeout, params=params)
+            resp = await self.session.post(url, data=data, json=json,
+                                           headers=headers, timeout=timeout,
+                                           params=params)
             await _raise_for_status(resp)
             return resp
 
@@ -126,9 +128,10 @@ if not BUILD_GCLOUD_REST:
             return resp
 
         async def put(self, url: str, headers: Dict[str, str], data: IO[Any],
-                      timeout: int = 10) -> aiohttp.ClientResponse:
-            resp = await self.session.put(url, data=data, headers=headers,
-                                          timeout=timeout)
+                      json: Optional[Any] = None, timeout: int = 10
+                      ) -> aiohttp.ClientResponse:
+            resp = await self.session.put(url, data=data, json=json,
+                                          headers=headers, timeout=timeout)
             await _raise_for_status(resp)
             return resp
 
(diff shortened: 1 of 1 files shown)


amartsinovich-dialpad (approved):
LGTM

jonathan-johnston (changes requested):
I'm not sure why this is required because `aiohttp` simply serializes the JSON into `data` anyway. Can we not just serialize into `data` ourselves? Additionally it doesn't look like the `json` argument changes the `headers` in `aiohttp` either.

https://github.com/aio-libs/aiohttp/blob/2a9d5e9ecacbddad0fba48480aa5417ee9c8f8d6/aiohttp/client.py#L353

(Kyle and I are Slack-ing offline but I'm just posting this here while we discuss)

TheKevJames (changes requested):
I agree with @jonathan-johnston -- better to just make use of the `data` argument which already exists rather than opening up the potential for footguns; if `data` and `json` are both non-null, `aiohttp` raises an Exception. It'd be much easier to just make this completely impossible.

Note that from the client side, the following two pieces of code are identical:
```python
await session.post(..., json=payload)
# vs
await session.post(..., data=json.dumps(payload))
```

kylepad:
> Note that from the client side, the following two pieces of code are identical:
```python
await session.post(..., json=payload)
# vs
await session.post(..., data=json.dumps(payload))
```

^ That is an incorrect assumption, there is a bunch of extra stuff that happens behind the scenes in both `requests` and `aiohttp` based on whether `data` or `json` is used.
Additionally the `json` argument is standard part of every single `requests` style library in almost every language, not just python.  We're breaking convention here by not exposing it to users.  Why should we force them to figure out how to correctly send JSON when every other library in the world (including the one's we're using behind the scenes) do all the heavy lifting for them. 
Additionally, I tried the above command `await session.post(..., data=json.dumps(payload))` and it resulted in Google returning a 400 because the payload was incorrectly formatted, every single related Stackoverflow article pointed to this being the solution, after digging through the source code of both libraries, I stand by that.

TheKevJames:
Would it be possible to do this in the opposite way, then? eg. exposing `json` and hiding `data`?

Note that this `Session` object is really meant to be _for internal gcloud-aio usage only_. I don't think we ever really planned for users to be making use of this directly in either case... what's our use-case for this?

kylepad:
> what's our use-case for this?

Good question, we're using this pattern in the autoscaler:
```python
async def _query_loadbalancing(selector: str, metric: str) -> Any:
    scope = 'https://www.googleapis.com/auth/monitoring'
    # can we get the stats just from the current region

    async with Token(scopes=[scope]) as token:
        project = await token.get_project()

        # MQL (Monitoring Query Language)
        # https://cloud.google.com/monitoring/mql/reference
        query = (
            'fetch https_lb_rule'
            f'| metric "loadbalancing.googleapis.com/{metric}"'
            '| align rate(1m)'
            '| every 1m'
            f'| filter (resource.backend_target_name="{selector}")'
        )

        url = f'{GCP_MONITORING_BASE_URL}/v3/projects/{project}/timeSeries:query'  # pylint: disable=line-too-long
        headers = {
            'Authorization': f'Bearer {await token.get()}',
        }
        body = {
            'query': str(query),
            'pageSize': 100,
        }

        response = await token.session._session.post(  # pylint: disable=protected-access
            url, headers=headers, json=body, timeout=10)
        response.raise_for_status()
        return await response.json()
```
and while:
```python
 token.session._session.post(  # pylint: disable=protected-access
```
is _fine_, it seems to me like we should either be passing all `kwargs` through to the underlying library, or at the very least exposing the `json` argument.
Honestly if you're ok with it I would propose passing through `kwargs` in all the `session.<http_method>` methods, that way users could use features of the underlying libraries with Google Auth if they desire. 🤷

TheKevJames:
So that really should work with the `data` argument -- it's the same thing we're doing in every single `gcloud-aio` library already. Does the following throw an error? If so, what?
```python
headers = {
    'Authorization': f'Bearer {await token.get()}',
    'Content-Type': 'application/json',
 }
body = json.dumps({
    'query': str(query),
    'pageSize': 100,
}).encode('utf-8')

response = await token.session._session.post(  # pylint: disable=protected-access
    url, headers=headers, json=body, timeout=10)
```

I basically just copied the above from a random `gcloud-aio` lib -- it looks like this usage is no different that any other use-case, at least at first glance.

On that note, I notice you're reaching into the `gcloud-aio` internals to accomplish the above... would it be easier to just write a `gcloud-aio-monitoring`? Looks like you've already implemented 99% of it, just within the autoscaler app instead of in the library itself.

kylepad:
> So that really should work with the data argument -- it's the same thing we're doing in every single gcloud-aio library already.

Why is this a good thing?  This is exactly the opposite of the DRY principle?  We're repeating code that already exists in an underlying library in a bunch of places, and that's supposed to be an argument against cleaning it up?  

What I'm trying to say here is that I think this:
```python
        headers = {
            'Authorization': f'Bearer {await token.get()}',
        }
        body = {
            'query': str(query),
            'pageSize': 100,
        }
        response = await token.session.post(
             url, headers=headers, json=body)
```
is simpler than:
```python
headers = {
    'Authorization': f'Bearer {await token.get()}',
    'Content-Type': 'application/json',
 }
body = {
    'query': str(query),
    'pageSize': 100,
}
response = await token.session.post(
     url, headers=headers, 
     data=json.dumps(body).encode('utf-8')
)
```
And applies DRY, which also implies Don't Repeat Others (like the underlying libraries here...).  If we're doing this^ in a bunch of places in this library, then we can clean it up, be it in this commit or in a follow up...
In what world is that not a good thing?
Why do I need to implement json conversion that is available in the underlying library for me?  
Why on earth is simplifying our code in this library by delegating to another library we're already using a bad thing???

PR #240, opened 2020-11-06
feat(datastore): Return mutation results in commit

### LINKED ISSUE #142: Datastore methods don't return the mutations results
AFAIK, it doesn't seem possible to retrieve the mutations results when using `insert`... or `operate` or `commit`.

It looks like there are some return statements missing here and there:

https://github.com/talkiq/gcloud-aio/blob/18612a995942dfc789cabe9d50549897330b0e9d/datastore/gcloud/aio/datastore/datastore.py#L375

https://github.com/talkiq/gcloud-aio/blob/18612a995942dfc789cabe9d50549897330b0e9d/datastore/gcloud/aio/datastore/datastore.py#L188

Let me know if I can help with a PR

KevinGDialpad:
Fixes #142 

I followed the same model as `lookup()` where it returns a dictionary that matches the JSON returned by the API.

One other idea I had was to create a [CommitResponse](https://github.com/KevinGDialpad/gcloud-aio/blob/return_mutation_results_in_datastore_commit/datastore/gcloud/aio/datastore/response.py) class and have `commit()` return an instance of that. Then the calling code doesn't have to deal with dictionaries:
```
insert_result = await ds.insert(key, {'value': 12})
saved_key = insert_result.mutation_results[0].key
```
Maybe it would be good to have a big breaking change at some point, and make everything objects. But today isn't the day :)


TheKevJames:
This looks great, thanks @KevinGDialpad !

PR #241, opened 2020-11-10
draft(pubsub): Subscriber Client

amartsinovich-dialpad:
This is my first take on `SubscriberClient`. That was a surprisingly hard thing to even think about. I feel like it takes a very small step to slip into full blown clone of what google has.

Please let me know what you think, this draft is not to be merged, it's for discussing general direction so that I don't go too far off a sensible path.

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -1,18 +1,36 @@
+import json
+import os
 from typing import Any
+from typing import AnyStr
 from typing import Callable
 from typing import Dict
+from typing import IO
 from typing import List
 from typing import Optional
 from typing import Tuple
 from typing import Union
 
+from gcloud.aio.auth import AioSession
 from gcloud.aio.auth import BUILD_GCLOUD_REST  # pylint: disable=no-name-in-module
+from gcloud.aio.auth import Token
 from google.api_core import exceptions
 from google.cloud.pubsub_v1.subscriber.futures import StreamingPullFuture
 from google.cloud.pubsub_v1.subscriber.message import Message
 from google.cloud.pubsub_v1.types import FlowControl as _FlowControl
 
 from .subscriber_message import SubscriberMessage
+
+API_ROOT = 'https://pubsub.googleapis.com'
+VERIFY_SSL = True
+
+SCOPES = [
+    'https://www.googleapis.com/auth/pubsub'
+]
+
+PUBSUB_EMULATOR_HOST = os.environ.get('PUBSUB_EMULATOR_HOST')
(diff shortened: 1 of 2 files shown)


pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -132,60 +166,75 @@ def create_subscription(self,
...
+            body = {
+                'ackIds': ack_ids,
+                'ackDeadlineSeconds': ack_deadline_seconds,
+            }
+            s = session or self.session
+            resp = await s.post(url, data=json.dumps(body).encode(), headers=headers)
+            return
+
+
+        async def subscribe(self,
```
TheKevJames, line 215: I'm not sure we need to worry about creating this method -- I think the goal here should be mostly to implement _just the API interface_ and leave the actual scheduling outside of the domain of this library. I think our main goal should be avoiding re-inventing the wheel here -- the exact threading models used in clients of this library are not something we can predict and each app will probably have different requirements; rather let them implement those cases and/or offload this to scheduling libraries.

We simply need to provide the primitives, not the higher-level operations.

If we do want to provide this sort of higher-level interface _as well_, it would probably be best to separate it out entirely from the API code -- perhaps a `SubscriptionManager` class or such. And perhaps only available in the `-aio-` lib, since it makes no sense in the case of a single thread? That said, I think we could just as easily toss this sort of thing in the docs for now and worry about creating the full interface only once we've explored other options and seen what it is we actually end up needing downstream and what we can offload to a more Broadway-like lib.

amartsinovich-dialpad: Yeah, that makes sense. My reasoning is that we'll have to have this interface distributed to all our services, so I need to put it _somewhere_. For the purpose of starting a discussion, that seemed like an ok place :)

TheKevJames: Makes sense! I'm curious how many of our apps will need the same code here and what we might need to customize -- for example, being able to customize the size of the `asyncio.Queue` is probably very important to clients (as an aside, check out https://nullprogram.com/blog/2020/05/24/ for an awesome read on some gotchas here), though maybe that's just a matter of passing in `max_messages` to the constructor?

amartsinovich-dialpad: wow that's a great article. I now feel a little better for using queues at the first place. I'll be playing with `max_message` and what not as first load tests reveals extreme inefficiency of this solution. Feel free to toss any ideas of realization, I'll test them out.

amartsinovich-dialpad:
Closing this as the design is already outdated

PR #243, opened 2020-11-19
feat(pubsub) Add rest endpoints to pull, ack and nack messages

amartsinovich-dialpad:
REST endpoints + grpc-free `SubscriberMessage`. 
`ack_callback` is here to let any message "producer" define how message can be acked by clients. There should be a name for this pattern, but I don't know it :(

- [x]  Manually tested

### DIFF
--- a/pubsub/README.rst
+++ b/pubsub/README.rst
@@ -19,18 +19,10 @@ Usage
 Subscriber
 ~~~~~~~~~~
 
-The official Google publisher returns a future which is mostly useable as-is.
-``gcloud-rest`` returns the future without change while ``gcloud-aio`` wraps it in ``asyncio`` terms.
+At the moment, ``gcloud-{aio,rest}-pubsub`` only provides a set of handles
+to call Pubsub REST API.
 
-An HTTP-oriented version, in keeping with the other ``gcloud-{aio,rest}-*`` libraries,
-will likely be coming soon -- though our current approach works reasonably well
-for allowing the official ``grpc`` client to be used under ``asyncio``, we
-continue to see threading oddities now and again which we've not been able to
-solve. As such, we do not wholeheartedly recommend using the
-``SubscriberClient`` of this library in production, though a resilient enough
-environment for your use-case may be possible.
-
-Here's the rough usage pattern for subscribing:
+Here's the rough usage pattern for using ``SubscriberClient``:
 
 .. code-block:: python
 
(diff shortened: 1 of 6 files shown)


pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -100,18 +118,30 @@ def _callback_wrapper(message: Message) -> None:
             return _callback_wrapper
 
 else:
```
TheKevJames, line 120: Now that we're turning this into a proper http interface rather than the grpc pass-through thing, you should be able to do away with a lot of these conditionals and build this for both `gcloud-aio` and `gcloud-rest` with the same code!

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -132,60 +162,56 @@ def create_subscription(self,
...
+                'maxMessages': max_messages,
+            }
+            encoded = json.dumps(body).encode()
+            s = session or self.session
+            resp = await s.post(url, data=encoded, headers=headers)
+            resp = await resp.json()
+            messages = []
+            for m in resp['receivedMessages']:
+                messages.append(SubscriberMessage.from_api_dict(m))
+            return messages
```
TheKevJames, line 183: performance/style nit: list comprehensions are preferable to repeatedly appending to a list since it has a lower performance overhead and tends to be easier to read:
```python
return [
    SubscriberMessage.from_api_dict(m)
    for m in resp['receivedMessages']
]
```

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -132,60 +162,56 @@ def create_subscription(self,
...
+            headers = await self._headers()
+            body = {
+                'maxMessages': max_messages,
+            }
+            encoded = json.dumps(body).encode()
+            s = session or self.session
+            resp = await s.post(url, data=encoded, headers=headers)
+            resp = await resp.json()
+            messages = []
+            for m in resp['receivedMessages']:
```
TheKevJames, line 181: Is this key guaranteed to exist if no messages exist? You may want to do `resp.get('receivedMessages', [])` if not.

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -132,60 +162,56 @@ def create_subscription(self,
...
-            return sub_keepalive
-
-        def run_forever(self, sub_keepalive: StreamingPullFuture) -> None:
+            url = f'{API_ROOT}/v1/{subscription}:pull'
+            headers = await self._headers()
+            body = {
+                'maxMessages': max_messages,
+            }
+            encoded = json.dumps(body).encode()
+            s = session or self.session
```
TheKevJames, line 177: Looks like a missing cast here, since `session` is an `aiohttp.ClientSession` but `self.session` is an AioSession. You probably want: `s = AioSession(session) if session else self.session`.

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -132,60 +162,56 @@ def create_subscription(self,
             except exceptions.AlreadyExists:
                 pass
 
-        def subscribe(self,
-                      subscription: str,
-                      callback: Callable[[SubscriberMessage], Any],
-                      *,
-                      flow_control: Union[FlowControl, Tuple[int, ...]] = ()
-                      ) -> StreamingPullFuture:
+        async def pull(self, subscription: str, max_messages: int,
```
TheKevJames, line 165: nit: would be awesome to get comments linking to the official docs for each method here, as we do for other gcloud-aio libs. Eg. this one would be:
```python
# https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/pull
async def pull(...)
```

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -132,60 +162,56 @@ def create_subscription(self,
...
-
-            return sub_keepalive
-
-        def run_forever(self, sub_keepalive: StreamingPullFuture) -> None:
+            url = f'{API_ROOT}/v1/{subscription}:pull'
+            headers = await self._headers()
+            body = {
+                'maxMessages': max_messages,
+            }
+            encoded = json.dumps(body).encode()
```
TheKevJames, line 176: Please specify `utf-8` encoding here, since otherwise this operation is system-dependant (ie. on systems with odd locales, this will encode incorrectly)

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -132,60 +162,56 @@ def create_subscription(self,
...
-            return _callback_wrapper
+            url = f'{API_ROOT}/v1/{subscription}:acknowledge'
+            headers = await self._headers()
+            body = {
+                'ackIds': ack_ids,
+            }
+            encoded = json.dumps(body).encode()
+            s = session or self.session
+            await s.post(url, data=encoded, headers=headers)
+            return
```
TheKevJames, line 198: nit: unnecessary `return` here, Python will already return `None` when you hit the end of a method.

pubsub/gcloud/aio/pubsub/subscriber_message.py
```diff
@@ -1,75 +1,137 @@
...
+            data = json.loads(
+                base64.b64decode(received_message['message']['data']))
+            attributes = received_message['message']['attributes']
+            publish_time: datetime.datetime = parse_publish_time(
+                received_message['message']['publishTime'])
+            return SubscriberMessage(ack_id=ack_id, message_id=message_id,
+                                     publish_time=publish_time, data=data,
+                                     attributes=attributes)
+
+        @staticmethod
```
TheKevJames, line 119: I'm not sure these convenience methods are necessary. Do we have any examples of code this will make significantly simpler? Eg. looks like without them acking is:
```python
await subscriber.ack(subscription, [m.ack_id for m in messages])
```
Which seems simple enough that I wouldn't want to complicate things without a good reason.

TheKevJames (changes requested):

pubsub/gcloud/aio/pubsub/subscriber_client.py
amartsinovich-dialpad: Do you think we can get rid of having `pass-through` thing for `gcloud-rest`? That will lock anyone using it to the current version. I'm all for it, just double checking :)

pubsub/gcloud/aio/pubsub/subscriber_client.py
amartsinovich-dialpad: I... I... I just need some more time to get used to multi-line list comprehensions.

pubsub/gcloud/aio/pubsub/subscriber_client.py
amartsinovich-dialpad: yep, no messages will be empty list

pubsub/gcloud/aio/pubsub/subscriber_client.py
amartsinovich-dialpad: Eagle eye!

pubsub/gcloud/aio/pubsub/subscriber_message.py
amartsinovich-dialpad: See draft for subscriber implementation code

pubsub/gcloud/aio/pubsub/subscriber_client.py
TheKevJames: Yeah, I think we could go ahead and release this whole thing as a major version bump -- totally reasonable for folks to have to pin to the current major until they're ready to change the approach to match :)

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -1,191 +1,162 @@
...
+            'Authorization': f'Bearer {token}'
+        }
+
+    # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/create
+    async def create_subscription(self,
+                                  subscription: str,
+                                  topic: str,
+                                  *,
+                                  session: Optional[Session] = None,
+                                  **kwargs: Dict[str, Any]) -> Dict[str, Any]:
```
TheKevJames, line 64: Rather than catch-all passing the splatted `kwargs` through to the request body, it'd probably be better to be explicit on this argument; perhaps something like `body: Dict[str, Any]` or `config`, etc? Otherwise, folks will need to manually serialize/deserialize their request between the dictionary form mentioned in the API docs and Python argument form, which in many cases might involve some ugly parsing. Additionally, this may swallow more arguments than we'd like in the future, if we augment this method signature -- it'd be difficult to add, say, a `timeout` argument to this method without running into version conflicts where old versions try to use the `timeout` value as part of the actual request body!

Eg. by writing `**kwargs` here, you're only accepting arguments of the form `.create_subscription(..., foo='foo', bar='bar'), rather than the more JSON-API form of `{'foo': 'foo', 'bar': 'bar'}`.

pubsub/gcloud/aio/pubsub/subscriber_message.py
```diff
@@ -1,75 +1,63 @@
...
-
-    @property
-    def publish_time(self) -> datetime.datetime:
-        published: datetime.datetime = self._message.publish_time
-        return published
-
-    @property
-    def ordering_key(self) -> str:
-        return str(self._message.ordering_key)
+    def from_api_dict(received_message: Dict[str, Any]
```
TheKevJames, line 32: Please this method `from_repr()` to match the other `gcloud-aio` APIs. See [an example](https://github.com/talkiq/gcloud-aio/blob/67948064935b592a1e26c3c14039c98be93b7d0a/datastore/gcloud/aio/datastore/key.py#L27-L41). Similarly, building a `to_repr()` for convenience while you're here would be fantastic.

amartsinovich-dialpad: omg, I've just realized that I'm doing `json.loads` even though pubsub message can contain anything.
Also, will do repr things

TheKevJames: Yeah, that one is a Vi-specific thing; we just-so-happen to only send around json-formatted messages.

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -1,191 +1,164 @@
...
+        return {
+            **default_headers,
+            'Authorization': f'Bearer {token}'
+        }
+
+    # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/create
+    async def create_subscription(self,
+                                  subscription: str,
+                                  topic: str,
+                                  body: Optional[Dict[str, Any]],
```
allandialpad, line 62: is there a reason you made this explicit `body=None`

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -1,191 +1,164 @@
...
+        body = {} if not body else body
+        url = f'{API_ROOT}/v1/{subscription}'
+        headers = await self._headers()
+        payload: Dict[str, Any] = {
+            'topic': topic,
+            **body
+        }
+        encoded = json.dumps(payload).encode()
+        s = AioSession(session) if session else self.session
+        resp = await s.put(url, data=encoded, headers=headers)
```
allandialpad, line 78: still not really familiar with the patterns with sessions, is `raise_for_status` useful here?

amartsinovich-dialpad: `raise_for_status` is executed down the stack https://github.com/talkiq/gcloud-aio/blob/master/auth/gcloud/aio/auth/session.py#L115

pubsub/gcloud/aio/pubsub/subscriber_client.py
amartsinovich-dialpad: ah, good catch, looks like I forgot default `None` value

amartsinovich-dialpad:
@TheKevJames @allandialpad @jonathan-johnston Hey folks can I get a final round? This PR should be ready for merge at this point

pubsub/README.rst
```diff
@@ -19,174 +19,27 @@ Usage
...
 
 .. code-block:: python
 
     from gcloud.aio.pubsub import SubscriberClient
     from gcloud.aio.pubsub import SubscriberMessage
 
     client = SubscriberClient()
-    # create subscription if it doesn't already exist
+    # create subscription
     client.create_subscription(
```
TheKevJames, line 34: Please `await` these in the docs (the `awaits` will get removed for the `gcloud-rest-*` version of the docs)

TheKevJames (approved):
Just a docs tweak, otherwise this looks fantastic!

pubsub/README.rst
```diff
@@ -19,174 +19,27 @@ Usage
...
-    example_flow_control = FlowControl(
-        max_bytes=100*1024*1024,
-        max_messages=1,
-        max_lease_duration=10,
-        max_duration_per_lease_extension=0,
-    )
-
-    keep_alive = client.subscribe(
+    # pull messages
+    await client.pull(
```
jonathan-johnston, line 39: Maybe you want to include here a variable to catch the return value to demonstrate that you now pull messages directly? Something like this?

```python
messages: List[SubscriberMessage] = await client.pull(...)
```

jonathan-johnston (approved):

jonathan-johnston (approved):

TheKevJames (approved):

PR #259, opened 2020-12-02
fix(datastore): Expose MutationResult as datastore.MutationResult

KevinGDialpad:
_... no comment ..._


datastore/gcloud/aio/datastore/__init__.py
```diff
@@ -20,6 +20,7 @@
 from gcloud.aio.datastore.key import Key
 from gcloud.aio.datastore.key import PathElement
 from gcloud.aio.datastore.lat_lng import LatLng
+from gcloud.aio.datastore.mutation import MutationResult
```
TheKevJames, line 23: We'll also want this added to the __all__ block below, which acts as the source of truth for what we want our public interface to be.

TheKevJames (changes requested):

TheKevJames (approved):
Thanks!

Issue #242, opened 2020-11-11
list is not a supported value type


ArcLightSlavik:
I'm trying to add values to datastore with a list

```
_id = 1
rec = {
    5732415771279360: [
        'Hello'
    ]
}
    
key = self.create_key(_id)
result = await self.datastore.client.upsert(key, rec)
return result
```

the rec above returns
```
E           NotImplementedError: <class 'list'> is not a supported value type (any of: {<class 'bool'>: <TypeName.BOOLEAN: 'booleanValue'>, <class 'bytes'>: <TypeName.BLOB: 'blobValue'>, <class 'datetime.datetime'>: <TypeName.TIMESTAMP: 'timestampValue'>, <class 'float'>: <TypeName.DOUBLE: 'doubleValue'>, <class 'int'>: <TypeName.INTEGER: 'integerValue'>, <class 'gcloud.aio.datastore.lat_lng.LatLng'>: <TypeName.GEOPOINT: 'geoPointValue'>, <class 'str'>: <TypeName.STRING: 'stringValue'>, <class 'NoneType'>: <TypeName.NULL: 'nullValue'>, <class 'gcloud.aio.datastore.key.Key'>: <TypeName.KEY: 'keyValue'>, <class 'gcloud.aio.datastore.array.Array'>: <TypeName.ARRAY: 'arrayValue'>, <class 'gcloud.aio.datastore.entity.Entity'>: <TypeName.ENTITY: 'entityValue'>, <class 'teyb.shared.index_datastore.NoIndexEntity'>: <TypeName.ENTITY: 'entityValue'>})
```

I've tried switching the list to an Array object you provide but it returns `E   AttributeError: 'str' object has no attribute 'to_repr'`
```
class NoIndexValue(Value):
    def __init__(self, value: Any, exclude_from_indexes: bool = True) -> None:
        super().__init__(value, exclude_from_indexes)
        self.value = value
        self.excludeFromIndexes = exclude_from_indexes

    def to_repr(self) -> Dict[str, Any]:
        if isinstance(self.value, list):
            self.value = array.Array(items=self.value)

        return super().to_repr()
```

KevinGDialpad:
As we can see [in the API docs](https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/runQuery#ArrayValue), ArrayValue is an array of Values, and you're passing a list of strings. So it should be something like `self.value = array.Array(items=[Value(s) for s in self.value])`

[...]

PR #260, opened 2020-12-02
fix(datastore): Explicitly ask for Values when creating an Array

KevinGDialpad:
Difficult to find a good name for that one.

I noticed that while looking at #242. Static checkers can help prevent this type of problem now.


TheKevJames (approved):
Makes sense, good catch!

PR #261, opened 2020-12-02
fix(datastore) Handle case where commit API doesn't return indexUpdates

KevinGDialpad:
The [docs](https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/commit#response-body) are misleading: the API can just not include 'indexUpdates' instead of setting the number to zero.

### DIFF
--- a/datastore/gcloud/aio/datastore/datastore.py
+++ b/datastore/gcloud/aio/datastore/datastore.py
@@ -193,19 +193,13 @@ class Datastore:
         s = AioSession(session) if session else self.session
         resp = await s.post(url, data=payload, headers=headers,
                             timeout=timeout)
-        response_body: Dict[str, Any] = await resp.json()
+        data: Dict[str, Any] = await resp.json()
 
-        commit_results = {
+        return {
             'mutationResults': [self.mutation_result_kind.from_repr(r)
-                                for r in response_body.get('mutationResults',
-                                                           [])
-                                ],
+                                for r in data.get('mutationResults', [])],
+            'indexUpdates': data.get('indexUpdates', 0),
         }
-
-        if 'indexUpdates' in response_body:
-            commit_results['indexUpdates'] = response_body['indexUpdates']
-
-        return commit_results
 
     # https://cloud.google.com/datastore/docs/reference/admin/rest/v1/projects/export
     async def export(self, output_bucket_prefix: str,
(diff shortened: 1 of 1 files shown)


datastore/gcloud/aio/datastore/datastore.py
```diff
@@ -194,14 +193,20 @@ async def commit(self, mutations: List[Dict[str, Any]],
         s = AioSession(session) if session else self.session
         resp = await s.post(url, data=payload, headers=headers,
                             timeout=timeout)
-        data: Dict[str, Any] = await resp.json()
+        response_body: Dict[str, Any] = await resp.json()
 
-        return {
+        commit_results = {
```
KevinGDialpad, line 198: It was a pain to format it correctly. I'm not allowed another local variable so having a `mutation_results` variable isn't an option. And extracting a method to do `return get_commit_results(resp.json())` seems overkill.

TheKevJames: Might be simpler if we treat `indexUpdates` as having a default value rather than being wholely optional. What does a missing value indicate -- `0`, right? So how about:
```python
        return {
            'mutationResults': [self.mutation_result_kind.from_repr(r)
                                for r in data.get('mutationResults', [])],
            'indexUpdates': data.get('indexUpdates', 0),
        }
```
?

TheKevJames (changes requested):

datastore/gcloud/aio/datastore/datastore.py
KevinGDialpad: I thought about that as well, but then I didn't want the library to guess what the API meant. I'd rather it be dumb and report exactly what the API returned.

TheKevJames: Hey @KevinGDialpad -- sorry I lost track of this PR.

Looking through the docs, there doesn't seem to be any semantic meaning between null and zero in this potential API response. Google actually does this in a few places, where occasionally their APIs return null instead of zero but are meant to be treated semantically the same. As a general rule, in `gcloud-aio` we've been trying to help gloss over these issues (eg. by adding defaults) to help avoid app developers from needing to have this API-parsing logic scattered in their code.

Let's set this to either`data.get('indexUpdates', 0)` or `data.get('indexUpdates')`, depending on whether it's worth distinguishing the null case -- I'm inclined to do the former as per their docs, but if you've seen any case where null indeed does not mean zero then we can go for the latter.

KevinGDialpad: Okay, I suppose that if it happens in other places as well, it makes sense to do the same thing. And in that case, using 0 is probably best. I'll make the change.

TheKevJames (approved):

PR #265, opened 2020-12-07
feat(pubsub): subcribe()

amartsinovich-dialpad:
The final PR for HTTP-bases subscriber.

This is an opportunity to take another look at the code, complete any hanging conversation and finally m e r g e.

I will also be testing this on staging and making sure performance is still ok.

tasks:
* [x] test in `egress`
* [x] test in `ctm`
* [x] write docs (F to Kevin's guide to subscriber 😢 )

### DIFF
--- a/pubsub/README.rst
+++ b/pubsub/README.rst
@@ -92,7 +92,7 @@ your application and tweak function parameter to your specific needs. Here's a f
 
 :``num_tasks_per_consumer``:
     how many ``handle`` calls a worker can make until it blocks
-    to wait for them to return. If you process messages independetly from each other you should
+    to wait for them to return. If you process messages independently from each other you should
     be good with the default value of ``1``. If you do something fancy (e.g. aggregate messages
     before processing them), you'll want a higher pool here. You can think of
     ``num_producers * num_tasks_per_consumer`` as an upper limit of how many messages can possibly
--- a/pubsub/README.rst
+++ b/pubsub/README.rst
@@ -19,10 +19,8 @@ Usage
 Subscriber
 ~~~~~~~~~~
 
-At the moment, ``gcloud-{aio,rest}-pubsub`` only provides a set of handles
-to call Pubsub REST API.
-
-Here's the rough usage pattern for using ``SubscriberClient``:
+``gcloud-{aio,rest}-pubsub`` provides ``SubscriberClient``
+as an interface to call pubsub's HTTP API:
 
 .. code-block:: python
 
@@ -39,6 +37,88 @@ Here's the rough usage pattern for using ``SubscriberClient``:
     messages: List[SubscriberMessage] = await client.pull(
         'projects/<project_name>/subscriptions/<subscription_name>',
         max_messages=10)
+
+
+There's also ``gcloud.aio.pubsub.subscribe`` helper function you can use to
+setup a pubsub processing pipeline. It is built with ``asyncio`` and thus only available
+in ``gcloud-aio-pubsub`` package. The usage is fairly simple:
+
+.. code-block:: python
+
+    from gcloud.aio.pubsub import SubscriberClient
+    from gcloud.aio.pubsub import subscribe
+    from gcloud.aio.pubsub.metrics_agent import MetricsAgent
+
+    subscriber_client = SubscriberClient()
+
+    async def handler(message):
+        return
+
+    await subscribe(
+        'projects/<my_project>/subscriptions/<my_subscription>',
+        handler,
+        subscriber_client,
+        num_workers=1,
+        max_messages=100,
+        ack_window=0.3,
+        consumer_pool_size=1,
+        metrics_client=MetricsAgent()
+    )
+
+While defaults are somewhat sensinble, it is highly recommended to performance test
+your application and tweak function parameter to your specific needs. Here's a few hints:
+
+:``handler``:
+    an async function that will be called for each message. It should accept an instance of
+    ``SubscriberMessage`` as its only argument and return ``None`` if the message should
+    be acked.
+
+:``num_workers``:
+    number of workers that will be making ``pull`` requests to pubsub.
+    Please note that a worker will only fetch new batch once the ``handler``
+    was called for each message from the previous batch. This means that running only a single worker
+    will most likely make your application IO bound. If you notice this being an issue don't hesitate
+    to bump this parameter.
+
+:``max_messages``:
+    number of pubsub messages a worker will try to fetch in a single batch. This
+    value is passed to ``pull`` `endpoint`_ as ``maxMessages`` parameter. A rule of thumb here is
+    the faster your handler is the bigger this value should be.
+
+:``ack_window``:
+    ack requests are handled separately and are done in batches. This parameters
+    specifies how often ack requests will be made. Setting it to ``0.0`` will effectively
+    disable batching.
+
+:``consumer_pool_size``:
+    how many ``handle`` calls a worker can make until it blocks
+    to wait for them to return. If you process messages independetly from each other you should
+    be good with the default value of ``1``. If you do something fancy (e.g. aggregate messages
+    before processing them), you'll want a higher pool here. You can think of
+    ``num_workers * consumer_pool_size`` as an upper limit of how many messages can possibly
+    be within your application state at any given moment.
+
+
+``subscribe`` has also an optional ``metrics_client`` argument. You can provide any metrics
+agent that implements the same interface as ``MetricsAgent`` (Datadog client will do 😉)
+and get the following metrics:
+
+- ``pubsub.producer.batch`` - [histogram] actual size of a batch retrieved from pubsub.
+
+- ``pubsub.consumer.failfast`` - [increment] a message was dropped due to its lease being expired.
+
+- ``pubsub.consumer.latency.receive`` - [histogram] how many seconds it took for a message to reach
+  handler after it was published.
+
+- ``pubsub.consumer.succeeded`` - [increment] ``handler`` call was successfull.
+
+- ``pubsub.consumer.failed`` - [increment] ``handler`` call raised an exception.
+
+- ``pubsub.consumer.latency.runtime`` - [histogram] ``handler`` execution time in seconds.
+
+- ``pubsub.acker.batch.failed`` - [increment] ack request failed.
+
+- ``pubsub.acker.batch`` - [histogram] actual number of messages that was acked in a single request.
 
 
 Publisher
@@ -117,3 +197,5 @@ Please see our `contributing guide`_.
 .. |pythons-rest| image:: https://img.shields.io/pypi/pyversions/gcloud-rest-pubsub.svg?style=flat-square&label=python (rest)
     :alt: Python Version Support (gcloud-rest-pubsub)
     :target: https://pypi.org/project/gcloud-rest-pubsub/
+
+.. _endpoint: https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/pull#request-body
--- a/pubsub/README.rst
+++ b/pubsub/README.rst
@@ -100,7 +100,7 @@ your application and tweak function parameter to your specific needs. Here's a f
 
 
 ``subscribe`` has also an optional ``metrics_client`` argument. You can provide any metrics
-agent that implements the same interface as ``MetricsAgent`` (Datadog client will do 😉)
+agent that implements the same interface as ``MetricsAgent`` (Datadog client will do ;) )
 and get the following metrics:
 
 - ``pubsub.producer.batch`` - [histogram] actual size of a batch retrieved from pubsub.
--- a/pubsub/README.rst
+++ b/pubsub/README.rst
@@ -58,14 +58,14 @@ in ``gcloud-aio-pubsub`` package. The usage is fairly simple:
         'projects/<my_project>/subscriptions/<my_subscription>',
         handler,
         subscriber_client,
-        num_workers=1,
-        max_messages=100,
+        num_producers=1,
+        max_messages_per_producer=100,
         ack_window=0.3,
-        consumer_pool_size=1,
+        num_tasks_per_consumer=1,
         metrics_client=MetricsAgent()
     )
 
-While defaults are somewhat sensinble, it is highly recommended to performance test
+While defaults are somewhat sensible, it is highly recommended to performance test
 your application and tweak function parameter to your specific needs. Here's a few hints:
 
 :``handler``:
@@ -73,14 +73,14 @@ your application and tweak function parameter to your specific needs. Here's a f
     ``SubscriberMessage`` as its only argument and return ``None`` if the message should
     be acked.
 
-:``num_workers``:
+:``num_producers``:
     number of workers that will be making ``pull`` requests to pubsub.
     Please note that a worker will only fetch new batch once the ``handler``
     was called for each message from the previous batch. This means that running only a single worker
     will most likely make your application IO bound. If you notice this being an issue don't hesitate
     to bump this parameter.
 
-:``max_messages``:
+:``max_messages_per_producer``:
     number of pubsub messages a worker will try to fetch in a single batch. This
     value is passed to ``pull`` `endpoint`_ as ``maxMessages`` parameter. A rule of thumb here is
     the faster your handler is the bigger this value should be.
@@ -90,12 +90,12 @@ your application and tweak function parameter to your specific needs. Here's a f
     specifies how often ack requests will be made. Setting it to ``0.0`` will effectively
     disable batching.
 
-:``consumer_pool_size``:
+:``num_tasks_per_consumer``:
     how many ``handle`` calls a worker can make until it blocks
     to wait for them to return. If you process messages independetly from each other you should
     be good with the default value of ``1``. If you do something fancy (e.g. aggregate messages
     before processing them), you'll want a higher pool here. You can think of
-    ``num_workers * consumer_pool_size`` as an upper limit of how many messages can possibly
+    ``num_producers * num_tasks_per_consumer`` as an upper limit of how many messages can possibly
     be within your application state at any given moment.
 
 
--- a/pubsub/README.rst
+++ b/pubsub/README.rst
@@ -40,8 +40,8 @@ as an interface to call pubsub's HTTP API:
 
 
 There's also ``gcloud.aio.pubsub.subscribe`` helper function you can use to
-setup a pubsub processing pipeline. It is built with ``asyncio`` and thus only available
-in ``gcloud-aio-pubsub`` package. The usage is fairly simple:
+setup a pubsub processing pipeline. It is built with ``asyncio`` and thus only
+available in ``gcloud-aio-pubsub`` package. The usage is fairly simple:
 
 .. code-block:: python
 
@@ -65,60 +65,68 @@ in ``gcloud-aio-pubsub`` package. The usage is fairly simple:
         metrics_client=MetricsAgent()
     )
 
-While defaults are somewhat sensible, it is highly recommended to performance test
-your application and tweak function parameter to your specific needs. Here's a few hints:
+While defaults are somewhat sensible, it is highly recommended to performance
+test your application and tweak function parameter to your specific needs.
+Here's a few hints:
 
 :``handler``:
-    an async function that will be called for each message. It should accept an instance of
-    ``SubscriberMessage`` as its only argument and return ``None`` if the message should
-    be acked.
+    an async function that will be called for each message. It should accept an
+    instance of ``SubscriberMessage`` as its only argument and return ``None``
+    if the message should be acked. An exception raised within the handler will
+    result in the message being left to expire, and thus it will be redelivered
+    according to your subscription's ack deadline.
 
 :``num_producers``:
-    number of workers that will be making ``pull`` requests to pubsub.
-    Please note that a worker will only fetch new batch once the ``handler``
-    was called for each message from the previous batch. This means that running only a single worker
-    will most likely make your application IO bound. If you notice this being an issue don't hesitate
-    to bump this parameter.
+    number of workers that will be making ``pull`` requests to pubsub. Please
+    note that a worker will only fetch new batch once the ``handler`` was called
+    for each message from the previous batch. This means that running only a
+    single worker will most likely make your application IO bound. If you notice
+    this being an issue don't hesitate to bump this parameter.
 
 :``max_messages_per_producer``:
     number of pubsub messages a worker will try to fetch in a single batch. This
-    value is passed to ``pull`` `endpoint`_ as ``maxMessages`` parameter. A rule of thumb here is
-    the faster your handler is the bigger this value should be.
+    value is passed to ``pull`` `endpoint`_ as ``maxMessages`` parameter. A rule
+    of thumb here is the faster your handler is the bigger this value should be.
 
 :``ack_window``:
     ack requests are handled separately and are done in batches. This parameters
-    specifies how often ack requests will be made. Setting it to ``0.0`` will effectively
-    disable batching.
+    specifies how often ack requests will be made. Setting it to ``0.0`` will
+    effectively disable batching.
 
 :``num_tasks_per_consumer``:
-    how many ``handle`` calls a worker can make until it blocks
-    to wait for them to return. If you process messages independently from each other you should
-    be good with the default value of ``1``. If you do something fancy (e.g. aggregate messages
-    before processing them), you'll want a higher pool here. You can think of
-    ``num_producers * num_tasks_per_consumer`` as an upper limit of how many messages can possibly
-    be within your application state at any given moment.
+    how many ``handle`` calls a worker can make until it blocks to wait for them
+    to return. If you process messages independently from each other you should
+    be good with the default value of ``1``. If you do something fancy (e.g.
+    aggregate messages before processing them), you'll want a higher pool here.
+    You can think of ``num_producers * num_tasks_per_consumer`` as an upper
+    limit of how many messages can possibly be within your application state at
+    any given moment.
 
 
-``subscribe`` has also an optional ``metrics_client`` argument. You can provide any metrics
-agent that implements the same interface as ``MetricsAgent`` (Datadog client will do ;) )
-and get the following metrics:
+``subscribe`` has also an optional ``metrics_client`` argument. You can provide
+any metrics agent that implements the same interface as ``MetricsAgent``
+(Datadog client will do ;) ) and get the following metrics:
 
-- ``pubsub.producer.batch`` - [histogram] actual size of a batch retrieved from pubsub.
+- ``pubsub.producer.batch`` - [histogram] actual size of a batch retrieved from
+  pubsub.
 
-- ``pubsub.consumer.failfast`` - [increment] a message was dropped due to its lease being expired.
+- ``pubsub.consumer.failfast`` - [increment] a message was dropped due to its
+  lease being expired.
 
-- ``pubsub.consumer.latency.receive`` - [histogram] how many seconds it took for a message to reach
-  handler after it was published.
+- ``pubsub.consumer.latency.receive`` - [histogram] how many seconds it took for
+  a message to reach handler after it was published.
 
 - ``pubsub.consumer.succeeded`` - [increment] ``handler`` call was successfull.
 
 - ``pubsub.consumer.failed`` - [increment] ``handler`` call raised an exception.
 
-- ``pubsub.consumer.latency.runtime`` - [histogram] ``handler`` execution time in seconds.
+- ``pubsub.consumer.latency.runtime`` - [histogram] ``handler`` execution time
+  in seconds.
 
 - ``pubsub.acker.batch.failed`` - [increment] ack request failed.
 
-- ``pubsub.acker.batch`` - [histogram] actual number of messages that was acked in a single request.
+- ``pubsub.acker.batch`` - [histogram] actual number of messages that was acked
+  in a single request.
 
 
 Publisher
--- a/pubsub/gcloud/aio/pubsub/__init__.py
+++ b/pubsub/gcloud/aio/pubsub/__init__.py
@@ -5,7 +5,8 @@ from gcloud.aio.pubsub.publisher_client import PublisherClient
 from gcloud.aio.pubsub.subscriber_client import SubscriberClient
 from gcloud.aio.pubsub.utils import PubsubMessage
 from gcloud.aio.pubsub.subscriber_message import SubscriberMessage
+from gcloud.aio.pubsub.subscriber import subscribe
 
 
 __all__ = ['__version__', 'PublisherClient', 'PubsubMessage',
-           'SubscriberClient', 'SubscriberMessage']
+           'SubscriberClient', 'SubscriberMessage', 'subscribe']
--- a/pubsub/gcloud/aio/pubsub/__init__.py
+++ b/pubsub/gcloud/aio/pubsub/__init__.py
@@ -1,12 +1,17 @@
 from pkg_resources import get_distribution
 __version__ = get_distribution('gcloud-aio-pubsub').version
 
+from gcloud.aio.auth import BUILD_GCLOUD_REST
 from gcloud.aio.pubsub.publisher_client import PublisherClient
 from gcloud.aio.pubsub.subscriber_client import SubscriberClient
 from gcloud.aio.pubsub.utils import PubsubMessage
 from gcloud.aio.pubsub.subscriber_message import SubscriberMessage
-from gcloud.aio.pubsub.subscriber import subscribe
 
+if BUILD_GCLOUD_REST:
+    __all__ = ['__version__', 'PublisherClient', 'PubsubMessage',
+               'SubscriberClient', 'SubscriberMessage']
+else:
+    from gcloud.aio.pubsub.subscriber import subscribe
 
-__all__ = ['__version__', 'PublisherClient', 'PubsubMessage',
-           'SubscriberClient', 'SubscriberMessage', 'subscribe']
+    __all__ = ['__version__', 'PublisherClient', 'PubsubMessage',
+               'SubscriberClient', 'SubscriberMessage', 'subscribe']
--- a/pubsub/gcloud/aio/pubsub/__init__.py
+++ b/pubsub/gcloud/aio/pubsub/__init__.py
@@ -7,11 +7,9 @@ from gcloud.aio.pubsub.subscriber_client import SubscriberClient
 from gcloud.aio.pubsub.utils import PubsubMessage
 from gcloud.aio.pubsub.subscriber_message import SubscriberMessage
 
-if BUILD_GCLOUD_REST:
-    __all__ = ['__version__', 'PublisherClient', 'PubsubMessage',
-               'SubscriberClient', 'SubscriberMessage']
-else:
-    from gcloud.aio.pubsub.subscriber import subscribe
+__all__ = ['__version__', 'PublisherClient', 'PubsubMessage',
+           'SubscriberClient', 'SubscriberMessage']
 
-    __all__ = ['__version__', 'PublisherClient', 'PubsubMessage',
-               'SubscriberClient', 'SubscriberMessage', 'subscribe']
+if not BUILD_GCLOUD_REST:
+    from gcloud.aio.pubsub.subscriber import subscribe
+    __all__.append('subscribe')
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -78,6 +78,7 @@ else:
                 except asyncio.TimeoutError:
                     break
                 time_budget -= (time.perf_counter() - start)
+
             try:
                 await subscriber_client.acknowledge(subscription,
                                                     ack_ids=ack_ids)
@@ -136,6 +137,8 @@ else:
             if metrics_client:
                 metrics_client.histogram(
                     'pubsub.consumer.latency.receive',
+                    # publish_time is in UTC Zulu
+                    # https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage
                     time.time() - message.publish_time.timestamp())
 
             asyncio.ensure_future(_fire_callback(
@@ -214,16 +217,12 @@ else:
                                  metrics_client=metrics_client)
                     ))
 
-                _done, pending = await asyncio.wait(
+                await asyncio.wait(
                     tasks, return_when=asyncio.FIRST_COMPLETED)
-                log.info(
-                    'One of subsriber workers exited, stopping the rest...')
-                for task in pending:
-                    task.cancel()
-                await asyncio.wait(pending, return_when=asyncio.ALL_COMPLETED)
+                raise Exception('A subscriber worker shut down unexpectedly!')
             except Exception:
                 log.exception('Subscriber exited')
                 for task in tasks:
                     task.cancel()
                 await asyncio.wait(tasks, return_when=asyncio.ALL_COMPLETED)
-            raise asyncio.CancelledError('One of subscription workers exited.')
+            raise asyncio.CancelledError('Subscriber shut down')
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -210,8 +210,10 @@ else:
                              metrics_client=metrics_client)
                 ))
 
-            await asyncio.wait(
-                tasks, return_when=asyncio.FIRST_COMPLETED)
+            done, _ = await asyncio.wait(tasks,
+                                         return_when=asyncio.FIRST_COMPLETED)
+            for task in done:
+                task.result()
             raise Exception('A subscriber worker shut down unexpectedly!')
         except Exception:
             log.exception('Subscriber exited')
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -60,7 +60,7 @@ else:
                     ack_queue: 'asyncio.Queue[str]',
                     subscriber_client: 'SubscriberClient',
                     ack_window: float,
-                    metrics_client: Optional[MetricsAgent]) -> None:
+                    metrics_client: MetricsAgent) -> None:
         ack_ids: List[str] = []
         while True:
             time_budget = ack_window
@@ -87,8 +87,7 @@ else:
                     'Ack request failed, better luck next batch', exc_info=e)
                 continue
 
-            if metrics_client:
-                metrics_client.histogram('pubsub.acker.batch', len(ack_ids))
+            metrics_client.histogram('pubsub.acker.batch', len(ack_ids))
 
             ack_ids = []
 
@@ -96,20 +95,18 @@ else:
                              message: SubscriberMessage,
                              callback: ApplicationHandler,
                              ack_queue: 'asyncio.Queue[str]',
-                             metrics_client: Optional[MetricsAgent]
+                             metrics_client: MetricsAgent
                              ) -> None:
         try:
             start = time.perf_counter()
             await callback(message)
             await ack_queue.put(message.ack_id)
-            if metrics_client:
-                metrics_client.increment('pubsub.consumer.succeeded')
-                metrics_client.histogram('pubsub.consumer.latency.runtime',
-                                         time.perf_counter() - start)
+            metrics_client.increment('pubsub.consumer.succeeded')
+            metrics_client.histogram('pubsub.consumer.latency.runtime',
+                                     time.perf_counter() - start)
         except Exception:
             log.exception('Application callback raised an exception')
-            if metrics_client:
-                metrics_client.increment('pubsub.consumer.failed')
+            metrics_client.increment('pubsub.consumer.failed')
         finally:
             semaphore.release()
 
@@ -119,7 +116,7 @@ else:
             ack_queue: 'asyncio.Queue[str]',
             ack_deadline_cache: AckDeadlineCache,
             consumer_pool_size: int,
-            metrics_client: Optional[MetricsAgent]) -> None:
+            metrics_client: MetricsAgent) -> None:
         semaphore = asyncio.Semaphore(consumer_pool_size)
         while True:
             await semaphore.acquire()
@@ -128,18 +125,16 @@ else:
 
             ack_deadline = await ack_deadline_cache.get()
             if (time.perf_counter() - pulled_at) >= ack_deadline:
-                if metrics_client:
-                    metrics_client.increment('pubsub.consumer.failfast')
+                metrics_client.increment('pubsub.consumer.failfast')
                 message_queue.task_done()
                 semaphore.release()
                 continue
 
-            if metrics_client:
-                metrics_client.histogram(
-                    'pubsub.consumer.latency.receive',
-                    # publish_time is in UTC Zulu
-                    # https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage
-                    time.time() - message.publish_time.timestamp())
+            metrics_client.histogram(
+                'pubsub.consumer.latency.receive',
+                # publish_time is in UTC Zulu
+                # https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage
+                time.time() - message.publish_time.timestamp())
 
             asyncio.ensure_future(_fire_callback(
                 semaphore,
@@ -155,7 +150,7 @@ else:
             message_queue: MessageQueue,
             subscriber_client: 'SubscriberClient',
             max_messages: int,
-            metrics_client: Optional[MetricsAgent]
+            metrics_client: MetricsAgent
     ) -> None:
         while True:
             try:
@@ -168,9 +163,8 @@ else:
             for m in new_messages:
                 await message_queue.put((m, pulled_at))
 
-            if metrics_client:
-                metrics_client.histogram(
-                    'pubsub.producer.batch', len(new_messages))
+            metrics_client.histogram(
+                'pubsub.producer.batch', len(new_messages))
 
             await message_queue.join()
 
@@ -192,6 +186,7 @@ else:
             ack_deadline_cache = AckDeadlineCache(subscriber,
                                                   subscription,
                                                   ack_deadline_cache_timeout)
+            metrics_client = metrics_client or MetricsAgent()
             tasks = []
             try:
                 tasks.append(asyncio.ensure_future(
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -93,12 +93,12 @@ else:
 
             ack_ids = []
 
-    async def _fire_callback(semaphore: asyncio.Semaphore,
-                             message: SubscriberMessage,
-                             callback: ApplicationHandler,
-                             ack_queue: 'asyncio.Queue[str]',
-                             metrics_client: MetricsAgent
-                             ) -> None:
+    async def _execute_callback(semaphore: asyncio.Semaphore,
+                                message: SubscriberMessage,
+                                callback: ApplicationHandler,
+                                ack_queue: 'asyncio.Queue[str]',
+                                metrics_client: MetricsAgent
+                                ) -> None:
         try:
             start = time.perf_counter()
             await callback(message)
@@ -138,7 +138,7 @@ else:
                 # https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage
                 time.time() - message.publish_time.timestamp())
 
-            asyncio.ensure_future(_fire_callback(
+            asyncio.ensure_future(_execute_callback(
                 semaphore,
                 message,
                 callback,
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -93,8 +93,7 @@ else:
 
             ack_ids = []
 
-    async def _execute_callback(semaphore: asyncio.Semaphore,
-                                message: SubscriberMessage,
+    async def _execute_callback(message: SubscriberMessage,
                                 callback: ApplicationHandler,
                                 ack_queue: 'asyncio.Queue[str]',
                                 metrics_client: MetricsAgent
@@ -109,8 +108,6 @@ else:
         except Exception:
             log.exception('Application callback raised an exception')
             metrics_client.increment('pubsub.consumer.failed')
-        finally:
-            semaphore.release()
 
     async def consumer(
             message_queue: MessageQueue,
@@ -138,13 +135,13 @@ else:
                 # https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage
                 time.time() - message.publish_time.timestamp())
 
-            asyncio.ensure_future(_execute_callback(
-                semaphore,
+            task = asyncio.ensure_future(_execute_callback(
                 message,
                 callback,
                 ack_queue,
                 metrics_client,
             ))
+            task.add_done_callback(lambda _f: semaphore.release())
             message_queue.task_done()
 
     async def producer(
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -1,0 +1,229 @@
+from gcloud.aio.auth import BUILD_GCLOUD_REST
+
+if BUILD_GCLOUD_REST:
+    pass
+else:
+    import asyncio
+    import logging
+    import time
+    from typing import Awaitable
+    from typing import Callable
+    from typing import List
+    from typing import Optional
+    from typing import Tuple
+    from typing import TYPE_CHECKING
+
+    import aiohttp
+    from gcloud.aio.pubsub.subscriber_client import SubscriberClient
+    from gcloud.aio.pubsub.subscriber_message import SubscriberMessage
+    from gcloud.aio.pubsub.metrics_agent import MetricsAgent
+
+    log = logging.getLogger(__name__)
+
+    if TYPE_CHECKING:
+        MessageQueue = asyncio.Queue[Tuple[SubscriberMessage,  # pylint: disable=unsubscriptable-object
+                                           float]]
+    else:
+        MessageQueue = asyncio.Queue
+    ApplicationHandler = Callable[[SubscriberMessage], Awaitable[None]]
+
+    class AckDeadlineCache:
+        def __init__(self, subscriber_client: SubscriberClient,
+                     subscription: str, cache_timout: int):
+            self.subscriber_client = subscriber_client
+            self.subscription = subscription
+            self.cache_timeout = cache_timout
+            self.ack_deadline: float = float('inf')
+            self.last_refresh: float = 0
+
+        async def get(self) -> float:
+            if self.ack_deadline == float('inf') or self.cache_outdated():
+                await self.refresh()
+            return self.ack_deadline
+
+        async def refresh(self) -> None:
+            try:
+                sub = await self.subscriber_client.get_subscription(
+                    self.subscription)
+                self.ack_deadline = float(sub['ackDeadlineSeconds'])
+            except Exception as e:
+                log.warning(
+                    'Failed to refresh ackDeadlineSeconds value', exc_info=e)
+            self.last_refresh = time.perf_counter()
+
+        def cache_outdated(self) -> bool:
+            if (time.perf_counter() - self.last_refresh) > self.cache_timeout:
+                return True
+            return False
+
+    async def acker(subscription: str,
+                    ack_queue: 'asyncio.Queue[str]',
+                    subscriber_client: 'SubscriberClient',
+                    ack_window: float,
+                    metrics_client: Optional[MetricsAgent]) -> None:
+        ack_ids: List[str] = []
+        while True:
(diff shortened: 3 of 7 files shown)


pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+                self.ack_deadline = float(sub['ackDeadlineSeconds'])
+            except Exception as e:
+                log.warning(
+                    'Failed to refresh ackDeadlineSeconds value', exc_info=e)
+            self.last_refresh = time.perf_counter()
+
+        def cache_outdated(self) -> bool:
+            if (time.perf_counter() - self.last_refresh) > self.cache_timeout:
+                return True
+            return False
```
amartsinovich-dialpad, line 57: @TheKevJames @allandialpad we were discussing if this is necessary
For: 
* will work regardless of subscription configuration management approach
Against:
* complexity

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+                    ack_queue.task_done()
+                except asyncio.TimeoutError:
+                    break
+                time_budget -= (time.perf_counter() - start)
+            try:
+                await subscriber_client.acknowledge(subscription,
+                                                    ack_ids=ack_ids)
+            except Exception as e:
+                log.warning(
+                    'Ack request failed, better luck next batch', exc_info=e)
```
amartsinovich-dialpad, line 86: @shaundialpad we were discussing potentially unbound growth of `ack_ids`. My last take:

> I wonder what would be the best way to safeguard against it? Worst case scenario is, I guess, google's outage specifically on ack endpoint. In this case we need to make sure we won't consume all the memory? So maybe only allow head N messages from last iteration to let through?

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+                continue
+
+            if metrics_client:
+                metrics_client.histogram('pubsub.acker.batch', len(ack_ids))
+
+            ack_ids = []
+
+    async def _fire_callback(semaphore: asyncio.Semaphore,
```
amartsinovich-dialpad, line 94: @allandialpad we were discussing naming here.

What I tried to capture here is the idea that we fire those tasks and pretty much forget about them. `task_wrapper` and `callback_wrapper` are a bit too generic for me. What others think?

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+                metrics_client.histogram(
+                    'pubsub.producer.batch', len(new_messages))
+
+            await message_queue.join()
+
+    async def subscribe(subscription: str,  # pylint: disable=too-many-locals
+                        handler: ApplicationHandler,
+                        *,
+                        num_workers: int = 1,
+                        max_messages: int = 100,
```
amartsinovich-dialpad, line 178: @shaundialpad @TheKevJames we were discussing whether `max_messages` should stay as is to signal its origin in API's `maxMessages` parameter.

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+                    'One of subsriber workers exited, stopping the rest...')
+                for task in pending:
+                    task.cancel()
+                await asyncio.wait(pending, return_when=asyncio.ALL_COMPLETED)
+            except Exception:
+                log.exception('Subscriber exited')
+                for task in tasks:
+                    task.cancel()
+                await asyncio.wait(tasks, return_when=asyncio.ALL_COMPLETED)
+            raise asyncio.CancelledError('One of subscription workers exited.')
```
amartsinovich-dialpad, line 229: @jonathan-johnston we were discussing error handling here. My latest take: 
> I added exception handling now, so what should happen:
>1. If any subscriber code throws during tasks creation, we catch it, cancel already created tasks, wait for them to complete and raise CancelledError
>2. If any of the tasks complete, we consider this an unrecoverable crash, cancel all other tasks, wait for them to complete and raise CancelledError.

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+                if metrics_client:
+                    metrics_client.increment('pubsub.consumer.failfast')
+                message_queue.task_done()
+                semaphore.release()
+                continue
+
+            if metrics_client:
+                metrics_client.histogram(
+                    'pubsub.consumer.latency.receive',
+                    time.time() - message.publish_time.timestamp())
```
amartsinovich-dialpad, line 139: @allandialpad we were discussing timezones here?

I think I might be missing something since both timestamps should be in utc?

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+                metrics_client.increment('pubsub.consumer.failed')
+        finally:
+            semaphore.release()
+
+    async def consumer(
+            message_queue: MessageQueue,
+            callback: ApplicationHandler,
+            ack_queue: 'asyncio.Queue[str]',
+            ack_deadline_cache: AckDeadlineCache,
+            consumer_pool_size: int,
```
amartsinovich-dialpad, line 120: @allandialpad we were discussing the nature of consumer pool size. 

So for normal use case I don't expect pool size to ever be more than 1. However, for specific use cases  (such as batching in egress) a bigger pool size is needed to ensure application have sufficient message input flow.

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+            if (time.perf_counter() - self.last_refresh) > self.cache_timeout:
+                return True
+            return False
+
+    async def acker(subscription: str,
+                    ack_queue: 'asyncio.Queue[str]',
+                    subscriber_client: 'SubscriberClient',
+                    ack_window: float,
+                    metrics_client: Optional[MetricsAgent]) -> None:
+        ack_ids: List[str] = []
```
TheKevJames, line 64: nit: initialize this at the beginning of each loop rather than here as well as at the end of the loop, to avoid doing it twice (where it could eventually out of sync) and to make it more clear that this value is meant to be temporary per-loop.

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: I don't think anything has changed in my mind since last I commented; I continue to be inclined to mark this as unnecessary complexity; ideally we should be referring to a `gcloud.aio.pubsub.Subscription()['ackDeadLineSubscription']` object here and only listening for changes in cases where users do `gcloud.aio.pubsub.Subscription().patch(ackDeadlineSubscription=new_value)`.

TheKevJames (changes requested):

pubsub/gcloud/aio/pubsub/subscriber.py
allandialpad: I'd vote for leaving the infrastructure here but just throwing a warning/error. Its possible that there is very common  -yet unused by us- usecase where apps dynamically modify their ack deadline

pubsub/gcloud/aio/pubsub/subscriber.py
allandialpad: executor?

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: @TheKevJames I still completely miss your point on patch. Do you mean `SubscriberClient` by `gcloud.aio.pubsub.Subscription()`? Could you throw some high level pseudo code here? If I drop refresh mechanism I need to replace it with something to get rid of possible bugs caused by desync and I don't see what less complex it could be.

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: @TheKevJames We were actually discussing the growth of `ack_ids` list where we store unsent ids. Our acknowledgement train named `acked` is going full speed and stops only to make a request every `0.3` seconds. If for some reason this request fails – not a bid deal, we just keep all the ids in the list and will send them on next batch. The problem I personally see here is long term outage of google's ack endpoint. In this case `ack_ids` list will keep growing until a) it consumes all memory and crash the app b) ack endpoint is available again and we're good c) we exceed POST body length limit and become a one way application 😱

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: re:exception sounds right to me

graceful shutdown is always nice. Is there any common way to gracefully shutdown asyncio tasks? Quick search didn't yield much results for me...

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+            for m in new_messages:
+                await message_queue.put((m, pulled_at))
+
+            if metrics_client:
+                metrics_client.histogram(
+                    'pubsub.producer.batch', len(new_messages))
+
+            await message_queue.join()
+
+    async def subscribe(subscription: str,  # pylint: disable=too-many-locals
```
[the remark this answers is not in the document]
amartsinovich-dialpad: 3. 
```
client = SubscriberClient()
manager = SubscriberManager(client)
manager.pull(...)
manager.subscribe(...)
```
vs
```
client = SubscriberClient()
client.pull()
subscribe(client)
```
same usability really, but with several ways of doing `pull`.

2. Well, if we shovel all the subscribe code into `SubscriberClient` we will kinda have this big `AND` in the list of its responsibilities. As in: `SubscriberClient` is responsible for calling pubsub subscription REST api **AND** also implements this big ass subscribe logic. But only for `aio` version, screw `rest`. 

TBH I'm having a hard time justifying any OO here. Let's see if I change my mind after closer examination of session usage.

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+            message_queue: MessageQueue,
+            subscriber_client: 'SubscriberClient',
+            max_messages: int,
+            metrics_client: Optional[MetricsAgent]
+    ) -> None:
+        while True:
+            try:
+                new_messages = await subscriber_client.pull(
+                    subscription=subscription, max_messages=max_messages)
+            except asyncio.TimeoutError:
```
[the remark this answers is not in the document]
amartsinovich-dialpad: Hmm, but what is it for? `TimeoutError` is thrown when there was no messages to pull for ~10 secs. In a worst case scenario when we have no message flow we'll be 100% ready to immediately consume messages for a price of 6 RPM * number of workers load. And we probably don't care about stressing google's api anyway?

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+                                                  subscription,
+                                                  ack_deadline_cache_timeout)
+            tasks = []
+            try:
+                tasks.append(asyncio.ensure_future(
+                    acker(subscription, ack_queue, subscriber,
+                          ack_window=ack_window, metrics_client=metrics_client)
+                ))
+                for _ in range(num_workers):
+                    q: MessageQueue = asyncio.Queue(
```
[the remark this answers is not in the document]
amartsinovich-dialpad: it'll depend on the tests results of `await message_queue.join()`

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+                    ack_queue: 'asyncio.Queue[str]',
+                    subscriber_client: 'SubscriberClient',
+                    ack_window: float,
+                    metrics_client: Optional[MetricsAgent]) -> None:
+        ack_ids: List[str] = []
+        while True:
+            time_budget = ack_window
+            if not ack_ids:
+                ack_ids.append(await ack_queue.get())
+                ack_queue.task_done()
```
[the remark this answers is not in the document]
amartsinovich-dialpad: hmm, currently it's pretty straightforward to me: we do not start timer until we get at least 1 message to ack. Once we get it, we have `time_budget` time to gather as many messages as possible until we send the batch. 
We'll need to start timer between receiving first message and  receiving the rest anyway?

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: I was thinking something like the following, which is user-driven via gcloud-aio in app rather than requiring polling from the external service:
```python
class SubscriberClient:
    # SNIP
    def get_subscription(...) -> Subscription:
        # SNIP
        return Subscriber.from_repr(result)

class Subscriber:
    def from_repr(cls, data: Dict[str, Any]) -> 'Subscriber':
        # SNIP
        return Subscriber(..., ack_deadline_seconds=['ackDeadlineSeconds']

    # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/patch
    def patch(self, ..., ack_deadline_seconds: Optional[int] = None) -> None:
        # SNIP
        if ack_deadline_seconds is not None:
            payload_body['ackDeadlineSeconds'] = ack_deadline_seconds
            self.ack_deadline_seconds = ack_deadline_seconds

async def consumer(subscription: Subscription, ...):
    while True:
        ack_deadline = subscription.ack_deadline_seconds
        # SNIP
```

The benefit there being that if users want to have their app code call `Subscription().patch(ack_deadline_seconds=42)`, it propagates to the running instance of the subscriber; but we don't need to worry about the extra logic of trying to keep our system in-sync with out-of-band changes.

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: Well isn't `ack_ids` initialized as `asyncio.Queue(maxsize=(max_messages * num_workers))`? If we try to insert `maxsize + 1` messages, the consumer attempting to do so will simply wait. Perhaps I'm misunderstanding something?

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: Executor could be reasonable, so long as we can avoid confusing it with the asyncio executor concept: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor . Should be fine if we're careful on naming.

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: I think `.cancel()` is the graceful way to do things! Or at least, I certainly haven't heard of any other options.

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
TheKevJames: Yeah, I think some of my other thoughts were the big driver of being a bit more OO here; eg. sharing the session, being able to gracefully shutdown all the threads `subscribe()` manages, the `Subscriber.patch()` thing, etc. Could also just be that I come from a bit more of an OO background than you and thus think that way without great reason!

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
TheKevJames: I wasn't quite thinking of the 10s thing -- I wonder if we should explicitly set the `timeout=10` arg to `pull()` here to make sure it's adjacent to the `except TimeoutError`... that might make it more clear when this case would occur.

I guess the case I was thinking about was mostly the permafailure thing: though your points are correct, if Google's APIs are down we might want to have a louder failure than everything running smoothly here and just-so-happening to be looping early without successfully grabbing messages. Users can solve this with alerts one their processing volume, but... well, if your subscriber is repeatedly failing to grab messages from the subscription, that feels like it should be a really loud error to me. If this happens a large amount of times in a row, we're basically in a "literally everything is non-functional" state.

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
TheKevJames: Fair enough. In my head I was envisioning something like the following, but honestly now that I write it out is doesn't seem any better. My bad!

```python
            time_budget = float('inf')
            while time_budget > 0:
                start = time.perf_counter()
                try:
                    ack_id = await asyncio.wait_for(
                        ack_queue.get(), timeout=time_budget)
                    if not ack_ids:
                        time_budget = ack_window
                    ack_ids.append(ack_id)
                    ack_queue.task_done()
                except asyncio.TimeoutError:
                    break
                time_budget -= (time.perf_counter() - start)
```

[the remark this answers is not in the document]
TheKevJames: Fair enough. In my head I was envisioning something like the following, but honestly now that I write it out is doesn't seem any better. My bad!

```python
            time_budget = float('inf')
            while time_budget > 0:
                start = time.perf_counter()
                try:
                    ack_id = await asyncio.wait_for(
                        ack_queue.get(), timeout=time_budget)
                    if not ack_ids:
                        time_budget = ack_window
                    ack_ids.append(ack_id)
                    ack_queue.task_done()
                except asyncio.TimeoutError:
                    break
                time_budget -= (time.perf_counter() - start)
```

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: That's more complexity.
* we now store a local representation of a subscription
* we introduce a wrapper over `SubscriberClient`'s method => we now have 2 ways to patch subscription
* `Subscriber` will need to be updated on any schema change from google
* we require users to treat this representation as a the only agent of change for things to work properly
* unless there is only one instance running we now have N copies of the state, each of which is read-only-at-startup. The only way to keep them in sync is to restart everything on any write.

With `ackDeadlineCache` things will just work regardless of pretty much anything

I also want to point out that we're not polling _some_ external service. We're polling the one and the only external entity the subscriber is specifically designed to interact with.

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: nonono, `ack_ids` is a simple list that is used as a temporary bucket where acker puts ids once they're taken from the queue. This bucket is emptied on successful ack request and is carried to the next iteration if not

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: maybe `execute_callback`? Would be nice to keep it a verb since all surrounding "doer" functions are workers

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: I mean, can we somehow determine we're being cancelled and perform a cleanup? Otherwise we should be 100% compliant with cancel interface :D

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+                        *,
+                        num_workers: int = 1,
+                        max_messages: int = 100,
+                        ack_window: float = 0.3,
+                        ack_deadline_cache_timeout: int = 10,
+                        consumer_pool_size: int = 1,
+                        subscriber_client: Optional[SubscriberClient] = None,
+                        metrics_client: Optional[MetricsAgent] = None
+                        ) -> None:
+        async with aiohttp.ClientSession() as session:
```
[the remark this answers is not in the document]
amartsinovich-dialpad: I revisited this piece and I think this explicit session was my workaround for an obscure `UnclosedSession` error I was getting. I removed it now and things seem to be working fine.

Now a user can pass session to `SubscriberClient` and it transparently will be used within subscriber. Does it seem right to you?

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
amartsinovich-dialpad: Hmm, let me clarify this a bit. `TimeoutError` here is a part of _normal_ flow. That's just how pull works.
Apart from that we don't catch any exception. That means that for any actual error (say, 500) the worker will throw -> the exception will surface -> app will crash and hopefully be restarted by kubernetes.

So if anything I'd say we need backoff here to not fail too loud immediately.

There's also the case when google is so down we're getting `TimeoutErrors` from it. This way we can't distinguish failures from no messages. Not sure how to go about this 🤔

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
amartsinovich-dialpad: oh, tricky! Thanks for sharing
those two lines side by side make me a little uncomfortable tho 😄 
```
time_budget = float('inf')
while time_budget > 0:
```

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: OOoh, that's the piece you were missing in `ack_ids` thread.

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: > unless there is only one instance running we now have N copies of the state, each of which is read-only-at-startup. The only way to keep them in sync is to restart everything on any write.

That's a good point; I didn't consider the multiple apps running simultaneously issue... though my base assumption was that folks would be scaling the same code horizontally in cases like that, so it would not be relevant. Bad assumption on my part!

> That's more complexity.

Yes and no -- it's more complexity in some ways, but I think it's more _useful_ complexity; eg. the complexity comes from defining the `Subscriber.patch()` API operation, which is probably something we'll want to do down the road _anyway_ as we find usecases for more of the Pubsub APIs. This solves the general class of "modify any subscription configs" rather than just the "respond to changes in ack deadline while subscribing case". It's always nice to try and avoid handling special cases specially when you can reframe the problem as it being the standard case.

That all said... fair enough, good points, I'm alright keeping this as-is.

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: Ah! Yes, the `continue` on the next line was missing in my mental model, so I thought you were clearing `ack_ids` in each case. Ah, so we do retry acking the same IDs, I see... Hmm, yeah, this is potentially worrying. Perhaps we need a max size on the `ack_ids` list so that we avoid pulling more off the `ack_queue` when it's too full? I imagine that could solve a potential bug as well: I betcha there's a max size on the `acknowledge` payload before it either rejects the request and/or starts being to slow for our timeout settings anyway. Perhaps we should never try to ack more than `n` messages in one go?

Note that relatedly, messages which are repeatedly failing to ack stay in this list forever -- do we need a max retry count?

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
TheKevJames: Ah, yeah, distinguishing the normal case from failure modes is where I got hung up. If Google returns Timeout errors when there are no message to-be-processed, your code looks 100% correct. If there's a chance of those Timeouts being due to API issues / networking problems / whatever, that's where I was thinking we needed more explicit error handling.

I'm not actually sure how to differentiate or react differently... are the payloads any different? Any thoughts from other @talkiq/engineering   folks?

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,223 @@
...
+                        max_messages: int = 100,
+                        ack_window: float = 0.3,
+                        ack_deadline_cache_timeout: int = 10,
+                        consumer_pool_size: int = 1,
+                        subscriber_client: Optional[SubscriberClient] = None,
+                        metrics_client: Optional[MetricsAgent] = None
+                        ) -> None:
+        ack_queue: 'asyncio.Queue[str]' = asyncio.Queue(
+            maxsize=(max_messages * num_workers))
+        subscriber = subscriber_client or SubscriberClient()
```
TheKevJames, line 183: Ah, this would be the cause of that unclosed session issue you might be running into -- in cases where users don't provide a `SubscriberClient`, the one you create here never explicitly calls `.close()` to properly shutdown its sessions (see the whole `__aenter__` / `__aexit__` thing).

Honestly, maybe it'd be easiest to just make `subscriber_client` a *non-optional parameter* to enforce proper usage. Thoughts?

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: Haha so in the end we'll have both complexities, classical software engineers :D

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: I actually... like this very very much. Like, don't be lazy, spend one line and pass me that client, we're doing _composition_ here.

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: I mean, your shutdown code looks good here, I think it's more a matter of the potential plumbing for shutting down the overall subscriber. Eg. you could imagine something like the following (switched to OO to make the shared variables between methods a bit clearer, but of course could be done via eg. return values)
```python
class Subscriber:
    def __init__(self):
        self._tasks = []

    async def subscribe(self):
        # your existing code, minus the shutdown stuff at the end
        self._tasks = tasks
        try:
            await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
         except CancelledError:
             pass
         except Exception:
             log.exception('Subscriber exited')
         finally:
             self.close()

    def close(self):
        for task in self._tasks:  # Even better would be to shut down more gracefully, eg. producers before consumers
            task.cancel()  # TODO: this might need an `if not task.cancelled` to avoid double-cancelling? Is double-cancelling even a problem?
```

The major changes above being that calling the `close()` method and thus stopping our subscriber can gracefully shutdown tasks in the correct order and propogate the `CancelledError` to the `await subscribe()` call the user may have -- that is no longer treated as an error, but actual errors are still reported and cause "as graceful a shutdown as possible"

pubsub/gcloud/aio/pubsub/subscriber.py
allandialpad: can you remind me again why we need to control num_producers _and_ consumers. I can understand needing to control the ratio between them but I would naively assume we wouldnt need more than one producer

also, given two nested ratios, I think you need to encompass that in the variable name. Maybe num_consumers_per_producer?

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
allandialpad: Is there seriously no way to distinguish between a network timeout and no messages in subscription -____-

if this is an concern would it be too much overhead to ping the server when a timeout happens to see if its a network timeout? or does pubsub support some sort of `echo` or `healthcheck` type thing

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,222 @@
...
+            pulled_at = time.perf_counter()
+            for m in new_messages:
+                await message_queue.put((m, pulled_at))
+
+            metrics_client.histogram(
+                'pubsub.producer.batch', len(new_messages))
+
+            await message_queue.join()
+
+    async def subscribe(subscription: str,  # pylint: disable=too-many-locals
```
allandialpad, line 170: sanity check: is there any reason not to enforce asyncio on the users of this library

Should the interface to this be non-async that returns a future, where it starts its own event loop, with the added option of being able to pass in your own event loop? Probably would also need a run_forever method

Im leaning towards _no_ as the library has asyncio in its name

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
amartsinovich-dialpad: I wonder if this is a bad thing at all. Imagine we _could_ tell that google is down? If we crash the app it will only increase our time to recovery, since google might be back any second but redeploy cycle takes minutes. Thoughts?

pubsub/tests/unit/subscriber_test.py
```diff
@@ -0,0 +1,424 @@
...
+        subscriber_client.get_subscription.assert_called_once()
+
+    @pytest.mark.asyncio
+    async def test_ack_deadline_cache_first_get_failed(subscriber_client):
+        subscriber_client.get_subscription = MagicMock(
+            side_effect=RuntimeError)
+        cache = AckDeadlineCache(
+            subscriber_client, 'fake_subscription', 10)
+        assert await cache.get() == float('inf')
+        assert cache.last_refresh
```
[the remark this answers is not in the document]
jonathan-johnston: Okay I think this is fine because I didn't yet take a look at `cache.get()`, where a refresh will be tried as long as the ack deadline keeps coming up as "not having been set" (so you can ignore my previous comment):

```python
        async def get(self) -> float:
            if self.ack_deadline == float('inf') or self.cache_outdated():
                await self.refresh()
            return self.ack_deadline
```

BTW can the `ack_deadline` ever be legitimately `float('inf')`?

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,222 @@
...
+            await callback(message)
+            await ack_queue.put(message.ack_id)
+            metrics_client.increment('pubsub.consumer.succeeded')
+            metrics_client.histogram('pubsub.consumer.latency.runtime',
+                                     time.perf_counter() - start)
+        except Exception:
+            log.exception('Application callback raised an exception')
+            metrics_client.increment('pubsub.consumer.failed')
+        finally:
+            semaphore.release()
```
jonathan-johnston, line 110: I wonder if we can make the locking and unlocking of this semaphore a responsibility of only one calling context instead of having function A lock and trust that function B unlocks. Could you set a done callback via `add_done_callback` on the `_fire_callback` future that runs `semaphore.release()`? This is a kind of "don't make other people clean up your own mess" pattern (or "clean up your own memory" pattern). On the other hand what you have is fine and in all reality is probably more performant.

pubsub/gcloud/aio/pubsub/subscriber.py
jonathan-johnston: Interesting also is whether the `acknowledge` will acknowledge a subset of the `ack_ids` or will only acknowledge all-or-nothing. If there is one ID in `ack_ids` that makes the request fail, will it continue to fail on each request until you run out of memory? How do you know which ID is "bad" and needs to be dropped?

jonathan-johnston: I like the idea to avoid trying to ack more than `n` messages in one go in that at some point you need to stop requesting more items from the `ack_queue` and "block" on the current batch to cause backpressure in the consumer and producer tasks.

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
jonathan-johnston: I mean if there are no messages to pull the response should be an empty list of messages and not a `TimeoutError` right?

https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/pull#response-body

> receivedMessages[] -> Received Pub/Sub messages. The list will be empty if there are no more messages available in the backlog. For JSON, the response can be entirely empty. The Pub/Sub system may return fewer than the maxMessages requested even if there are more messages available in the backlog.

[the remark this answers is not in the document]
jonathan-johnston: So in the case where there are no messages we go through `for m in messages` and do nothing, then `await queue.join()` and cycle right back to a new request over and over right? I'm sure that in that scenario the biggest thing we have to worry about is getting rate limited.

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: Well, naively, what if you want to consume messages at a rate faster than a single thread can pull from Google's endpoints? I assume there's a max QPS from a single producer given TCP latency + message max sizes + any API limits on Google's side, etc.

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
TheKevJames: @jonathan-johnston my understanding is that returning zero messages from that endpoint might be depreacated -- see `returnImmediately` on that page, which used to be the way to say "return immediately if there are less than `maxMessages` items available instead of waiting for some arbitrary amount of time.

...that said, I realize that value is deprecated, but it sure sounds like it might be what we want. "Otherwise, the system may wait (for a bounded amount of time) until at least one message is available, rather than returning no messages." scares me -- this line might have arbitrary latency values which we can't control? Is there any way to set that timeout value? This might be the reason why we're needing multiple producers right now, since each of them is arbitrarily waiting sometimes...

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,222 @@
...
+            ack_deadline_cache: AckDeadlineCache,
+            consumer_pool_size: int,
+            metrics_client: MetricsAgent) -> None:
+        semaphore = asyncio.Semaphore(consumer_pool_size)
+        while True:
+            await semaphore.acquire()
+
+            message, pulled_at = await message_queue.get()
+
+            ack_deadline = await ack_deadline_cache.get()
```
TheKevJames, line 125: Thinking more about the ackDeadline stuff, it occurs to me that we're only doing this check upfront rather than immediately before acking, so we'd only catch a subset of the potential failures here (eg. imagine ackDeadline=5s, the message sits on the queue for 4s [so this check passes], then takes >1s to process). I wonder if this best-effort only-a-subset-of-cases failfast stuff is worth having at all or if its just a premature optimization. Perhaps we should remove this entirely?

IIUC a good chunk of the reason to have this is for metrics reporting, eg. so users know to tune their ackDeadline value. Is there another way we can get this? What happens if we `ack` a message which has exceeded it's deadline -- do we get a response there which we could use to get the metric?

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: @TheKevJames @jonathan-johnston 
acknowledging endpoint is "all-or-nothing" in a sense that it's always "all" if you could get through network, authentication and validation:
>Acknowledging a message whose ack deadline has expired may succeed, but such a message may be redelivered later. Acknowledging a message more than once will not result in an error.

So basically if we got a response there is nothing left to do about those ids. That's why "messages failing to ack" is not even a concept here.

Google does not mention any specific limit for this endpoint. For publishing messages it is 1000 messages or 10MB, so I think it's safe to only keep 1000 messages between post attempts?

@jonathan-johnston acker backpressure is actually an interesting topic. I'd argue we don't need it _mostly_, because if a batch failed to ack, best we can do is to try again. But the priority should always go to the most recent messages, because those have higher chances to succeed in time. If we make acker a "fair" worker that wants to reliably ack everything we effectively will be increasing damage caused by random hiccups.

amartsinovich-dialpad: Alright so the limit is ~2744 messages, I've added a simple slice to cover that. Let me know what you think

jonathan-johnston: @amartsinovich-dialpad Thanks for digging on that, it's good to know that expired messages are just successful. So the only failure cases we'd see here are rate limiting, network timeouts, and server errors. This makes a lot more sense now and seems safe with the message limit.

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
amartsinovich-dialpad: @TheKevJames why this scares you tho? I mean if we trust google that request will only wait until new messages are available it should be ok. If we don't trust google... we'll retry in `session timeout` seconds anyway?
So how it works right now -> we send `pull` request and, if the queue is empty, just hang there for 10 seconds (default session timeout in `SubscriberClient`. If new messages available - cool. If not - let's try again.

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: Huh, I never thought of that. I'll gladly trust an opinion of actual `asyncio` experts here!

pubsub/tests/unit/subscriber_test.py
```diff
@@ -0,0 +1,424 @@
...
+
+    @pytest.mark.asyncio
+    async def test_ack_deadline_cache_cache_outdated(subscriber_client):
+        cache = AckDeadlineCache(
+            subscriber_client, 'fake_subscription', 1000)
+        cache.last_refresh = time.perf_counter()
+        assert not cache.cache_outdated()
+
+    @pytest.mark.asyncio
+    async def test_ack_deadline_cache_cache_outdated_false(subscriber_client):
```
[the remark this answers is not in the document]
amartsinovich-dialpad: huh, yes. I'm also not sure what `test_ack_deadline_cache_cache_outdated_true` is testing 🤦

pubsub/tests/unit/subscriber_test.py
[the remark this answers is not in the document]
amartsinovich-dialpad: you're right, we refresh cache every time to limit our potential network output. There is actually a bug here. For the very first fetch we will retry every time thanks to `if self.ack_deadline == float('inf') or self.cache_outdated()` condition. I had a pleasure to see an impact today on staging -> application very quickly becomes IO bound.

>BTW can the ack_deadline ever be legitimately float('inf')?

nah, max value is 600s

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: Hm, that makes sense to me. Something like this?
```
task = asyncio.ensure_future(_execute_callback(
                message,
                callback,
                ack_queue,
                metrics_client,
            ))
task.add_done_callback(lambda x: semaphore.release())
```

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: What you saying is true, failfast is by no means exhaustive. It will only make sure we don't run potentially heavy processing on lost cause messages and show it in metrics.

>do we get a response there which we could use to get the metric?

"Acknowledging a message whose ack deadline has expired may succeed, but such a message may be redelivered later."
I'm not 100% sure what they mean by _may succeed_, but in practice it always succeeds.

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: > However this doesn't seem to solve the case of unbounded list growth, is that right?

I might have pushed the updated version a bit after you checked the code. Check reassignment at line 83, it should solve the unbound list. Maximum one-time list size should be 2500 messages from last iteration + messages we gathered during one batch window.

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: Yeah, I'd be content requiring users to only pass in `async` functions here

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: > What you saying is true, failfast is by no means exhaustive. It will only make sure we don't run potentially heavy processing on lost cause messages and show it in metrics.

I think I'd be ok with that, assuming we can get the failfast metric elsewhere -- if a user has set their ackDeadline to `n` seconds and it takes `>n` seconds to process them, they'll have to tweak settings anyway. If we can make our system simpler and easier to reason about in exchange for that case also burning CPU cycles, that seems like a fair tradeoff to me (note that we're not actually wasting processing time in that case anyway, we're just replacing noop loops with calls to their worker; they'd have to be running the app already anyway so nothing is really "wasted" here)

> I'm not 100% sure what they mean by may succeed, but in practice it always succeeds.

Well that's garbage. Maybe this should be moved to the acker thread then? It's worth comparing against the ackdeadline to get that failfast metric to help users debug, but it would be even more valuable if they could rely on it as a 100% source of truth.

pubsub/gcloud/aio/pubsub/subscriber.py
allandialpad: Would the right pattern be to just run more threads from the user-app's context? 

more of a curiosity, I wasnt around when you guys designed interface

pubsub/gcloud/aio/pubsub/subscriber.py
jonathan-johnston: Ah, gotcha, thanks!

pubsub/gcloud/aio/pubsub/subscriber.py
jonathan-johnston: The typical way to ensure a "graceful" shutdown is via `asyncio.Event` which is checked each loop in the tasks instead of `while True:`. That said, you should be careful with this method because it doesn't stop blocking actions (ex. `await` on a blocking queue) from making your program hang instead of shut down.

pubsub/gcloud/aio/pubsub/subscriber.py
shaundialpad: I like it. I think :P So with no back pressure, worse case scenario is messages will just be organically redelivered. 

Should we add a metric for failed ack requests?

amartsinovich-dialpad: I don't see why not

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: Huh, yeah I see why graceful shutdown might be a thing of the future

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: Not all CPU cycles are equal, we'll be running blocking single-threaded models in that callback, won't we? Every second of processing there adds +1 to each pending message timer. Not to say that 1 second could be spent on a message that could actually be acked. So they're "wasted" in economics sense :D
Moving this metric to acker will take away all the value from that `if` in exchange for a metric that is not 100% accurate, because there is a time gap between acker reporting metric and making a request.

By the way, google now exposes `subscription/ack_latencies` metric so we already have a metric for the whole chain.

pubsub/gcloud/aio/pubsub/subscriber.py
shaundialpad: We use this for egress and it hasn't been an issue (yet). Also I recall in google's pubsub source code they relied on time.time() as well. Sometimes we do get small negative values in the receive latency 😄

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,223 @@
...
+            task.add_done_callback(lambda _f: semaphore.release())
+            message_queue.task_done()
+
+    async def producer(
+            subscription: str,
+            message_queue: MessageQueue,
+            subscriber_client: 'SubscriberClient',
+            max_messages: int,
+            metrics_client: MetricsAgent
+    ) -> None:
```
shaundialpad, line 154: nit: `) -> None:` in line with `async def` is not consistent with the rest of the declarations

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+
+            pulled_at = time.perf_counter()
+            for m in new_messages:
+                await message_queue.put((m, pulled_at))
+
+            if metrics_client:
+                metrics_client.histogram(
+                    'pubsub.producer.batch', len(new_messages))
+
+            await message_queue.join()
```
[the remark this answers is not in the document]
amartsinovich-dialpad: Intermediate results. Tested with `egress`, i.e. very fast consumer.
In high load scenarios there's no difference in performance or CPU usage.
Under small load not having `message_queue.join()`  generate occasional spikes in receive latency:
image
Not sure why, tho.

I'll be testing against ctm next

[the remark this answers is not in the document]
amartsinovich-dialpad: Hmm, `ctm` looks alright. Now when I think about it, it looks like `egress` profile _is_ the worst case scenario for this 🤔  When `callback` is fast and asynchronous, there are plenty of time slots for consumer to fetch too many messages. But this seems to be not the case for the current design.  I'll go ahead and implement single message queue variant and test it

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,229 @@
...
+            tasks = []
+            try:
+                tasks.append(asyncio.ensure_future(
+                    acker(subscription, ack_queue, subscriber,
+                          ack_window=ack_window, metrics_client=metrics_client)
+                ))
+                for _ in range(num_workers):
+                    q: MessageQueue = asyncio.Queue(
+                        maxsize=max_messages)
+                    tasks.append(asyncio.ensure_future(
```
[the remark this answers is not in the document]
amartsinovich-dialpad: OK so tests look promising, but I'd really like to discuss switching from consumer task pool to just number of consumers. Currently in my tests I use 4 consumer workers each with pool of 1000 tasks. That would translate to 4000 consumer workers. It works alright, performance is the same, but it takes ~20MB more memory than current subscriber. I'm also not sure this is a good idea to statically define large number of workers that all will be occupying scheduler time no matter what load we have. Maybe it'd be more rational to go another way around: a single consumer with arbitrary large semaphore pool?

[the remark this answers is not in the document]
amartsinovich-dialpad: Oh wow semaphore version isn't great at all and also doesn't really help with memory consumption 🤔

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
amartsinovich-dialpad: Alright, bottomline:
0. current solution
1. no `message_queue.join()` -> same performance
2. no `message_queue.join()` + single queue + arbitrary number of consumer workers -> same performance, higher memory consumption
3. no `message_queue.join()` + single queue + single consumer with arbitrary large semaphore-base tasks pool -> poor performance, higher memory consumption

We could go with either 0 or 1.  Personally I lean toward going with current solution, since it's exactly what was tested throughout the whole process and it does not maintain an implicit pool of messages in memory.

Please let me know what you think. With this decision the PR should be unblocked for final polishing.

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: Yeah -- these arguments are to give users exactly that option: the `subscribe()` method is basically a convenience function for setting up the various threads for producing/consuming/acking/etc, so the `consumer_pool_size` is the way that a user-app context could determine how many threads to create via `subscribe()`

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
TheKevJames: It's mostly the "trust Google" part that worries me haha. We're relying here on their idea of "bounded amount of time" being reasonable and matching our assumptions. See for example how aiohttp's default idea of a "reasonable default timeout" was something like ten minutes, which makes absolutely no sense in a lot of the places we use it.

What happens if Google decides "reasonable" is "24 hours"? What if the http connection breaks in a way what our client can't detect, so we sit waiting for them to timeout the connection but never get that notification? Could this lead to zombie threads sitting around doing nothing?

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
TheKevJames: Awesome results, thanks! Option 0 seems reasonable to me, but I'm glad we tested the other options just in case. The added memory consumption seems a bit odd to me, so I might play around on personal time to figure out what's up here, but we can always make changes as a future optimization rather than blocking this PR

pubsub/gcloud/aio/pubsub/subscriber.py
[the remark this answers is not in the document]
amartsinovich-dialpad: Our default timeout is `10s` and totally overridable by user. We never hang on a line for more than 10 seconds by default.

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: > By the way, google now exposes `subscription/ack_latencies` metric so we already have a metric for the whole chain.

Oh, fantastic, that renders this whole conversation moot then -- happy to rely on their metrics for this rather than trying to coerce our own system into giving us a copy of that data. Sounds good :+1:

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,223 @@
...
+
+    async def subscribe(subscription: str,  # pylint: disable=too-many-locals
+                        handler: ApplicationHandler,
+                        subscriber_client: SubscriberClient,
+                        *,
+                        num_workers: int = 1,
+                        max_messages: int = 100,
+                        ack_window: float = 0.3,
+                        ack_deadline_cache_timeout: int = 10,
+                        consumer_pool_size: int = 1,
```
TheKevJames, line 179: This might be better named `num_consumers_per_worker` to line it up with the terminology used in `num_workers` and proposed `max_messages_per_worker`.

Relatedly, `worker` is a bit broad with all the various types of workers we have here (eg. it might not be obvious that worker here means "pair of producers and n-thread0consumers". Perhaps having `num_producers`, `max_messages_per_producer`, and `num_consumers_per_producer` would be nice and clear?

TheKevJames (changes requested):

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: ok, but let me test it first. currently egress uses a pool of  `4000` tasks combined, that might surprise us a bit

amartsinovich-dialpad: alright, the tests didn't go very well :( I guess we'll stick to current architecture for now

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: alright so now when the design is settled time to decide on this. We can go with "imaginary composite workers" and name parameters `num_workers`, `max_messages_per_worker` and `worker_pool_size`. Or we may not try to hide implementation complexity from users and call them simply "num_producers", "max_messages_per_producer", "max_tasks_per_consumer"? I'm leaning towards the latter one, since we'll probably have to expose implementation in the docs anyway. And of course feel free to nominate your suggestions :)

/cc @TheKevJames @shaundialpad @jonathan-johnston @allandialpad

shaundialpad: I like hiding implementation details, but I like clarity even more I think. Without any context, the former option would be very confusing to me and I'd have to dive into the code to see what each setting is referring to, particularly the difference between `num_workers` and `worker_pool_size`. That being said, even with the latter option, I would not know exactly how I should go about tuning these anyhow. I suppose we're going to have some sane defaults and then elaborate on how one might do this in the documentation anyway.

So yeah I vote for `num_producers`, `max_messages_per_producer` and `num_consumers_per_producer`

TheKevJames: Agreed with option 2 :+1: `num_tasks_per_consumer` makes sense for now, and we can change it to `num_consumers_per_producer` if we ever figure out what's up with the perf on that and switch over.

pubsub/README.rst
```diff
@@ -41,6 +39,88 @@ Here's the rough usage pattern for using ``SubscriberClient``:
...
+        handler,
+        subscriber_client,
+        num_workers=1,
+        max_messages=100,
+        ack_window=0.3,
+        consumer_pool_size=1,
+        metrics_client=MetricsAgent()
+    )
+
+While defaults are somewhat sensinble, it is highly recommended to performance test
```
TheKevJames, line 68: nit: `s/sensinble/sensible/`

pubsub/README.rst
```diff
@@ -41,6 +39,88 @@ Here's the rough usage pattern for using ``SubscriberClient``:
...
+your application and tweak function parameter to your specific needs. Here's a few hints:
+
+:``handler``:
+    an async function that will be called for each message. It should accept an instance of
+    ``SubscriberMessage`` as its only argument and return ``None`` if the message should
+    be acked.
```
TheKevJames, line 74: Add something like "The message will be nacked if the handler raises an exception"?

pubsub/gcloud/aio/pubsub/__init__.py
```diff
@@ -1,11 +1,17 @@
...
 __version__ = get_distribution('gcloud-aio-pubsub').version
 
+from gcloud.aio.auth import BUILD_GCLOUD_REST
 from gcloud.aio.pubsub.publisher_client import PublisherClient
 from gcloud.aio.pubsub.subscriber_client import SubscriberClient
 from gcloud.aio.pubsub.utils import PubsubMessage
 from gcloud.aio.pubsub.subscriber_message import SubscriberMessage
 
+if BUILD_GCLOUD_REST:
+    __all__ = ['__version__', 'PublisherClient', 'PubsubMessage',
```
TheKevJames, line 11: nit: rather than duplicating this list, define it once at the top-level and do:
```python
if not BUILD_GCLOUD_REST:
    from gcloud.aio.pubsub.subscriber import subscribe
    __all__.append('subscribe')
```

pubsub/README.rst
amartsinovich-dialpad: psst, Kevin, it won't ba acked in the current version, it will be left to expire...

pubsub/README.rst
```diff
@@ -90,12 +90,12 @@ your application and tweak function parameter to your specific needs. Here's a f
     specifies how often ack requests will be made. Setting it to ``0.0`` will effectively
     disable batching.
 
-:``consumer_pool_size``:
+:``num_tasks_per_consumer``:
     how many ``handle`` calls a worker can make until it blocks
     to wait for them to return. If you process messages independetly from each other you should
```
allandialpad, line 95: independently

allandialpad (approved):

pubsub/README.rst
```diff
@@ -41,6 +39,88 @@ Here's the rough usage pattern for using ``SubscriberClient``:
...
+
+:``num_producers``:
+    number of workers that will be making ``pull`` requests to pubsub.
+    Please note that a worker will only fetch new batch once the ``handler``
+    was called for each message from the previous batch. This means that running only a single worker
+    will most likely make your application IO bound. If you notice this being an issue don't hesitate
+    to bump this parameter.
+
+:``max_messages_per_producer``:
+    number of pubsub messages a worker will try to fetch in a single batch. This
```
shaundialpad, line 84: The line width throughout this readme is inconsistent. Should it be 80 chars throughout?

pubsub/README.rst
shaundialpad: I think emphasizing this is also important. i.e. adding something like "An exception raised within the handler will result in the message being left to expire, and thus it will be redelivered according to your subscription's ack deadline"

amartsinovich-dialpad: makes a lot of sense, thanks!

pubsub/README.rst
amartsinovich-dialpad: Fair enough, time to get a wrap plugin

pubsub/tests/unit/subscriber_test.py
```diff
@@ -0,0 +1,469 @@
...
+        message, pulled_at = await asyncio.wait_for(queue.get(), 0.1)
+        producer_task.cancel()
+        assert message
+        assert isinstance(pulled_at, float)
+
+    @pytest.mark.asyncio
+    async def test_producer_timeout_error_is_ok(subscriber_client):
+        async def f(*_args, **_kwargs):
+            await asyncio.sleep(0)
+            raise RuntimeError
```
shaundialpad, line 190: I'm a bit confused with this test. It sounds like we are trying to test that if `subscriber_client.pull` raises `TimeoutError` then we don't raise the exception. But in the test we mock it to raise `RuntimeError` which would have a different codepath than the `TimeoutError` case.

In fact, I was curious and just ran the test locally, and it _seems_ that somehow this raises `CancelledError`. Could you confirm this test is doing what it is supposed to be doing? 😄

pubsub/tests/unit/subscriber_test.py
```diff
@@ -0,0 +1,469 @@
...
+        acker_task.cancel()
+        subscriber_client.acknowledge.assert_called_once_with(
+            'fake_subscription', ack_ids=['ack_id'])
+        assert queue.qsize() == 0
+
+    @pytest.mark.asyncio
+    async def test_acker_handles_exception(subscriber_client):
+        async def f(*_args, **_kwargs):
+            await asyncio.sleep(0)
+            raise RuntimeError
```
shaundialpad, line 376: Similar to my above comment - I don't think we get cancelled before raising `RuntimeError`

pubsub/tests/unit/subscriber_test.py
amartsinovich-dialpad: You absolutely right. There was also mock object missing so we didn't really know if anything was called at all. Please check now!

pubsub/tests/unit/subscriber_test.py
shaundialpad: Looks good now!

pubsub/README.rst
```diff
@@ -41,6 +39,96 @@ Here's the rough usage pattern for using ``SubscriberClient``:
...
+    )
+
+While defaults are somewhat sensible, it is highly recommended to performance
+test your application and tweak function parameter to your specific needs.
+Here's a few hints:
+
+:``handler``:
+    an async function that will be called for each message. It should accept an
+    instance of ``SubscriberMessage`` as its only argument and return ``None``
+    if the message should be acked. An exception raised within the handler will
```
shaundialpad, line 75: A thought..

Is it worth calling out how one should go about explicitly `nack`ing a message? i.e. via modify_ack_deadline = 0. Take `egress` for example. `egress` has a pretty lengthy ack deadline of 20 seconds because its downstream service can have some long running requests. On top of this, `egress` also processes messages very quickly (aside from the time it takes to POST them), and also retries a lot of messages. I suspect we may want to explicitly nack messages in the case of failures in `egress`' case because of this to reduce potential latency as a result of retrying.

This may have been mentioned in the pr already but there are so many comments now 😄

amartsinovich-dialpad: I think explicit nacking just naturally creeped out of scope for this one. That said, now when I'm testing egress it looks like we really need this in version 1.

amartsinovich-dialpad: Created this as a separate PR to not clutter this one. Please do take a look! https://github.com/talkiq/gcloud-aio/pull/270

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -0,0 +1,226 @@
...
+            # which is ~2744 ack_ids
+            if len(ack_ids) > 2500:
+                log.error(
+                    'acker is falling behind, dropping %d unacked messages',
+                    len(ack_ids) - 2500)
+                ack_ids = ack_ids[-2500:]
+            try:
+                await subscriber_client.acknowledge(subscription,
+                                                    ack_ids=ack_ids)
+            except Exception as e:
```
shaundialpad, line 91: If `acker` gets cancelled, we probably want to raise `CancelledError`. Since we are catching the general `Exception` here this won't happen I believe and `acker` will continue looping. Should we add a separate `except CancelledError:` here and raise it?

amartsinovich-dialpad: Right, I didn't realize `CancelledError` can be thrown at any point

pubsub/README.rst
```diff
@@ -41,6 +39,96 @@ Here's the rough usage pattern for using ``SubscriberClient``:
...
+        num_producers=1,
+        max_messages_per_producer=100,
+        ack_window=0.3,
+        num_tasks_per_consumer=1,
+        metrics_client=MetricsAgent()
+    )
+
+While defaults are somewhat sensible, it is highly recommended to performance
+test your application and tweak function parameter to your specific needs.
+Here's a few hints:
```
jonathan-johnston, line 70: I love this list of hints and how you linked parameter value to subscriber behaviour so the user knows which knob to turn when they see those behaviours.

jonathan-johnston (approved):
🚢

shaundialpad (approved):
Alright I think that's enough commentary 😄 ✅

TheKevJames (approved):

PR #270, opened 2020-12-22
feat(pubsub): Optional nacker worker in subscriber

amartsinovich-dialpad:
This is mostly copypaste of acker ~~but it's our copypaste~~ but there's not much room I can see for deduplication without introducing crazy worker factories.


pubsub/README.rst
```diff
@@ -102,6 +104,14 @@ Here's a few hints:
     limit of how many messages can possibly be within your application state at
     any given moment.
 
+:``enable_nack``:
+    if enabled messages for which ``callback`` raised an exception will be
+    explicitely nacked using ``modifyAckDeadline`` endpoint so they can be
+    retried immediately.
+
+:``nack_windows``:
```
shaundialpad, line 112: s/nack_windows/nack_window

pubsub/README.rst
```diff
@@ -102,6 +104,14 @@ Here's a few hints:
     limit of how many messages can possibly be within your application state at
     any given moment.
 
+:``enable_nack``:
+    if enabled messages for which ``callback`` raised an exception will be
+    explicitely nacked using ``modifyAckDeadline`` endpoint so they can be
```
shaundialpad, line 109: s/explicitely/explicitly

shaundialpad (commented):
I think it looks good! Unfortunate we cannot reduce the duplication between acker and nacker. Interested in others thoughts on this as well.

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -98,9 +104,45 @@ async def acker(subscription: str,
...
+                log.error(
+                    'nacker is falling behind, dropping %d unacked messages',
+                    len(ack_ids) - 2500)
+                ack_ids = ack_ids[-2500:]
+            try:
+                await subscriber_client.modify_ack_deadline(
+                    subscription,
+                    ack_ids=ack_ids,
+                    ack_deadline_seconds=0)
+            except Exception as e:
```
shaundialpad, line 132: https://github.com/talkiq/gcloud-aio/pull/265#discussion_r552023087 applies here as well

shaundialpad (approved):
Approved pending the one comment

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -55,28 +56,33 @@ def cache_outdated(self) -> bool:
                 return True
             return False
 
+    async def _budgeted_queue_get(queue: 'asyncio.Queue[Any]',
```
TheKevJames, line 59: You can be a bit more specific with the type hints here: if you use a [TypeVar](https://docs.python.org/3/library/typing.html#typing.TypeVar) then you can say:
```
async def _budgeted_queue_get(queue: 'asyncio.Queue[T]', ..) -> List[T]
```
which in turn means you won't lose the response type value when calling this method.

TheKevJames (approved):
Other than the type hint, looks great!

jonathan-johnston (approved):

PR #273, opened 2021-01-04
feat(gcloud-aio): narrow pin for aiohttp + requirement chardet

eddiedialpad:
Okay, so a few apps in our stack are emitting: `aiohttp 3.7.3 has requirement chardet<4.0,>=2.0, but you have chardet 4.0.0.`

We believe that this is all tracked down to `aiohttp` included in this repo -- version `3.0.4` is the final version of chardet in the 3.x.x major version.

Note that this pin would need manual removal if a later version of aiohttp requires chardet >= 4.0.0; I'm not 100% sure this is the best way to do this, but I figured we can talk about it in this PR.

### DIFF
--- a/auth/requirements.txt
+++ b/auth/requirements.txt
@@ -1,4 +1,4 @@
-aiohttp>=3.7.3,<4.0.0
+aiohttp>=3.3.0,<4.0.0
 backoff>=1.0.0,<2.0.0
 chardet>=2.0,<4.0  # required by aiohttp 3.7.3
 cryptography>=2.0.0,<4.0.0
--- a/auth/requirements.txt
+++ b/auth/requirements.txt
@@ -1,5 +1,6 @@
-aiohttp>=3.3.0,<4.0.0
+aiohttp>=3.7.3,<4.0.0
 backoff>=1.0.0,<2.0.0
+chardet>=2.0,<4.0  # required by aiohttp 3.7.3
 cryptography>=2.0.0,<4.0.0
 future>=0.17.0,<0.18.3
 pyjwt>=1.5.3,<2.0.0


harsh-dialpad (approved):
LGTM as long as all the tests pass

eddiedialpad:
@harsh-dialpad -- I changed the pin back to an interval -- chardet>=2.0,<4.0 -- this is because I realized downstream users might want to intentionally pick a specific chardet version, so a range means we can maintain compatibility for them.

Does this change seem ok to you too?

(Previously: was fixed to 3.0.4)

auth/requirements.txt
```diff
@@ -1,5 +1,6 @@
-aiohttp>=3.3.0,<4.0.0
+aiohttp>=3.7.3,<4.0.0
 backoff>=1.0.0,<2.0.0
+chardet>=2.0,<4.0  # required by aiohttp 3.7.3
```
harsh-dialpad, line 3: Seems fine. Just curious, do we know of clients who would want `chardet==2.x`? Do we know aiohttp 3.7.3 works okay with chardet 2.x?

eddiedialpad: I don't know what our users want TBH -- but, chardet>=2.0 lower bound is actually copied over from aiohttp 3.7.3 's requirements.

harsh-dialpad: Okay then that's fine I guess. Thanks for confirming!

harsh-dialpad (approved):

auth/requirements.txt
```diff
@@ -1,5 +1,6 @@
-aiohttp>=3.3.0,<4.0.0
+aiohttp>=3.7.3,<4.0.0
```
TheKevJames, line 1: Do we have to bump the aiohttp minimum pin? I'd really rather avoid artificially limiting our users if we can help it.

TheKevJames (changes requested):

TheKevJames (approved):

eddiedialpad:
Okay -- there's an extra job github is waiting on -- `ci/circleci: lint-rest` that isn't part of the circle yaml.

Anyway -- @TheKevJames if the current solution is satisfactory, please merge this PR and release it. Yay!

PR #275, opened 2021-01-06
chore(pubsub): bump requirement for gcloud-aio-auth

eddiedialpad:
Now that gcloud-aio-auth==3.4.5 has the correct version pin for chardet, let's have pubsub pick it up so our downstream apps can import pubsub without further requirements conflicts.

### DIFF
--- a/pubsub/requirements.txt
+++ b/pubsub/requirements.txt
@@ -1,1 +1,1 @@
-gcloud-aio-auth >= 3.3.0, < 4.0.0
+gcloud-aio-auth >= 3.4.5, < 4.0.0


TheKevJames (changes requested):
Er, isn't 3.4.5 already between 3.3.0 and 4.0.0? gcloud-aio-pubsub doesn't rely on any features which were only introduced in auth==3.4.5, there's no reason to bump the lower bound. auth==3.3.0 still works just fine so long as the application that is eventually installing it resolves dependencies reasonably.

eddiedialpad:
Oh right, I forgot that by default, the latest version inside a range is picked first.

Thanks Kevin!

PR #277, opened 2021-01-08
feat(pubsub) Gracefully shutdown subscriber

amartsinovich-dialpad:
An attempt to gracefully shutdown subscriber. First we shutdown producers, making sure they pushed everything they already had downstream _and wait until consumers consume everything from the message queue_ (this makes consumer logic a lot easier). Then we shut down consumer. By this time we know that message queue is empty, so all we need to do is to exhaust semaphore and wait until ack/nack queues are empty. After that we're good to go.

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -216,18 +216,20 @@ else:
             metrics_client: MetricsAgent) -> None:
         try:
             while True:
+                new_messages = []
                 try:
                     new_messages = await subscriber_client.pull(
                         subscription=subscription, max_messages=max_messages)
                 except (asyncio.TimeoutError, KeyError):
                     continue
 
-                pulled_at = time.perf_counter()
-                for m in new_messages:
-                    await message_queue.put((m, pulled_at))
-
                 metrics_client.histogram(
                     'pubsub.producer.batch', len(new_messages))
+
+                pulled_at = time.perf_counter()
+                while new_messages:
+                    await message_queue.put((new_messages[-1], pulled_at))
+                    new_messages.pop()
 
                 await message_queue.join()
         except asyncio.CancelledError:
@@ -260,18 +262,18 @@ else:
                                               subscription,
                                               ack_deadline_cache_timeout)
         metrics_client = metrics_client or MetricsAgent()
-        tasks = []
+        acker_tasks = []
         consumer_tasks = []
         producer_tasks = []
         try:
-            tasks.append(asyncio.ensure_future(
+            acker_tasks.append(asyncio.ensure_future(
                 acker(subscription, ack_queue, subscriber_client,
                       ack_window=ack_window, metrics_client=metrics_client)
             ))
             if enable_nack:
                 nack_queue = asyncio.Queue(
                     maxsize=(max_messages_per_producer * num_producers))
-                tasks.append(asyncio.ensure_future(
+                acker_tasks.append(asyncio.ensure_future(
                     nacker(subscription, nack_queue, subscriber_client,
                            nack_window=nack_window,
                            metrics_client=metrics_client)
@@ -296,7 +298,7 @@ else:
                              metrics_client=metrics_client)
                 ))
 
-            all_tasks = [*producer_tasks, *consumer_tasks, *tasks]
+            all_tasks = [*producer_tasks, *consumer_tasks, *acker_tasks]
             done, _ = await asyncio.wait(all_tasks,
                                          return_when=asyncio.FIRST_COMPLETED)
             for task in done:
@@ -314,7 +316,7 @@ else:
             await asyncio.wait(consumer_tasks,
                                return_when=asyncio.ALL_COMPLETED)
 
-            for task in tasks:
+            for task in acker_tasks:
                 task.cancel()
-            await asyncio.wait(tasks, return_when=asyncio.ALL_COMPLETED)
+            await asyncio.wait(acker_tasks, return_when=asyncio.ALL_COMPLETED)
         raise asyncio.CancelledError('Subscriber shut down')
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -169,34 +169,44 @@ else:
             max_tasks: int,
             nack_queue: 'Optional[asyncio.Queue[str]]',
             metrics_client: MetricsAgent) -> None:
-        semaphore = asyncio.Semaphore(max_tasks)
-        while True:
-            await semaphore.acquire()
+        try:
+            semaphore = asyncio.Semaphore(max_tasks)
+            while True:
+                await semaphore.acquire()
 
-            message, pulled_at = await message_queue.get()
+                message, pulled_at = await message_queue.get()
 
-            ack_deadline = await ack_deadline_cache.get()
-            if (time.perf_counter() - pulled_at) >= ack_deadline:
-                metrics_client.increment('pubsub.consumer.failfast')
+                ack_deadline = await ack_deadline_cache.get()
+                if (time.perf_counter() - pulled_at) >= ack_deadline:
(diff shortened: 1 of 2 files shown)


pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -169,56 +169,76 @@ async def consumer(
...
+                    ack_queue,
+                    nack_queue,
+                    metrics_client,
+                ))
+                task.add_done_callback(lambda _f: semaphore.release())
+                message_queue.task_done()
+        except asyncio.CancelledError:
+            log.info('Consumer worker cancelled. Gracefully terminating...')
+            for _ in range(max_tasks):
+                await semaphore.acquire()
```
shaundialpad, line 204: So this relies on the producers having stopped pulling messages and cancelling the consumers themselves, correct?

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -169,56 +169,76 @@ async def consumer(
...
+
+                pulled_at = time.perf_counter()
+                for m in new_messages:
+                    await message_queue.put((m, pulled_at))
 
+                metrics_client.histogram(
+                    'pubsub.producer.batch', len(new_messages))
+
+                await message_queue.join()
+        except asyncio.CancelledError:
```
shaundialpad, line 233: The behavior here looks like it will depend on when and where we get cancelled in lines 217-233 above. If we get cancelled before we even make one successful pull, I don't think `new_messages` will be defined and will result in a `NameError`. This will be caught and handled elsewhere, though.

If we get cancelled on https://github.com/talkiq/gcloud-aio/blob/am/pubsub/graceful_shutdown/pubsub/gcloud/aio/pubsub/subscriber.py#L220 after at least one pull, then I think the `new_messages` we end up putting in the queue on https://github.com/talkiq/gcloud-aio/blob/am/pubsub/graceful_shutdown/pubsub/gcloud/aio/pubsub/subscriber.py#L237 will be those from the previous iteration (so processing them twice)

Same goes for elsewhere in the above code block.. some messages could be processed twice. Any issues with this?

Perhaps we can initialize each iteration with `new_messages=[]`, and then change the for loop to something like 

```python
                while new_messages:
                    await message_queue.put((new_messages[-1], pulled_at))
                    new_messages.pop()
```

(I wanted to pop() on the same line as the `await` but that could result in a dropped message)

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -241,6 +261,8 @@ async def subscribe(subscription: str,  # pylint: disable=too-many-locals
                                               ack_deadline_cache_timeout)
         metrics_client = metrics_client or MetricsAgent()
         tasks = []
```
shaundialpad, line 263: Maybe `tasks` should be renamed to `acker_tasks` or something now since we have a separate vars for `consumer_tasks` and `producer_tasks`?

shaundialpad, line 263: Maybe tasks should be renamed to acker_tasks or something now since we have a separate vars for consumer_tasks and producer_tasks?

pubsub/tests/unit/subscriber_test.py
```diff
@@ -269,6 +269,38 @@ async def f(*args, **kwargs):
         assert producer_task.done()
         assert producer_task.exception()
 
+    @pytest.mark.asyncio
+    async def test_producer_gracefully_shutdowns(subscriber_client):
```
shaundialpad, line 273: shutdowns -> shutsdown? 😄

pubsub/tests/unit/subscriber_test.py
```diff
@@ -269,6 +269,38 @@ async def f(*args, **kwargs):
...
+        mock = MagicMock()
+        def g():
```
shaundialpad, line 275: Interesting 😄 I just want to point out that I think setting up `g` here as a generator doesn't quite represent the actual behavior we would see. For example, the `UnboundLocalError` case I mentioned above doesn't occur when it maybe should. 

You could also do something like

```python
async def f():
    raise asyncio.CancelledError
    
subscriber_clent.pull = f
```

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: yes, except that consumers are stopped by top level `subscribe` after producers reported their proper shutdown

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: oh yes, very good point! 🙏

pubsub/tests/unit/subscriber_test.py
amartsinovich-dialpad: haha, broken mental regex :D

pubsub/tests/unit/subscriber_test.py
amartsinovich-dialpad: Oh wait this test is outright incorrect. The first message is already published by the time generator throws an exception 🤔

amartsinovich-dialpad: Should be better now!

pubsub/tests/unit/subscriber_test.py
```diff
@@ -270,36 +271,34 @@ async def f(*args, **kwargs):
         assert producer_task.exception()
 
     @pytest.mark.asyncio
-    async def test_producer_gracefully_shutdowns(subscriber_client):
-        mock = MagicMock()
-        def g():
-            yield 1
-            mock()
+    async def test_producer_gracefully_shutsdown(subscriber_client):
+        def f():
```
shaundialpad, line 275: Do we expect `f` to be called twice here? Just wondering as this first raises the CancelledError, and _then_ returns an integer. But it looks like reversing this (returning integer, _then_ raising) also results in the test passing.

Also maybe using `Mock(side_effect=(1, asyncio.CancelledError))` is clearer than adding the function attribute?

amartsinovich-dialpad: Yeah, first we need to raise to simulate being cancelled and then it should be called normally to not raise during graceful shutdown.

Ha, I didn't know `side_effect` work like that, awesome

pubsub/tests/unit/subscriber_test.py
```diff
@@ -282,11 +307,11 @@ async def test_producer_fetches_once_then_blocks(subscriber_client):
             )
         )
         await asyncio.sleep(0)
-        await asyncio.wait_for(queue.get(), 1)
+        await asyncio.wait_for(queue.get(), 1.0)
+        producer_task.cancel()
```
shaundialpad, line 311: What is this test trying to accomplish? It's not immediately clear to me. I think I'm not understanding the "then_blocks" part.

pubsub/tests/unit/subscriber_test.py
```diff
@@ -452,6 +477,46 @@ async def f(*args):
...
+        mock.assert_called_once()
+        consumer_task.cancel()
+        await asyncio.sleep(0)
+        assert not consumer_task.done()
+        event.set()
+        await asyncio.sleep(0)
+        assert ack_queue.qsize() == 1
+        await ack_queue.get()
+        ack_queue.task_done()
+        await asyncio.sleep(0.3)
```
shaundialpad, line 517: Maybe a comment of why we need 0.3 here?

amartsinovich-dialpad: Even better, we don't need `0.3` here anymore!

pubsub/tests/unit/subscriber_test.py
amartsinovich-dialpad: It tests then consumer won't fetch new batch if consumer has not yet exhausted the queue. Probably worth renaming to something like `test_producer_fetches_once_then_waits_for_consumer`

shaundialpad (approved):
LGTM!

TheKevJames (approved):

PR #280, opened 2021-01-13
fix(pubsub) Fix a bug in consumer's graceful shutdown

amartsinovich-dialpad:
Well, this is a tricky bug. During graceful shutdown in consumer we can be sure that previous tasks were correctly scheduled, but not the last one. The last one is a total uncertainty: it could be  that it wasn't scheduled, or we could fail to attach a callback to it.

This is the best solution I could come up with: properly wait for previous tasks but just a little bit for the very last lock.

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -171,17 +171,17 @@ else:
             metrics_client: MetricsAgent) -> None:
         try:
             semaphore = asyncio.Semaphore(max_tasks)
-            while True:
-                await semaphore.acquire()
 
-                message, pulled_at = await message_queue.get()
+            async def _consume_one(message: SubscriberMessage,
+                                   pulled_at: float) -> None:
+                await semaphore.acquire()
 
                 ack_deadline = await ack_deadline_cache.get()
                 if (time.perf_counter() - pulled_at) >= ack_deadline:
                     metrics_client.increment('pubsub.consumer.failfast')
                     message_queue.task_done()
                     semaphore.release()
-                    continue
+                    return
 
                 metrics_client.histogram(
                     'pubsub.consumer.latency.receive',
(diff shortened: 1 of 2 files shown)


pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -200,8 +200,23 @@ async def consumer(
                 message_queue.task_done()
         except asyncio.CancelledError:
             log.info('Consumer worker cancelled. Gracefully terminating...')
-            for _ in range(max_tasks):
-                await semaphore.acquire()
+            # we can't know if last task was even scheduled,
```
TheKevJames, line 203: I wonder if we could solve this using the `locked()` method ([source](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Semaphore.locked)). ie. something like "acquire the first `n-1` locks, then check if the semaphore is still locked and if so acquire once more"?

amartsinovich-dialpad: But we have no idea if this last lock will ever unlock. Odds are, it won't. We need to be cancelled precisely on line 200 to be able to acquire the last lock.

jonathan-johnston (commented):
I think I see what's going on here. There's a race condition between `await semaphore.acquire()` in the `while True` loop and the actual scheduling of the task and its associated callback to `release` the semaphore. When the consumer is cancelled it may still be holding the lock but tries to wait on itself, thus the deadlock.

Something like this should work in theory, using `asyncio.shield` to make the `acquire` and `add_done_callback` block essentially guaranteed to run uninterrupted by cancellation:

```python
async def consumer(...):
    try:
        semaphore = asyncio.Semaphore(max_tasks)

        async def create_task() -> None:
            await semaphore.acquire()
            # the rest of the code
            task.add_done_callback(lambda _f: semaphore.release())
            message_queue.task_done()

        while True:
            await asyncio.shield(create_task())
    except asyncio.CancelledError:
        # exception handling code as you wrote it previously
```

pubsub/gcloud/aio/pubsub/subscriber.py
shaundialpad: What if we move the `await semaphore.acquire()` somewhere where it is the last thing awaited in the `while True` loop? Like just before we schedule the task:

```python
                await semaphore.acquire()
                task = asyncio.ensure_future(_execute_callback(
                    message,
                    callback,
                    ack_queue,
                    nack_queue,
                    metrics_client,
                ))
```

amartsinovich-dialpad: But we still can be cancelled after acquiring a lock but before scheduling a task => deadlock

amartsinovich-dialpad:
@jonathan-johnston thank you sir you prevented a very miserable solution from going to prod.
I tested shielded consumer in egress and it seems to be working fine.

pubsub/gcloud/aio/pubsub/subscriber.py
shaundialpad: Can we? AFAIK a task can only be cancelled at `await` statements (and I suppose before it begins) and not on any given line. Please enlighten me if I am mistaken! 

But I think `shield`ing as @jonathan-johnston pointed out would also work.

shaundialpad (commented):
Great! How did this scenario evade the tests before, and is there a test we can add to ensure it works going forward? I presume we expect users to cancel the `subscribe` task to engage shutdown. Maybe we can test this does what is expected?

TheKevJames (approved):
LGTM. Fantastic find on `asyncio.shield`, @jonathan-johnston !

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: Tbh I don't know how to read docs on that:
image
image
So to be cancelled a task needs a new cycle but if we're awaiting on a future, the future will get cancelled instead. I think you're right here.

amartsinovich-dialpad:
@shaundialpad Yeah good call. The test to guard against it is to simply cancel subscriber that has nothing to do. Added that

shaundialpad (approved):

jonathan-johnston (approved):

PR #282, opened 2021-01-14
chore(pubsub) Close subscriber session during shutdown

amartsinovich-dialpad:
This should make shutdown a little better on logging side

PR #283, opened 2021-01-14
chore(pubsub): bump to 4.0.3

jonathan-johnston:
Pull in #282


amartsinovich-dialpad (approved):
omg that was quick! <3

PR #285, opened 2021-01-14
Revert "chore(pubsub) Close subscriber session during shutdown"

amartsinovich-dialpad:
Reverts talkiq/gcloud-aio#282

Yes, so `SubscriberClient` is a divine artifact given to subscriber by the client, and we're not in a position to decide its fate

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -324,5 +324,4 @@ else:
             for task in acker_tasks:
                 task.cancel()
             await asyncio.wait(acker_tasks, return_when=asyncio.ALL_COMPLETED)
-            await subscriber_client.session.close()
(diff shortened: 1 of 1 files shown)


jonathan-johnston (approved):

TheKevJames (approved):
Ah, yes, you should only deconstruct objects which you were explicitly responsible for creating. Since users need to pass in the `subscriber_client`, deconstructing them is their responsibility.

Looks like the SubscriberClient might be missing some of the methods we've added to make this convenient in other libs -- see, for example, [these KMS convenience methods](https://github.com/talkiq/gcloud-aio/blob/5dc9c21379c01b77ea38b3b126c99988b940728b/kms/gcloud/aio/kms/kms.py#L85-L92) which allow users to do either:
```python
kms = KMS(..)
# SNIP
await kms.close()
```
or
```python
async with KMS(..) as kms:
    # SNIP
# auto-closed here
```

PR #286, opened 2021-01-15
feat(pubsub): list, create, delete topics

shaundialpad:
Adds some more methods to the `PublisherClient`. Method signatures analogous to those implemented for our [SubscriberClient](https://github.com/talkiq/gcloud-aio/blob/daed41cac123db134835329d9e7365f7b9ad8911/pubsub/gcloud/aio/pubsub/subscriber_client.py).
- [x] manual testing has passed (creation of topics, deletion of topics, listing of topics)

### DIFF
--- a/pubsub/gcloud/aio/pubsub/publisher_client.py
+++ b/pubsub/gcloud/aio/pubsub/publisher_client.py
@@ -65,6 +65,49 @@ class PublisherClient:
     # TODO: implement that various methods from:
     # https://github.com/googleapis/python-pubsub/blob/master/google/cloud/pubsub_v1/gapic/publisher_client.py
 
+    # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics/list
+    async def list_topics(self, project: str,
+                          query_params: Optional[Dict[str, str]] = None,
+                          *, session: Optional[Session] = None
+                          ) -> Dict[str, Any]:
+        """
+        List topics
+        """
+        url = f'{API_ROOT}/{project}/topics'
+        headers = await self._headers()
+        s = AioSession(session) if session else self.session
+        resp = await s.get(url, headers=headers, params=query_params)
+        result: Dict[str, Any] = await resp.json()
+        return result
+
+    # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics/create
+    async def create_topic(self, topic: str,
+                           body: Optional[Dict[str, Any]] = None,
+                           *, session: Optional[Session] = None
+                           ) -> Dict[str, Any]:
+        """
+        Create topic.
+        """
+        url = f'{API_ROOT}/{topic}'
+        headers = await self._headers()
+        encoded = json.dumps(body or {}).encode()
+        s = AioSession(session) if session else self.session
+        resp = await s.put(url, data=encoded, headers=headers, timeout=60)
+        result: Dict[str, Any] = await resp.json()
+        return result
+
+    # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics/delete
+    async def delete_topic(self, topic: str,
+                           *, session: Optional[Session] = None
+                           ) -> None:
+        """
+        Delete topic.
+        """
+        url = f'{API_ROOT}/{topic}'
+        headers = await self._headers()
+        s = AioSession(session) if session else self.session
+        await s.delete(url, headers=headers)
+
     # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics/publish
     async def publish(self, topic: str, messages: List[PubsubMessage],
                       session: Optional[Session] = None,
--- a/pubsub/gcloud/aio/pubsub/publisher_client.py
+++ b/pubsub/gcloud/aio/pubsub/publisher_client.py
@@ -83,8 +83,8 @@ class PublisherClient:
     # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics/create
     async def create_topic(self, topic: str,
                            body: Optional[Dict[str, Any]] = None,
-                           timeout: int = 60,
-                           *, session: Optional[Session] = None
+                           *, session: Optional[Session] = None,
+                           timeout: Optional[int] = None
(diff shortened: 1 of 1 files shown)


pubsub/gcloud/aio/pubsub/publisher_client.py
```diff
@@ -65,6 +65,49 @@ async def _headers(self) -> Dict[str, str]:
...
+                           *, session: Optional[Session] = None
+                           ) -> Dict[str, Any]:
+        """
+        Create topic.
+        """
+        url = f'{API_ROOT}/{topic}'
+        headers = await self._headers()
+        encoded = json.dumps(body or {}).encode()
+        s = AioSession(session) if session else self.session
+        resp = await s.put(url, data=encoded, headers=headers, timeout=60)
```
shaundialpad, line 95: This is worth notable. I increased the timeout here from our default of 10 seconds, as it appears (based on my own experimentation) the deletion of a topic and immediate re-creation of the same topic requires at least 30 seconds to pass.

egalpin: Can we expose it as a parameter of `create_topic` and get the best of both worlds?

egalpin (approved):
LGTM, just the outstanding question of 60s `create_topic` timeout

pubsub/gcloud/aio/pubsub/publisher_client.py
shaundialpad: Sounds reasonable! If not provided we probably want to use the default that `AioSession.put` uses (which is 10 seconds). So might be best to except the general `**kwargs` and pass those long? 

```python
    async def create_topic(self, topic: str,
                           body: Optional[Dict[str, Any]] = None,
                           *, session: Optional[Session] = None,
                           **kwargs) -> Dict[str, Any]:
        # ...
        resp = await s.put(url, data=encoded, headers=headers, **kwargs)
        # ...
```

Unless we wanted make it explicit that this is intended for the `timeout` parameter, in which case we could do something like 

```python
    async def create_topic(self, topic: str,
                           body: Optional[Dict[str, Any]] = None,
                           *, session: Optional[Session] = None,
                           timeout: Optional[int] = None,
                           ) -> Dict[str, Any]:
        # ...
        kwargs = {'timeout': timeout} if timeout is not None else {}
        resp = await s.put(url, data=encoded, headers=headers, **kwargs)
        # ...
```

shaundialpad: Hmm okay well here's another consideration: the default for the grpc client is just to wait the extra time until the topic is successfully created. Perhaps we _should_ make the default 60s in light of this, and allow it to be overridden if desired (instead of using our current default of 10 seconds).

amartsinovich-dialpad: oh, I was just adding same timeouts to `SubscriberClient` https://github.com/talkiq/gcloud-aio/pull/287
We should probably align timeout usage between them

amartsinovich-dialpad: Hmm, what is the semantics of putting an argument before of after `*`? I was assuming we put endpoint-specific staff first and generic http overridables last

shaundialpad: Ah yes :P That's a good question and I am not exactly sure but that makes sense to me. So, last?

amartsinovich-dialpad: I'm biased because that's where I put it :D

amartsinovich-dialpad: @TheKevJames @egalpin opinions?

egalpin: before the `*` are required positional args (also possible to specify as keyword), after `*` are keyword-only args i.e. cannot be specified by position Ex. `create_topic(a, b, c, session=...)` is ok, `create_topic(a, b, c, session)` is not allowed

shaundialpad: Right, so when would one want to enforce a keyword argument?

TheKevJames: Hmm, I don't think either the deletion or the creation endpoint is the right place to add this extra latency. We probably want to keep the defaults low (ie. 10s like we use elsewhere), but perhaps explain something in the docs? Like, if users are calling `create_topic` without having deleted that topic recently, they probably want the default low timeout value. If they happen to have the following code, though, they probably want to override:
```python
await client.delete_topic(..)
await client.create_topic(.., timeout=30)  # creating a topic right after deleting it takes longer than usual
```

We should try to avoid making the standard case worse to acocunt for an edge-case, if possible!

shaundialpad: Alright, its settled, I will adapt @amartsinovich-dialpad 's pattern for consistency and we will default to our current 10 seconds (or whatever it happens to be), but allow overriding as @TheKevJames illustrated.

pubsub/gcloud/aio/pubsub/publisher_client.py
```diff
@@ -65,6 +65,54 @@ async def _headers(self) -> Dict[str, str]:
...
+        Create topic.
+        """
+        url = f'{API_ROOT}/{topic}'
+        headers = await self._headers()
+        encoded = json.dumps(body or {}).encode()
+        s = AioSession(session) if session else self.session
+        kwargs = {
+            'timeout': timeout,
+        }
+        kwargs = {k: v for k, v in kwargs.items() if v is not None}
```
TheKevJames, line 99: Rather than this kwargs stuff, I think it makes more sense to just pass in the `timeout` default which makes the most sense for `create_topic` and pass it as-is; sounds like `10` is still the right value here. See my comment on Alex's PR for more info!

TheKevJames (changes requested):

amartsinovich-dialpad (approved):

TheKevJames (approved):

PR #287, opened 2021-01-15
fix(pubsub) Increase long polling timeout to 30s

amartsinovich-dialpad:
So at the moment we use a default `10s` receive timeout for long-polling pull requests. It turned out it works horribly under low traffic.
What user sees: things not working
What monitoring shows: both `num_of_unacked_messages` and `oldest_unacked_message` would spike, then few messages would get into a deadletter topic.
What debugging showed: some messages simply never get processed by application
Mind that I'm talking about the queue which has several actively listening consumers.

What I think is happening. Producer workers would open long-polling connections, drop them after 10 seconds, then open a new set etc. When a message is finally published, pubsub server tries to evenly distribute them between open polling requests and a lot of them have no actual client connected. So after several such publishes a message gets into deadletter even though it was never delivered.

Now for the fix. I check a few libs to see what timeout they're using. Broadway uses `infinity`, spotify's sunsetting pubsub lib does `returnImmediately: True`, google libs strongly prefer `StreamingPull`. I also find [this obscure](https://tools.ietf.org/id/draft-loreto-http-bidirectional-07.html#timeouts) link which has an abbreviation as a domain name and simple html design which usually indicates quality HTTP documentation. It recommends 30 seconds.

I tested 30 seconds and it works like a charm. I couldn't lose a single message.

image

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -224,7 +224,9 @@ else:
                 new_messages = []
                 try:
                     new_messages = await subscriber_client.pull(
-                        subscription=subscription, max_messages=max_messages)
+                        subscription=subscription,
+                        max_messages=max_messages,
+                        timeout=30)
                 except (asyncio.TimeoutError, KeyError):
                     continue
 
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -226,6 +226,10 @@ else:
                     new_messages = await subscriber_client.pull(
                         subscription=subscription,
                         max_messages=max_messages,
+                        # it is important to have this value reasonably high
+                        # as long lived connections may be left hanging
+                        # on a server which will cause delay in message
+                        # delivery or even false deadlettering
                         timeout=30)
                 except (asyncio.TimeoutError, KeyError):
                     continue
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -229,7 +229,7 @@ else:
                         # it is important to have this value reasonably high
                         # as long lived connections may be left hanging
                         # on a server which will cause delay in message
-                        # delivery or even false deadlettering
+                        # delivery or even false deadlettering if it is enabled
                         timeout=30)
                 except (asyncio.TimeoutError, KeyError):
                     continue
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -63,6 +63,7 @@ class SubscriberClient:
                                   body: Optional[Dict[str, Any]] = None,
                                   *,
                                   session: Optional[Session] = None,
+                                  timeout: Optional[int] = None
                                   ) -> Dict[str, Any]:
         """
         Create subscription.
@@ -74,7 +75,11 @@ class SubscriberClient:
         payload.update({'topic': topic})
         encoded = json.dumps(payload).encode()
         s = AioSession(session) if session else self.session
-        resp = await s.put(url, data=encoded, headers=headers)
+        kwargs = {
+            'timeout': timeout,
+        }
+        kwargs = {k: v for k, v in kwargs.items() if v is not None}
+        resp = await s.put(url, data=encoded, headers=headers, **kwargs)
         result: Dict[str, Any] = await resp.json()
         return result
 
@@ -82,7 +87,8 @@ class SubscriberClient:
     async def delete_subscription(self,
                                   subscription: str,
                                   *,
-                                  session: Optional[Session] = None
+                                  session: Optional[Session] = None,
+                                  timeout: Optional[int] = None
                                   ) -> None:
         """
         Delete subscription.
@@ -90,11 +96,16 @@ class SubscriberClient:
         url = f'{API_ROOT}/v1/{subscription}'
         headers = await self._headers()
         s = AioSession(session) if session else self.session
-        await s.delete(url, headers=headers)
+        kwargs = {
+            'timeout': timeout,
+        }
+        kwargs = {k: v for k, v in kwargs.items() if v is not None}
+        await s.delete(url, headers=headers, **kwargs)
 
     # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/pull
     async def pull(self, subscription: str, max_messages: int,
-                   *, session: Optional[Session] = None
+                   *, session: Optional[Session] = None,
+                   timeout: Optional[int] = None
                    ) -> List[SubscriberMessage]:
         """
         Pull messages from subscription
@@ -106,7 +117,11 @@ class SubscriberClient:
         }
         encoded = json.dumps(payload).encode()
         s = AioSession(session) if session else self.session
-        resp = await s.post(url, data=encoded, headers=headers)
+        kwargs = {
+            'timeout': timeout,
+        }
+        kwargs = {k: v for k, v in kwargs.items() if v is not None}
+        resp = await s.post(url, data=encoded, headers=headers, **kwargs)
         resp = await resp.json()
         return [
             SubscriberMessage.from_repr(m)
@@ -115,7 +130,8 @@ class SubscriberClient:
 
     # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/acknowledge
     async def acknowledge(self, subscription: str, ack_ids: List[str],
-                          *, session: Optional[Session] = None) -> None:
+                          *, session: Optional[Session] = None,
+                          timeout: Optional[int] = None) -> None:
         """
         Acknowledge messages by ackIds
         """
@@ -126,13 +142,18 @@ class SubscriberClient:
         }
         encoded = json.dumps(payload).encode()
         s = AioSession(session) if session else self.session
-        await s.post(url, data=encoded, headers=headers)
+        kwargs = {
+            'timeout': timeout,
+        }
+        kwargs = {k: v for k, v in kwargs.items() if v is not None}
+        await s.post(url, data=encoded, headers=headers, **kwargs)
 
     # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/modifyAckDeadline
     async def modify_ack_deadline(self, subscription: str,
                                   ack_ids: List[str],
                                   ack_deadline_seconds: int,
-                                  *, session: Optional[Session] = None
+                                  *, session: Optional[Session] = None,
+                                  timeout: Optional[int] = None
                                   ) -> None:
         """
         Modify messages' ack deadline.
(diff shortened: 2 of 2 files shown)


shaundialpad (commented):
Very interesting indeed! So based on your theory, perhaps we are not closing our long-polling requests properly with goog?

amartsinovich-dialpad:
@shaundialpad Yes, I think so. We just tell `aiohttp` to timeout after 10 seconds. 30 seconds seem to lack this problem or at least it makes the issue very very unlikely to happen. Do you have a better way to manage connection in mind?

shaundialpad:
> Do you have a better way to manage connection in mind?

Not really - a brief search on the www seems to indicate this is the accepted solution, e.g. https://github.com/googleapis/google-cloud-php/issues/710

Interesting the above post mentions a quote corroborating this from [google's pubsub troubleshooting doc](https://cloud.google.com/pubsub/docs/troubleshooting), but the quote no longer exists 😱

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -74,27 +75,37 @@ async def create_subscription(self,
         payload.update({'topic': topic})
         encoded = json.dumps(payload).encode()
         s = AioSession(session) if session else self.session
-        resp = await s.put(url, data=encoded, headers=headers)
+        kwargs = {
+            'timeout': timeout,
+        }
+        kwargs = {k: v for k, v in kwargs.items() if v is not None}
```
shaundialpad, line 81: Since we're not accepting the general `**kwargs` pattern in the function signature, this seems acceptable for adding new kwargs in the future to pass along. Is there anything else we may want to override? If not it might be simpler to just do `kwargs = {} if timeout is None else {'timeout': timeout}`

shaundialpad (approved):
Just a comment, but the increased timeout seems to be the accepted practice.

pubsub/gcloud/aio/pubsub/subscriber_client.py
amartsinovich-dialpad: well yeah. It just seems more welcoming for future change to add 1 key to the dicts, than changing whole expression in every function?

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -224,7 +224,9 @@ async def producer(
                 new_messages = []
                 try:
                     new_messages = await subscriber_client.pull(
-                        subscription=subscription, max_messages=max_messages)
+                        subscription=subscription,
+                        max_messages=max_messages,
+                        timeout=30)
```
jonathan-johnston, line 229: Could you document your tribal knowledge here? Otherwise this 30 second timeout will look like an arbitrarily-selected value to those who wish to modify this in the future.

pubsub/gcloud/aio/pubsub/subscriber_client.py
jonathan-johnston: Aren't `None` values acceptable for some arguments?

jonathan-johnston (commented):
It's a shame that `returnImmediately` is deprecated... There seems to be no good way of closing the long-polling HTTP session then?

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: a comment will do or you mean something more solid?

pubsub/gcloud/aio/pubsub/subscriber_client.py
amartsinovich-dialpad: @jonathan-johnston right, they are. It doesn't directly conflicts with the current code, since you can always put the key in the kwargs dict after the filter was applied, but if you think `kwargs = {} if timeout is None else {'timeout': timeout}` approach is better I'm totally ok with it

jonathan-johnston: Ah, maybe I'm misunderstanding the need for the filter on `None` values. The original code had a 10s timeout but if you allow `None` values here (which would mean `timeout=None` is piped through) then timeouts would no longer be a default 10s but infinite right? Therefore filtering out `None` values keeps the default `timeout=10`.

pubsub/gcloud/aio/pubsub/subscriber.py
jonathan-johnston: A comment would be fine 🙂

pubsub/gcloud/aio/pubsub/subscriber_client.py
amartsinovich-dialpad: Yes, `None` filter here is to ensure "not specified" semantics. But you have a point of `None` being a valid value for "timeout" param the means no timeout at all... ugh, how do we get out of wrapping hell then?

jonathan-johnston: Do we need to use `kwargs` here? Maybe just set a default value for `timeout` that matches the existing timeout of 10 seconds and pass `timeout` through directly?

amartsinovich-dialpad: Yeah, that'll work, just feels wrong to me duplicating default value. @shaundialpad @TheKevJames opinions?

shaundialpad: If `None` is a valid timeout, then I suppose `kwargs` might be the best option to avoid duplicating the default value.

amartsinovich-dialpad: I can recall @TheKevJames suggesting to avoid `kwargs` when possible. But this might actually be _that case_?

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -224,7 +224,13 @@ async def producer(
                 new_messages = []
                 try:
                     new_messages = await subscriber_client.pull(
-                        subscription=subscription, max_messages=max_messages)
+                        subscription=subscription,
+                        max_messages=max_messages,
+                        # it is important to have this value reasonably high
+                        # as long lived connections may be left hanging
+                        # on a server which will cause delay in message
+                        # delivery or even false deadlettering
```
shaundialpad, line 232: Maybe - "even false deadlettering if it is enabled"

pubsub/gcloud/aio/pubsub/subscriber_client.py
TheKevJames: So my general rule would be to avoid `kwargs` by default since it hides the exact arguments (ie. "explicit is better than implicit") and tends to make debugging more difficult accordingly. That's a lot more on the method arguments side of things than in calling code, mind -- ie. I appreciate that you have this method taking `session = None, timeout = None` as args instead of `**kwargs` and pulling them out from a readability perspective -- ie. this tells us that "this function cares about timeout values". That makes sense given how the `put()` method works; eg. see its method definition:
```python
@abstractmethod
async def put(self, url: str, headers: Dict[str, str], data: IO[Any],
              timeout: int) -> Response:
    pass
```
Since that function explicitly asks for a `timeout` argument, we should in turn explicitly pass it one. In that case, since we're in a different context, it makes sense to consider what our default should be here (which may just-so-happen to be a duplicate value!) and mark it as such. In other words:
```python
async def create_subscription(...,  timeout: int = 10) -> Dict[str, Any]:
    # SNIP..
    resp = await s.put(url, data=encoded, headers=headers, timeout=timeout)
```

### Why Is Duplicating This Value Ok?

So I think the whole "Don't Repeat Yourself" thing has taken on a muddled meaning over the years. "Don't Repeat Yourself" should not mean "literally do not have duplicate code anywhere". It instead should mean "do not duplicate _concepts_ which should be constant in all usecases".

_This is not one of those cases_! The session's `def put(.., timeout=10)` value is describing "what should the default timeout value be for _arbitrary put requests made to Google services_"? The subscriber's `def create_subscription(.., timeout=10)` value is describing "what should the default timeout value be for _the create subscription request in particular_"? The fact that those have identical values **is not duplication**.

### When Would We Use Kwargs?

Kwargs are super useful in cases where you want to signal that your method does not (**and should not**) care about a given argument. Most good examples along these lines have to do with inheritence-style composition: for example, consider the standard OO case of having an `Animal` class and a `Dog` subclass. You probably want the `Dog` subclass to be defined with methods along the lines of "do these  Dog-specific things... and then also whatever the more general Animal stuff is". For example:
```python
class Animal:
    def __init__(self, name: str, species: str) -> None:
        # SNIP

class Dog(Animal):
    def __init__(self, **kwargs) -> None:
        super()._init__(self, **kwargs)
```

This lets you be more explicit about which things you are changing, ie. something along the lines of:
```python
class Animal:
    def __init__(self, name: str, species: str) -> None:
        # SNIP

class Dog(Animal):
    def __init__(self, has_dog_breath: bool, **kwargs) -> None:
        self._needs_a_toothbrush = has_dog_breath
        super()._init__(self, **kwargs)
```

In this specific example, using `kwargs` in the `create_subscription()` argument would be tying the implementation details together too closely -- it would be akin to saying "whatever extra stuff the put method does"... which would be pretty confusing and implicit to users, since they then need to know that we just-so-happen to be using `Session.put` later on in the method, which is an implementation detail our users should not have to care about!

amartsinovich-dialpad: @TheKevJames That's ok, I don't mind duplicating this value at this point.

but for the joy of intellectual discussion let me clarify one point! I don't like having a stack of wrappers not because I think it violates DRY (anyone who saw my tests knows I have no problems hitting cmd+v hitting cmd+v) but mostly because it _feels wrong_. And I think this time there is an actual reason to this.

>it would be akin to saying "whatever extra stuff the put method does"... which would be pretty confusing and implicit to users, since they then need to know that we just-so-happen to be using Session.put later on in the method, which is an implementation detail our users should not have to care about!

I think we already have a much more confusing behavior. If you need to browse the source code to understand what's hiding behind the `kwargs`, if you have to know that the package uses `aiohttp` under the hood to effectively work with it – I agree this is a bit inconvenient, but ultimately this is a virtue because in expands user's true knowledge about the system.
What we did is that we restricted api flexibility by being overly zealous about explicit args. `aiohttp` has a ton of parameters one can customize per-request, but we decided we know better. There are actually 3 timeout parameters, but we only allow one of them.  What I find the most confusing is `session` parameter. Our type hints clearly suggests that we expect an optional `aiohttp.ClientSession` session. A user can put a lot of per-session parameters in there and most of them will work... but not `timeout`, because we force override it on per-request level. And this is truly confusing because a user will rightly assume that session-level timeout will work, but it won't because of an implementation detail.

tl;dr I think saying "this is our method and by the way we will happily pass any kwargs to `aiohttp`, so please consult its docs" is better then "this is out method's signature and that's all you need to know"

TheKevJames: That's a great point, Alex! I do agree that we could benefit from expanding this (although, perhaps with some renames; eg. something like `**http_kwargs` might make more sense, since our code is not 100% strictly passthroughs only to http requests and IIRC we have cases were the kwargs are used for other things).

I think the main point of contention for me is that this is not how the library _currently_ works. I definitely agree that allowing users this flexible would be a benefit, but I think keeping things consistent is more valuable than having this only be the case for _some_ methods; I'd support switching the `gcloud-aio` libs over to this style more holistically, but not here and there as we modify things, since having each method behave differently would be quite confusing!

As to whether the value of exposing these params is worth the sum effort of changing everything? _Probably_, yeah, though for it to fit into our scheduled non-20% work we'd have to be running into places where we'd make use of those params ourselves :P

amartsinovich-dialpad: totally agree with `http_kwargs`. And yeah, not worth to change everything at this point, those are mostly consideration for the next time we will design client libraries or anything like that.

PR #289, opened 2021-01-22
Shaun/pubsub/fix publisher emulation

shaundialpad:
Pubsub topic creation was failing when using the emulator:

```
aiohttp.client_exceptions.ClientResponseError: 404, message='Not Found: Not Found\n', url=URL('http://pubsub:8681/v1/projects/my-proj/topics/my-topic')
```

Specifying the default headers `{'Content-Type': 'application/json'}` (which we use in our subscriber client) included in this PR fixed it.

I also updated the `verify_ssl` behavior for emulation we use for the subscriber client to exist in the publisher client.

- [x] manual testing passed (with and without emulator)

### DIFF
--- a/pubsub/gcloud/aio/pubsub/publisher_client.py
+++ b/pubsub/gcloud/aio/pubsub/publisher_client.py
@@ -56,13 +56,17 @@ class PublisherClient:
         return f'{cls.project_path(project)}/topics/{topic}'
 
     async def _headers(self) -> Dict[str, str]:
+        headers = {
+            'Content-Type': 'application/json'
+        }
         if PUBSUB_EMULATOR_HOST:
-            return {}
+            return headers
 
         token = await self.token.get()
-        return {
-            'Authorization': f'Bearer {token}',
-        }
+        headers.update({
+            'Authorization': f'Bearer {token}'
+        })
+        return headers
 
     # TODO: implement that various methods from:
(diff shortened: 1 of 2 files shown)


amartsinovich-dialpad (approved):

pubsub/gcloud/aio/pubsub/publisher_client.py
```diff
@@ -54,13 +56,17 @@ def topic_path(cls, project: str, topic: str) -> str:
...
+        }
         if PUBSUB_EMULATOR_HOST:
-            return {}
+            return headers
 
         token = await self.token.get()
-        return {
-            'Authorization': f'Bearer {token}',
-        }
+        headers.update({
```
TheKevJames, line 66: nit: overkill, use `headers['Authorization'] = f'Bearer {token}'`

Generally, treat `update` as a slower multitool that you need to pull out only for edge-cases; it lets you do more advanced dictionary merging behavior, but that comes with something like a >5x perf loss. Obviously, dictionary updates are "fast enough" for this to not matter all that much, but it's a good default to have.

TheKevJames (changes requested):

pubsub/gcloud/aio/pubsub/publisher_client.py
shaundialpad: Thanks for the stat! I just copied this from subscriber so I'll update it there too to keep them consistent.

TheKevJames (approved):

PR #291, opened 2021-01-26
fix(pubsub) account for missing data key in SubscriberMessage

amartsinovich-dialpad:
Just like `attributes` key, `data` can also be missing.


pubsub/gcloud/aio/pubsub/subscriber_message.py
```diff
@@ -39,12 +40,15 @@ def from_repr(received_message: Dict[str, Any]
...
             'message': {
                 'messageId': self.message_id,
-                'attributes': self.attributes,
-                'data': base64.b64encode(self.data),
                 'publishTime': self.publish_time.strftime('%Y-%m-%dT%H:%M:%SZ')
             }
         }
+        if self.attributes:
+            r['message']['attributes'] = self.attributes
+        if self.data:
```
jonathan-johnston, line 52: Should this be `is not None`? Is sending an empty string a valid message?

jonathan-johnston (approved):

PR #293, opened 2021-01-28
feat(pubsub): retrieve deliveryAttempt on subscriber pull

shaundialpad:
The `delivery_attempt` attribute is used with deadlettering and comes standard with the grpc python subscriber client. When a subscription is enabled with deadlettering, this attribute will be an integer. If it is not, it will default to `None` (same behavior as the grpc clients).

Note also the direct http responses to subscription pull requests:

Deadlettering enabled
```
{
  "receivedMessages": [
    {
      "ackId": "<trimmed>",
      "message": {
        "data": "<b64encoded>",
        "messageId": "1961954308846553",
        "publishTime": "2021-01-28T20:32:34.033Z"
      },
      "deliveryAttempt": 1
    }
  ]
}
```

Deadlettering not enabled (note deliveryAttempt is absent, despite [docs](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/pull#receivedmessage) saying it will be 0)
```
{
  "receivedMessages": [
    {
      "ackId": "<trimmed>",
      "message": {
        "data": "<b64encoded>",
        "attributes": {
          "foo": "bar"
        },
        "messageId": "1961954411924135",
        "publishTime": "2021-01-28T20:35:04.738Z"
      }
    }
  ]
}
```

- [x] tested locally on subscriptions with and without deadlettering
- [x] tests added

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber_message.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_message.py
@@ -19,7 +19,7 @@ class SubscriberMessage:
                  publish_time: 'datetime.datetime',
                  data: Optional[bytes],
                  attributes: Optional[Dict[str, Any]],
-                 delivery_attempt: Optional[int]):
+                 delivery_attempt: Optional[int] = None):
         self.ack_id = ack_id
         self.message_id = message_id
         self.publish_time = publish_time
--- a/pubsub/gcloud/aio/pubsub/subscriber_message.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_message.py
@@ -18,12 +18,14 @@ class SubscriberMessage:
     def __init__(self, ack_id: str, message_id: str,
                  publish_time: 'datetime.datetime',
                  data: Optional[bytes],
-                 attributes: Optional[Dict[str, Any]]):
+                 attributes: Optional[Dict[str, Any]],
+                 delivery_attempt: Optional[int]):
         self.ack_id = ack_id
         self.message_id = message_id
         self.publish_time = publish_time
         self.data = data
         self.attributes = attributes
+        self.delivery_attempt = delivery_attempt
 
     @staticmethod
     def from_repr(received_message: Dict[str, Any]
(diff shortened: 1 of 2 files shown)


TheKevJames (approved):

pubsub/gcloud/aio/pubsub/subscriber_message.py
```diff
@@ -46,5 +50,6 @@ def to_repr(self) -> Dict[str, Any]:
                 'attributes': self.attributes,
                 'data': base64.b64encode(self.data),
                 'publishTime': self.publish_time.strftime('%Y-%m-%dT%H:%M:%SZ')
-            }
+            },
+            'deliveryAttempt': self.delivery_attempt
```
amartsinovich-dialpad, line 54: that... is not how master code looks like https://github.com/talkiq/gcloud-aio/blob/master/pubsub/gcloud/aio/pubsub/subscriber_message.py#L50
what happened here?

amartsinovich-dialpad (changes requested):

pubsub/gcloud/aio/pubsub/subscriber_message.py
shaundialpad: Welp that's why there's merge conflicts. Lemme rebase

shaundialpad: Fixed thanks!

amartsinovich-dialpad (approved):
totally not surprised by the missing key, exactly same thing happening with data and attributes

pubsub/gcloud/aio/pubsub/subscriber_message.py
```diff
@@ -18,12 +18,14 @@ class SubscriberMessage:
     def __init__(self, ack_id: str, message_id: str,
                  publish_time: 'datetime.datetime',
                  data: Optional[bytes],
-                 attributes: Optional[Dict[str, Any]]):
+                 attributes: Optional[Dict[str, Any]],
+                 delivery_attempt: Optional[int]):
```
shaundialpad, line 22: Hmm just thought of something. I think this is technically a breaking change, since if anyone is instantiating `SubscriberMessage` directly now without specifying `delivery_attempt`, it will fail. Should be perhaps default this to `None` to avoid bumping major versions? @amartsinovich-dialpad @TheKevJames

TheKevJames: Ah, yes, good catch -- make this default to `None` for now, and if we think it _should_ be something users have to specify, we can add a TODO to deprecate this in a future release

shaundialpad: True - I'm not sure it is something users need to specify, as if they are not using deadlettering, it is of no concern to them.  I'll leave out the TODO.

PR #298, opened 2021-02-11
fix(pubsub) Drop whole batch if (n)ack request throws 400

amartsinovich-dialpad:
Context: VAE-2332

tl;dr There are cases when  a certain `ack_id` cannot be acked no matter what. In that case whole batch will be contaminated and won't be able to get through. The best we can do if we encounter such error is to drop the batch and schedule individual ack requests for each `ack_id`.

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -106,11 +106,12 @@ else:
                     aiohttp.client_exceptions.ClientResponseError)
                 if is_response_error and e.status == 400:  # type: ignore # pylint: disable=no-member
                     log.warning(
-                        'Ack error is unrecoverable, dropping the whole batch')
-                    for i in ack_ids:
+                        'Ack error is unrecoverable, '
+                        'one or more messages may be dropped')
+                    for ack_id in ack_ids:
                         asyncio.ensure_future(
                             subscriber_client.acknowledge(subscription,
-                                                          ack_ids=[i])
+                                                          ack_ids=[ack_id])
                         )
                     ack_ids = []
 
@@ -157,13 +158,13 @@ else:
                     aiohttp.client_exceptions.ClientResponseError)
                 if is_response_error and e.status == 400:  # type: ignore # pylint: disable=no-member
                     log.warning(
-                        'Nack error is unrecoverable,'
-                        ' dropping the whole batch')
-                    for i in ack_ids:
+                        'Nack error is unrecoverable, '
+                        'one or more messages may be dropped')
+                    for ack_id in ack_ids:
                         asyncio.ensure_future(
                             subscriber_client.modify_ack_deadline(
                                 subscription,
-                                ack_ids=[i],
+                                ack_ids=[ack_id],
                                 ack_deadline_seconds=0)
                         )
                     ack_ids = []
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -3,6 +3,7 @@ from gcloud.aio.auth import BUILD_GCLOUD_REST
 if BUILD_GCLOUD_REST:
     pass
 else:
+    import aiohttp
     import asyncio
     import logging
     import time
@@ -99,6 +100,20 @@ else:
                 log.warning(
                     'Ack request failed, better luck next batch', exc_info=e)
                 metrics_client.increment('pubsub.acker.batch.failed')
+
+                is_response_error = isinstance(
+                    e,
+                    aiohttp.client_exceptions.ClientResponseError)
+                if is_response_error and e.status == 400:  # type: ignore # pylint: disable=no-member
+                    log.warning(
+                        'Ack error is unrecoverable, dropping the whole batch')
+                    for i in ack_ids:
+                        asyncio.ensure_future(
+                            subscriber_client.acknowledge(subscription,
+                                                          ack_ids=[i])
+                        )
+                    ack_ids = []
+
                 continue
 
             metrics_client.histogram('pubsub.acker.batch', len(ack_ids))
@@ -136,6 +151,23 @@ else:
                 log.warning(
                     'Nack request failed, better luck next batch', exc_info=e)
                 metrics_client.increment('pubsub.nacker.batch.failed')
+
+                is_response_error = isinstance(
+                    e,
+                    aiohttp.client_exceptions.ClientResponseError)
+                if is_response_error and e.status == 400:  # type: ignore # pylint: disable=no-member
+                    log.warning(
+                        'Nack error is unrecoverable,'
+                        ' dropping the whole batch')
+                    for i in ack_ids:
+                        asyncio.ensure_future(
+                            subscriber_client.modify_ack_deadline(
+                                subscription,
+                                ack_ids=[i],
+                                ack_deadline_seconds=0)
+                        )
+                    ack_ids = []
+
                 continue
 
             metrics_client.histogram('pubsub.nacker.batch', len(ack_ids))
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -96,24 +96,27 @@ else:
             try:
                 await subscriber_client.acknowledge(subscription,
                                                     ack_ids=ack_ids)
-            except Exception as e:
-                log.warning(
-                    'Ack request failed, better luck next batch', exc_info=e)
-                metrics_client.increment('pubsub.acker.batch.failed')
-
-                is_response_error = isinstance(
-                    e,
-                    aiohttp.client_exceptions.ClientResponseError)
-                if is_response_error and e.status == 400:  # type: ignore # pylint: disable=no-member
-                    log.warning(
+            except aiohttp.client_exceptions.ClientResponseError as e:
+                if e.status == 400:
+                    log.error(
                         'Ack error is unrecoverable, '
-                        'one or more messages may be dropped')
+                        'one or more messages may be dropped', exc_info=e)
                     for ack_id in ack_ids:
                         asyncio.ensure_future(
                             subscriber_client.acknowledge(subscription,
                                                           ack_ids=[ack_id])
                         )
                     ack_ids = []
+
+                log.warning(
+                    'Ack request failed, better luck next batch', exc_info=e)
+                metrics_client.increment('pubsub.acker.batch.failed')
+
+                continue
+            except Exception as e:
+                log.warning(
+                    'Ack request failed, better luck next batch', exc_info=e)
+                metrics_client.increment('pubsub.acker.batch.failed')
 
                 continue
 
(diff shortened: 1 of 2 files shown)


pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -99,6 +100,20 @@ async def acker(subscription: str,
...
+                is_response_error = isinstance(
+                    e,
+                    aiohttp.client_exceptions.ClientResponseError)
+                if is_response_error and e.status == 400:  # type: ignore # pylint: disable=no-member
+                    log.warning(
+                        'Ack error is unrecoverable, dropping the whole batch')
+                    for i in ack_ids:
+                        asyncio.ensure_future(
+                            subscriber_client.acknowledge(subscription,
+                                                          ack_ids=[i])
```
jonathan-johnston, line 113: I think this code makes sense and is a good tradeoff between batching for throughput and ensuring valid `ack_id`s get acked without being held back by one or more bad IDs, but does this constitute dropping the whole batch? Maybe I'm just nitpicking on wording but some IDs may be successfully acknowledged and others dropped, making it a "partial drop". I would personally be a bit confused by the warning message.

Again, functionally it makes sense but could the warning align with the behaviour? Something like `Ack error is unrecoverable, one or more messages may be dropped`? (and likewise for the nacker)

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -99,6 +100,20 @@ async def acker(subscription: str,
                 log.warning(
                     'Ack request failed, better luck next batch', exc_info=e)
                 metrics_client.increment('pubsub.acker.batch.failed')
+
+                is_response_error = isinstance(
+                    e,
+                    aiohttp.client_exceptions.ClientResponseError)
+                if is_response_error and e.status == 400:  # type: ignore # pylint: disable=no-member
```
jonathan-johnston, line 107: Any other HTTP error code relevant? Other `4xx` class errors maybe (although some are transient in GCP, like occasional 404s when network is bad)?

jonathan-johnston (approved):
Overall, my comments are not blocking (although would certainly be nice to address) so I'll just toss an approve on here.

pubsub/gcloud/aio/pubsub/subscriber.py
shaundialpad: I agree with the wording. We're not really dropping the batch, right? We're giving each ack id another chance (if I'm reading this correctly)

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -99,6 +100,20 @@ async def acker(subscription: str,
...
+                    log.warning(
+                        'Ack error is unrecoverable, dropping the whole batch')
+                    for i in ack_ids:
+                        asyncio.ensure_future(
```
shaundialpad, line 111: Perhaps we can `gather` here with `return_exceptions=True` and then subsequently remove only the poison'ed `ack_id`s based on the results?  Just a thought, maybe a bad idea :P

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -99,6 +100,20 @@ async def acker(subscription: str,
...
                     'Ack request failed, better luck next batch', exc_info=e)
                 metrics_client.increment('pubsub.acker.batch.failed')
+
+                is_response_error = isinstance(
+                    e,
+                    aiohttp.client_exceptions.ClientResponseError)
+                if is_response_error and e.status == 400:  # type: ignore # pylint: disable=no-member
+                    log.warning(
+                        'Ack error is unrecoverable, dropping the whole batch')
+                    for i in ack_ids:
```
shaundialpad, line 110: nit: s/i/ack_id

shaundialpad (commented):
Just tell the server to return the failed ack ids /s (egress joke)

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: My reasoning was that from `Acker` point of view it bears no responsibility for this batch from now on. Sure we'll make a final effort to maybe ack what can be acked, but we won't monitor the result, not will we catch or log any exceptions.

But suggested wording is more precise anyway, so sure why not

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: Invalid `ack_id` also throws 400, and I'm out of other ideas :D

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: well theoretically a batch can be as big as 2500 items, so we'd be waiting for 2500 requests to complete. That kinda contradicts the general idea that `Acker` should be an fast moving train that acks messages as fast as `ack_window` allows.

pubsub/gcloud/aio/pubsub/subscriber.py
jonathan-johnston: Great, 400 works 😛

TheKevJames, line 107: I think this would be cleaner by doing `except aiohttp.client_exceptions.ClientResponseError` as a fully different branch -- it would lead to duplicating the log warning and metrics, but I think that's OK -- we fundamentally want to react to the different exception type in a different way, so that's not the sort of duplication which I consider to be a bad thing.

Alternatively, you could wrap the exceptions at a different level, eg.
```python
try:
  try:
    await subscriber_client.acknowledge(..)
  except aiohttp.client_exceptions.ClientResponseError as e:
    if e.status == 400:
      # SNIP: your new code here
    raise
except Exception:
  log.warning(..)
  metrics_client.increment(..)
```

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -99,6 +100,20 @@ async def acker(subscription: str,
                 log.warning(
                     'Ack request failed, better luck next batch', exc_info=e)
                 metrics_client.increment('pubsub.acker.batch.failed')
+
+                is_response_error = isinstance(
+                    e,
+                    aiohttp.client_exceptions.ClientResponseError)
+                if is_response_error and e.status == 400:  # type: ignore # pylint: disable=no-member
+                    log.warning(
```
TheKevJames, line 108: This probably warrants an `error` level notification IMO -- this feels like the sort of thing which folks should be loudly told about.

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: I've been going a bit back and forth on this. I _think_ it might be nice to err on the side of gathering these; it's really no different than how we `await subscriber_client.acknowledge(..)` above -- if we were to just spawn that (and these acks) as unmanaged futures, we could easily "get ahead of ourselves" and spin up more threads than we can handle; I think that's what we've been seeing with the registration aredis stuff -- if we spin up an unmanageable number of coroutines, there's no guarantee the previously spawned (ie. and now completed) ones are returned to in time.

I think this could also be a nice way for giving us an even better error message here, eg. we could do something like:
```python
results = await asyncio.gather(
    *[asyncio.ensure_future(subscriber_client.acknowledge(subscription, ack_ids=[ack_id]))
      for ack_id in ack_ids],
    return_exceptions=True)
failures = sum(1 for result in results if isinstance(result, Exception))
log.error('Unrecoverable ack error: %d messages have been dropped', failures)
```

What do you think?

TheKevJames (changes requested):

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: My vision is that `Acker` prefers present over past. Any potential latency added here will negatively affect all messages in the pipeline, since they're all on timer to be acked. I don't see how logging the exact number of dropped messages can justify that?

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: Hmm, yes, that makes sense. It is kinda actionable meaning that it won't be triggered by random network blip, but rathe an actual bug or unfortunate but otherwise valid infrastructure even (like in our case). 👍

pubsub/gcloud/aio/pubsub/subscriber.py
shaundialpad: Can also wrap the gather in a timeout as a compromise?

amartsinovich-dialpad: well, in that case we can then gather `ack_ids` that failed to ack but for "retryable" reasons and return them into buffer to try next time... the complexity can grow but what's the real benefit?

TheKevJames: I guess my worry is more that we could theoretically spawn 2500 tasks here, which would all show up on the event loop and compete for time with the scheduler. That's going to introduce a whole bunch of potential latency _anyway_, eg. we might do so, then grab a new task off the queue and start work, then preempt that task 2500 times as we get ack responses back and never get around to completing that task before it times out anyway! It might be better, then, to purposely "pause and collect our thoughts" here rather than potentially have odd latency occur in odd places until all those tasks are dealt with.

To be clear, including the number of failures in the log message is just a nice bonus if we go with the above anyway rather than a good reason to switch in itself!

amartsinovich-dialpad: But why it might be better? Spawning 2500 tasks _and awaiting on them_ will consolidate latency and guarantee to include _server response_ latency, while fire-and-forget tasks will at least distribute latency between all tasks and worst case will be as bad as awaiting.

TheKevJames: > distribute latency between all tasks

I guess it depends if this ends up being too much latency or not. Is it better to have one of our Pubsub processing apps pause while it deals with this sort of spike in its backlog, or keep going at the potential expense of causing a bunch of upcoming tasks to retry? I suppose the implication here is that _probably_ all processors would be going through this at the same time...

Honestly, it seems really hard to say for me. Like, what if those 2500 tasks means the next ~10s (random number) worth of _new_ tasks take too long and thus fail and need a retry? Then we've added ~10s of latency to everything there -- contrast that with (potentially!) pausing to handle the 2500 up front in what would probably be <~10s of latency since we're doing less at a time.

But if we think we can absorb those 2500 tasks _without_ causing timeout issues in the next batch of messages, then you're 100% right with keeping things as you've written them

shaundialpad: To step back a second, the reason we're seeing this is that the subscription was deleted. The grpc clients simply considered that a fatal error and we would shutdown and restart, picking up the new version of the deploy. Should we maybe be doing something similar?

amartsinovich-dialpad: >Is it better to have one of our Pubsub processing apps pause while it deals with this sort of spike in its backlog, or keep going at the potential expense of causing a bunch of upcoming tasks to retry?
But it won't pause immediately, will it? Acker will stop yes, but all other tasks will still have to compete with those 2500 requests until all internal queues are full and back pressure make it all the way from acker to producers.
What is the worst case scenario? Assuming there's no free CPU resources.
1. (Await) Acker is blocked. In the meantime subscriber continues to work until all internal queues are filled. Then it sends 2500 requests, waits until the last one receives a response and then resumes work as usual.
2. (Fire and forget) Our really mean scheduler puts acker on halt, then lets internal queues to fill, then schedules 2500 requests, but then it has to give acker a turn.

We're really choosing between degrading latency for unknown random amount for `time to schedule 2500 requests + time to schedule receiving their responses` time window or add up to `time to schedule 2500 requests + receive response for the longest of them` of latency to unknown number of messages in a range from `[number of messages currently in the system]` to `[max number of messages in the system]`.

I don't know, maybe just drop all messages all together if we're not sure what's better?

amartsinovich-dialpad: @shaundialpad if the subscription is deleted, producer will definitely shutdown everything. We could shutdown here too, _but_ graceful shutdown will try to finish processing everything currently in work, and that wouldn't be able to accomplish...

TheKevJames: Ah, shite, I'd forgotten that awaiting here would only pause our _acking_ thread, not the actual subscriber&worker. Thanks for laying it out clearly there Alex, that's really helpful!! I think fire-and-forget is just fine.

TheKevJames (approved):

PR #302, opened 2021-02-24
chore(pubsub) Log ack/nack last call attempt in case of exception

amartsinovich-dialpad:
When we do a final sendout, invalid messages raise and thus flood sentry with non actionable errors

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -101,11 +101,19 @@ else:
                     log.error(
                         'Ack error is unrecoverable, '
                         'one or more messages may be dropped', exc_info=e)
+
+                    async def maybe_ack(ack_id: str) -> None:
+                        try:
+                            await subscriber_client.acknowledge(
+                                subscription,
+                                ack_ids=[ack_id])
+                        except Exception as e:
+                            log.warning('Ack failed for ack_id=%s',
+                                        ack_id,
+                                        exc_info=e)
+
                     for ack_id in ack_ids:
-                        asyncio.ensure_future(
-                            subscriber_client.acknowledge(subscription,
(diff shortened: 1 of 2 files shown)


pubsub/tests/unit/subscriber_test.py
```diff
@@ -643,7 +644,10 @@ async def f(*args, **kwargs):
         )
 
     @pytest.mark.asyncio
-    async def test_acker_batches_not_retried_on_400(subscriber_client):
+    async def test_acker_batches_not_retried_on_400(caplog,
```
shaundialpad, line 647: TIL `caplog` 👍

shaundialpad (approved):

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -101,11 +101,19 @@ async def acker(subscription: str,
                     log.error(
                         'Ack error is unrecoverable, '
                         'one or more messages may be dropped', exc_info=e)
+
+                    async def maybe_ack(ack_id: str) -> None:
```
TheKevJames, line 105: nit: since you've defined this function in a loop, the interpreter will have to spend cycles building and rebuilding this function definition. Always better to pull functions up as much as possible; either to near the top of the `acker` method (eg. outside of the loop), or even a top-level function entirely (ie. and pass in the subscriber_client).

TheKevJames (approved):
LGTM, but moving the method definition would be a nice perf improvement.

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: I agree with the point, but i'd prefer to keep it here, because logically this is a lambda and it makes most sense when read in this context.

PR #304, opened 2021-03-01
Shaun/pubsub/shutdown fixes

shaundialpad:
Noticed in one of our apps that there were a number of redeliveries on shutdown. Tracked this down to at least 2 causes:

1. We prematurely mark an ack task as done, thereby joining on the corresponding queue early and failing to ack messages.
2. The long running pull request we make to pubsub can return messages after producer cancellation (discovered this by noticing that there were some messages with `delivery_attempt=2` that we did not detect having `delivery_attempt=1`)

Latest test on staging in realtime-nlp resulted in no redelivered messages after manually killing pods.

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -288,14 +288,17 @@ else:
             while True:
                 new_messages = []
                 try:
-                    new_messages = await subscriber_client.pull(
-                        subscription=subscription,
-                        max_messages=max_messages,
-                        # it is important to have this value reasonably high
-                        # as long lived connections may be left hanging
-                        # on a server which will cause delay in message
-                        # delivery or even false deadlettering if it is enabled
-                        timeout=30)
+                    pull_task = asyncio.ensure_future(
+                        subscriber_client.pull(
+                            subscription=subscription,
+                            max_messages=max_messages,
+                            # it is important to have this value reasonably
+                            # high as long lived connections may be left
+                            # hanging on a server which will cause delay in
+                            # message delivery or even false deadlettering if
+                            # it is enabled
+                            timeout=30))
+                    new_messages = await asyncio.shield(pull_task)
                 except (asyncio.TimeoutError, KeyError):
                     continue
 
@@ -310,6 +313,15 @@ else:
                 await message_queue.join()
         except asyncio.CancelledError:
             log.info('Producer worker cancelled. Gracefully terminating...')
+
+            if not pull_task.done():
+                # Leaving the connection hanging can result in redelivered
+                # messages, so try to finish before shutting down
+                try:
+                    new_messages += await asyncio.wait_for(pull_task, 5)
+                except (asyncio.TimeoutError, KeyError):
+                    pass
+
             pulled_at = time.perf_counter()
             for m in new_messages:
                 await message_queue.put((m, pulled_at))
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -67,7 +67,6 @@ else:
                 message = await asyncio.wait_for(
                     queue.get(), timeout=time_budget)
                 result.append(message)
-                queue.task_done()
             except asyncio.TimeoutError:
                 break
             time_budget -= (time.perf_counter() - start)
@@ -82,7 +81,6 @@ else:
         while True:
             if not ack_ids:
                 ack_ids.append(await ack_queue.get())
-                ack_queue.task_done()
 
             ack_ids += await _budgeted_queue_get(ack_queue, ack_window)
 
@@ -93,9 +91,14 @@ else:
                     'acker is falling behind, dropping %d unacked messages',
                     len(ack_ids) - 2500)
                 ack_ids = ack_ids[-2500:]
+                for _ in range(len(ack_ids) - 2500):
+                    ack_queue.task_done()
+
             try:
                 await subscriber_client.acknowledge(subscription,
                                                     ack_ids=ack_ids)
+                for _ in ack_ids:
+                    ack_queue.task_done()
             except aiohttp.client_exceptions.ClientResponseError as e:
                 if e.status == 400:
                     log.error(
@@ -111,6 +114,8 @@ else:
                             log.warning('Ack failed for ack_id=%s',
                                         ack_id,
                                         exc_info=e)
+                        finally:
+                            ack_queue.task_done()
 
                     for ack_id in ack_ids:
                         asyncio.ensure_future(maybe_ack(ack_id))
@@ -121,6 +126,8 @@ else:
                 metrics_client.increment('pubsub.acker.batch.failed')
 
                 continue
+            except asyncio.CancelledError:  # pylint: disable=try-except-raise
+                raise
             except Exception as e:
                 log.warning(
                     'Ack request failed, better luck next batch', exc_info=e)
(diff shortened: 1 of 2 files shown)


pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -121,6 +126,8 @@ async def maybe_ack(ack_id: str) -> None:
                 metrics_client.increment('pubsub.acker.batch.failed')
 
                 continue
+            except asyncio.CancelledError:  # pylint: disable=try-except-raise
+                raise
```
amartsinovich-dialpad, line 130: Hmm, I think we might be better off implementing graceful shutdown flow in `acker` worker.

Otherwise with this changes `ack_window` will punish consumer for being too performant. Consider ack_queue of `length=10` and `ack_window=30`. No matter how fast producer is, acker will not ack more than 10 messages per 30 seconds.

Alternatively we can also hardcode ack queue length to 2500, but that will require us to check `ack_ids` list length after each message pulled from the queue. But that might not interfere nicely with "inherited" batch logic...

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -298,6 +313,15 @@ async def producer(
                 await message_queue.join()
         except asyncio.CancelledError:
             log.info('Producer worker cancelled. Gracefully terminating...')
+
+            if not pull_task.done():
+                # Leaving the connection hanging can result in redelivered
+                # messages, so try to finish before shutting down
+                try:
+                    new_messages += await asyncio.wait_for(pull_task, 5)
```
amartsinovich-dialpad, line 321: What happens when we wait for a shielded task? Isn't it the same as `await pull_task`?

amartsinovich-dialpad (changes requested):
Excellent catch!

pubsub/gcloud/aio/pubsub/subscriber.py
shaundialpad: Hmm not sure I understand your question :P Are you talking about the `wait_for` ? I enforced 5 seconds here because we're shutting down and we need time to do that. Or, are you talking about how we're awaiting a shielded task above, and then once cancelled, we await the unshielded task? If you try to await the shielded task again, the `CancelledError` will be raised.

pubsub/gcloud/aio/pubsub/subscriber.py
shaundialpad: >Hmm, I think we might be better off implementing graceful shutdown flow in acker worker.

I agree that right now, the shutdown logic for acker is wrapped up in the `consumer`, but I thought this change was the most straight forward extension of the current logic since we're already joining the acker queue during consumer shutdown. I suppose we could change this and wrap the main worker loop of acker in a separate coroutine and shield it from cancellation as well. Thoughts?

> Otherwise with this changes ack_window will punish consumer for being too performant. Consider ack_queue of length=10 and ack_window=30. No matter how fast producer is, acker will not ack more than 10 messages per 30 seconds.

How is consumer being punished any more with these changes than it was before? I just changed where we were signalling finished tasks, which should only affect when `join.()`ing on the ack queue, which is only during shutdown. Or are you saying this will negatively impact shutdown?

amartsinovich-dialpad: > How is consumer being punished any more with these changes than it was before? I just changed where we were signalling finished tasks, which should only affect when .join()ing on the ack queue, which is only during shutdown. Or are you saying this will negatively impact shutdown?

It will also be blocked on `queue.put(...)` if the queue size limit is reached? It seems quite unlikely with proper setup, but with `num_producers=1` and `max_messages_per_producer=6` 🌝  we will be doomed to ack no more than 6 messages per `ack_window`.

>I suppose we could change this and wrap the main worker loop of acker in a separate coroutine and shield it from cancellation as well. Thoughts?

I was thinking about just doing a final `await subscriber_client.acknowledge(subscription, ack_ids=sack_ids)` in `except asyncio.CancelledError` block, but shielding also sounds reasonable. It just a bit harder for me to understand without testing as you can see by my other questions :D

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: Ohh, I see now, we wait for unshielded version of the task, that's why `wait_for` is able to cancel it after 5 seconds! Thank you!

But wait... why does it help the issue? Whenever we close the connection on our side, there is a chance that pubsub will deliver a message to stale connection thus incrementing its `delivery_attempt`. Or do we just increase our odds here so our probability to terminate normally is 5/30?

shaundialpad: > why does it help the issue

That _is_ the question! I don't have all the details, but I've observed is that we do in fact hit this code path in staging (i.e.  the request is still running after being cancelled). As long as the response is returned within 5 seconds we will successfully recover the otherwise lost messages. Will this work all the time? Probably not. I am guessing it will be more successful under heavy load (i.e. shorter response times). But we shouldn't wait too long if we're going to be terminated soon anyhow. We could make this configurable?

pubsub/gcloud/aio/pubsub/subscriber.py
shaundialpad: > It will also be blocked on queue.put(...) if the queue size limit is reached?

`.put()` seems to be unaffected by `.task_done()`. i.e. something like

```python
q = asyncio.Queue(maxsize=1)
await q.put(1)
await q.get()
await q.put(1)
await q.get()
q.task_done()
q.task_done()
await q.join()
```

succeeds just fine. I think that is where the confusion lies? Or I could just be confused 😄

amartsinovich-dialpad: Oh, I see now. Good call, sorry for confusion. Yeah, I believe in that case you initial approach is the most straightforward

pubsub/gcloud/aio/pubsub/subscriber.py
amartsinovich-dialpad: Got it. I think making this configurable might be too much, subscriber already has quite a few knobs. I'm happy with how it is now!

pubsub/tests/unit/subscriber_test.py
```diff
@@ -686,6 +692,30 @@ async def f(*args, **kwargs):
...
+                'fake_subscription',
+                queue,
+                subscriber_client,
+                0.1,
+                MagicMock()
+            )
+        )
+        await queue.put('ack_id_1')
+        await asyncio.wait_for(queue.join(), 1.)
+        mock.assert_called_with('fake_subscription', ack_ids=['ack_id_1'])
```
amartsinovich-dialpad, line 716: Is this test conceptually different from basic positive case? It looks like what we want to test is that `queue.join()` will return when and only when the mock object is called.

shaundialpad: Yeah that's essentially the only difference. Could just throw that logic in the `test_acker_does_ack` test? Something like

```python
@pytest.mark.asyncio
async def test_acker_does_ack(subscriber_client):
    queue = asyncio.Queue()
    acker_task = asyncio.ensure_future(
        acker(
            'fake_subscription',
            queue,
            subscriber_client,
            0.0,
            MagicMock()
        )
    )
    await queue.put('ack_id')
    await queue.join()
    subscriber_client.acknowledge.assert_called_once_with(
        'fake_subscription', ack_ids=['ack_id'])
    acker_task.cancel()
```

shaundialpad: Ok done 👍

amartsinovich-dialpad (approved):
looks good!

PR #307, opened 2021-03-11
Add a function to get a streamed response

domZippilli:
Hi @TheKevJames & company,

I was taking a look at how I might get downloads from your library without downloading object contents into memory. I found some [documentation](https://docs.aiohttp.org/en/stable/client_quickstart.html#streaming-response-content) for aiohttp which recommended using the response.content attribute directly, instead of the response.read() method, for downloads of significant size. Since a lot of GCS objects are pretty big, I think this change is useful.

It doesn't modify the original `download` behavior, since it is a breaking change. Instead, this PR:

- adds a new method, `get_content`. This is just a clone of the download method, but with a `get_content` flag set when calling the private implementation
- changes the private implementation `_download` method to honor the `get_content` flag, returning response.content when that is set to True. When False (default), the original behavior is retained.
- changes the implementation of `download_to_filename` to now use the chunked writing from content, therefore using less memory while still being quite performant. A conservative buffer size of 4kB (typical small kernel page size) is used. This last change is not necessary to move this PR forward and I'm happy to back it out. It does illustrate how simple it is to use the `aiohttp.StreamReader` API, at least.

edit: I should add, I'm not a very opinionated programmer. If you'd prefer to just add a switch to the `download` method and go with that, I'm good. Might make it hard to discover, which is why I went with creating another method.

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -40,6 +40,7 @@ SCOPES = [
 ]
 
 MAX_CONTENT_LENGTH_SIMPLE_UPLOAD = 5 * 1024 * 1024  # 5 MB
+WRITE_BUFFER_SIZE = 4096
 
 
 STORAGE_EMULATOR_HOST = os.environ.get('STORAGE_EMULATOR_HOST')
@@ -216,8 +217,13 @@ class Storage:
     async def download_to_filename(self, bucket: str, object_name: str,
                                    filename: str, **kwargs: Any) -> None:
         with open(filename, 'wb+') as file_object:
-            file_object.write(await self.download(bucket, object_name,
-                                                  **kwargs))
+            content = await self.get_content(bucket, object_name,
+                                                  **kwargs)
+            while True:
+                chunk = await content.read(WRITE_BUFFER_SIZE)
+                if not chunk:
+                    break
+                file_object.write(chunk)
 
     async def download_metadata(self, bucket: str, object_name: str, *,
                                 headers: Optional[Dict[str, Any]] = None,
@@ -227,6 +233,14 @@ class Storage:
                                     timeout=timeout, session=session)
         metadata: Dict[str, Any] = json.loads(data.decode())
         return metadata
+
+    async def get_content(self, bucket: str, object_name: str, *,
+                       headers: Optional[Dict[str, Any]] = None,
+                       timeout: int = 10,
+                       session: Optional[Session] = None) -> bytes:
+        return await self._download(bucket, object_name, headers=headers,
+                                    timeout=timeout, params={'alt': 'media'},
+                                    session=session, get_content=True)
 
     async def list_objects(self, bucket: str, *,
                            params: Optional[Dict[str, str]] = None,
@@ -368,7 +382,8 @@ class Storage:
                         params: Optional[Dict[str, str]] = None,
                         headers: Optional[Dict[str, str]] = None,
                         timeout: int = 10,
-                        session: Optional[Session] = None) -> bytes:
+                        session: Optional[Session] = None,
+                        get_content: bool = False) -> bytes:
         # https://cloud.google.com/storage/docs/json_api/#encoding
         encoded_object_name = quote(object_name, safe='')
         url = f'{API_ROOT}/{bucket}/o/{encoded_object_name}'
@@ -378,6 +393,10 @@ class Storage:
         s = AioSession(session) if session else self.session
         response = await s.get(url, headers=headers, params=params or {},
                                timeout=timeout)
+
+        if get_content:
+            return response.content
+
         # N.B. the GCS API sometimes returns 'application/octet-stream' when a
         # string was uploaded. To avoid potential weirdness, always return a
         # bytes object.
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -249,6 +249,23 @@ class Storage:
                               timeout: int = 10,
                               session: Optional[Session] = None
                               ) -> StreamResponse:
+        """Download a GCS object in a buffered stream. 
+
+        Args:
+            bucket (str): The bucket from which to dowonload.
+            object_name (str): The object within the bucket to download.
+            headers (Optional[Dict[str, Any]], optional): Custom header values
+                for the request, such as range. Defaults to None.
+            timeout (int, optional): Timeout, in seconds, for the request. Note
+                that with this function, this is the time to the beginning of
+                the response data (TTFB). Defaults to 10.
+            session (Optional[Session], optional): A specific session to
+                (re)use. Defaults to None.
+
+        Returns:
+            StreamResponse: A object encapsulating the stream, similar to
+            io.BufferedIOBase, but it only supports the read() function.
+        """
         return await self._download_stream(bucket, object_name,
                                            headers=headers, timeout=timeout,
                                            params={'alt': 'media'},
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -27,12 +27,13 @@ if BUILD_GCLOUD_REST:
     from time import sleep
     from requests import HTTPError as ResponseError
     from requests import Session
-    StreamResponse = Generator[bytes]
+    StreamResponse = Generator[bytes, None, None]
 else:
     from asyncio import sleep  # type: ignore[misc]
     from aiohttp import ClientResponseError as ResponseError  # type: ignore[no-redef]  # pylint: disable=line-too-long
     from aiohttp import ClientSession as Session  # type: ignore[no-redef]
-    StreamResponse = aiohttp.StreamReader
+    from aiohttp import StreamReader
+    StreamResponse = StreamReader  # type: ignore[misc]
 
 
 API_ROOT = 'https://www.googleapis.com/storage/v1/b'
@@ -191,7 +192,7 @@ class Storage:
                      params: Optional[Dict[str, str]] = None,
                      headers: Optional[Dict[str, str]] = None,
                      session: Optional[Session] = None) -> str:
-        # https://cloud.google.com/storage/docs/json_api/#encoding
+        # https://cloud.google.com/storage/docs/request-endpoints#encoding
         encoded_object_name = quote(object_name, safe='')
         url = f'{API_ROOT}/{bucket}/o/{encoded_object_name}'
         headers = headers or {}
@@ -232,13 +233,14 @@ class Storage:
         return metadata
 
     async def download_stream(self, bucket: str, object_name: str, *,
-                       headers: Optional[Dict[str, Any]] = None,
-                       timeout: int = 10,
-                       session: Optional[Session] = None
-                       ) -> StreamResponse:
-        return await self._download_stream(bucket, object_name, headers=headers,
-                                    timeout=timeout, params={'alt': 'media'},
-                                    session=session, get_content=True)
+                              headers: Optional[Dict[str, Any]] = None,
+                              timeout: int = 10,
+                              session: Optional[Session] = None
+                              ) -> StreamResponse:
+        return await self._download_stream(bucket, object_name,
+                                           headers=headers, timeout=timeout,
+                                           params={'alt': 'media'},
+                                           session=session)
 
     async def list_objects(self, bucket: str, *,
                            params: Optional[Dict[str, str]] = None,
@@ -381,7 +383,7 @@ class Storage:
                         headers: Optional[Dict[str, str]] = None,
                         timeout: int = 10,
                         session: Optional[Session] = None) -> bytes:
-        # https://cloud.google.com/storage/docs/json_api/#encoding
+        # https://cloud.google.com/storage/docs/request-endpoints#encoding
         encoded_object_name = quote(object_name, safe='')
         url = f'{API_ROOT}/{bucket}/o/{encoded_object_name}'
         headers = headers or {}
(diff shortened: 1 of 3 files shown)


TheKevJames:
@domZippilli thanks for the contribution! It looks like we've actually got two ongoing approaches in PR for trying to solve this problem or similar: @jackton1 is working on #309 which would make use of [aiofiles](https://pypi.org/project/aiofiles/) which it seems may do the same thing as this PR? Note that PR got closed prematurely due to some Github shenanigans, but should be back up soon; you can find the original diff [here](https://github.com/talkiq/gcloud-aio/compare/master...f87e024e1b1f745deac17cd6fd0f75ec9dc09185).

I haven't had a chance yet to take a look at which approach makes the most sense here -- or even if both should be done together? Once @jackton1's PR is back up, we can see how the two will work together -- if you have any thoughts on how best to work through this, I'm all ears. It seems to me that the two approaches might supplement each other, eg. if `aiofiles` can accept a response stream from your new method here, we can use your code for downloading in chunks and the `aiofiles` code for non-blocking file IO.

As for this PR in particular, I like how you have it structured; `get_content()` being a new function makes sense to me -- that said, I think renaming it to `download_stream()` might be a bit clearer for the public interface. I think your changes to `download_to_filename()` will have the most conflicts with the `aiofiles` approach -- lbacking those out for now to keep the merge simple might be for the best, if it seems the `aiofiles` approach is worth integrating here.

domZippilli:
Yeah, a little difficult for me to compare #309 and #307 right now, I'll be interested to see them when the commit history is sorted out on #309.

I'm new to the Python asyncio world, but looking at aiofiles, it seems like these two changes are different and complementary. #307 moves to using a write-from-stream approach instead of download-to-memory and then write-from-memory. That reduces memory footprint and likely speeds things up. I think #309 and aiofiles speeds things up a bit further by making the writes async. 

It's probably unnatural to do this without also writing from the stream, so I wouldn't be surprised if there are conflicting changes that do the same thing. I also would probably think slightly differently about buffers, for example, if I were doing async tasks for each write to disk. The combination of the two probably really helps with slower disks, like some NFS and spinning disks that have high seek time.

Anyway, I can remove the download to file changes and do a quick rename so that this is easier to de-conflict with the other. I'll follow it so that I can know when the git issues are straightened out and I can more easily take a look at it.

jackton1:
> Yeah, a little difficult for me to compare #309 and #307 right now, I'll be interested to see them when the commit history is sorted out on #309.
> 
> I'm new to the Python asyncio world, but looking at aiofiles, it seems like these two changes are different and complementary. #307 moves to using a write-from-stream approach instead of download-to-memory and then write-from-memory. That reduces memory footprint and likely speeds things up. I think #309 and aiofiles speeds things up a bit further by making the writes async.
> 
> It's probably unnatural to do this without also writing from the stream, so I wouldn't be surprised if there are conflicting changes that do the same thing. I also would probably think slightly differently about buffers, for example, if I were doing async tasks for each write to disk. The combination of the two probably really helps with slower disks, like some NFS and spinning disks that have high seek time.
> 
> Anyway, I can remove the download to file changes and do a quick rename so that this is easier to de-conflict with the other. I'll follow it so that I can know when the git issues are straightened out and I can more easily take a look at it.

@domZippilli Should be resolved now #315

domZippilli:
Hi all.

@TheKevJames to make this easier for you to review/deconflict/hopefully merge, I removed revision of the `download_to_file` method. I also renamed `get_content` to `download_stream`. I think now this can be merged independent of #315 (née #309). I did take a look and it looks like the changes in #315 to `download_to_file` are just formatting, but no matter -- the smaller the PR the better.

@jackton1 yeah, doing something similar for uploads makes sense to me, although the upload implementation here is a lot bigger and I was going to look at it in another PR. It _may_ be really easy... I think all we'd have to do is modify the `_preprocess_data` method to accept the output of `aiofile.async_open`. It _might_ already, but I don't think so because aiofiles seems to use [its own ABC](https://github.com/mosquito/aiofile/blob/61c1919f846d44094f341d8d643a9a0aaf4883a5/aiofile/utils.py#L158) and not `io.IOBase`. But probably, we could just add a case to `_preprocess_data` for `FileWrapperIOBase` and it should "just work." That would provide methods like `.tell` and `.seek` which the existing implementations require.

But... yeah I'm unsure about it enough and it seems like enough work that I'd rather just punt it to another PR.

storage/gcloud/aio/storage/storage.py
```diff
@@ -368,7 +376,8 @@ async def _download(self, bucket: str, object_name: str, *,
                         params: Optional[Dict[str, str]] = None,
                         headers: Optional[Dict[str, str]] = None,
                         timeout: int = 10,
-                        session: Optional[Session] = None) -> bytes:
+                        session: Optional[Session] = None,
+                        get_content: bool = False) -> bytes:
```
TheKevJames, line 380: Looks like this breaks the expected return value here, since this method now returns a `aiohttp.StreamReader` when `get_content` is true and we're running in async mode -- I think this is a bit confusing since in the current method implementation, `response.content` refers to `bytes` since that code path is only hit by gcloud *rest* and the `requests` library uses that attribute to grab the raw non-stream bytes from the object. In this new code path, `response.content` refers to the attribute in `aiohttp`, which is completely different.

It may be easiest to just copy this out into a `_download_stream` method, for clarity sake -- see below comment for needing to support `gcloud-rest-storage` in this code anyway.

storage/gcloud/aio/storage/storage.py
```diff
@@ -378,6 +387,10 @@ async def _download(self, bucket: str, object_name: str, *,
         s = AioSession(session) if session else self.session
         response = await s.get(url, headers=headers, params=params or {},
                                timeout=timeout)
+
+        if get_content:
+            return response.content
```
TheKevJames, line 392: This will only return a stream for `gcloud-aio` -- for gcloud-rest-storage, we'll need to make use of [iter_content()](https://docs.python-requests.org/en/latest/api/#requests.Response.iter_content), I think, and attach `stream=True` to the request.

storage/gcloud/aio/storage/storage.py
TheKevJames: Note that for a case like this, you may find it useful to make use of the `if BUILD_GCLOUD_REST` above for type hinting, for example you can do something like:
```python 
if BUILD_GCLOUD_REST:
    StreamResponse = io.BytesIO
else:
    StreamResponse = aiohttp.StreamReader

# SNIP
    async def _download_stream(..) -> StreamResponse
```

TheKevJames (changes requested):
@domZippilli thanks for splitting this apart a bit, I agree with keeping this a bit more granular. Would definitely love to accept a follow-up PR to tackle the upload side of things!

Looks like a couple outstanding linting issues described in the `pre-commit` output -- I added a comment to help describe how you might be able to tackle the typing issue.

storage/gcloud/aio/storage/storage.py
domZippilli: > and we're running in async mode

Ohhhh, yeah I forgot this library has a synchronous mode as well. I was a little thrown off by [this try/except](https://github.com/talkiq/gcloud-aio/blob/a3bcfff89645b0717acc137414201c4550ab0c63/storage/gcloud/aio/storage/storage.py#L400), it looked like it was already breaking the return type. But I guess the Attribute or Type errors arise when you try to await a synchronous read, and therefore you know response.content will be `bytes` in that situation. Is that more or less it?

Anyway, _happy_ to break it out into another method. I almost did, and only didn't because I misunderstood that try/catch. I might add a comment to that try catch for posterity, if that's OK. :)

TheKevJames: More or less: the `await`s all get stripped out of the `gcloud-rest` code, but the `.read` attribute on `requests.Response` is (IIRC): not a method on old versions of requests (thus, raises a TypeError) and non-existent at all on newer versions (thus, raises an AttributeError).

Adding a comment mentioning what the except case is for sounds absolutely lovely, thank you!

storage/gcloud/aio/storage/storage.py
```diff
@@ -401,6 +401,25 @@ async def _download(self, bucket: str, object_name: str, *,
...
+        headers.update(await self._headers())
+
+        s = AioSession(session) if session else self.session
+        response = await s.get(url, headers=headers, params=params or {},
+                               timeout=timeout, stream=True)
+
+        if BUILD_GCLOUD_REST:
+            return response.iter_content
+        return response.content
+
```
domZippilli, line 422: @TheKevJames WDYT about this implementation?

As you suggested, I added to the conditional types at the top, creating a `StreamResponse` type variable. After wading through `requests.Response` for a while, I concur -- `iter_content` is the way to go, and I think the most similar thing to return is the `iter_content` generator itself. There are no type hints in requests, so I just went with the Generator type there.

domZippilli:
Will fix merge conflicts after we get to a implementation lgtm

storage/gcloud/aio/storage/storage.py
```diff
@@ -388,6 +401,25 @@ async def _download(self, bucket: str, object_name: str, *,
 
         return data
 
+    async def _download_stream(self, bucket: str, object_name: str, *,
+                        params: Optional[Dict[str, str]] = None,
+                        headers: Optional[Dict[str, str]] = None,
+                        timeout: int = 10,
+                        session: Optional[Session] = None) -> StreamResponse:
+        # https://cloud.google.com/storage/docs/json_api/#encoding
```
jackton1, line 409: Probably meant to use 
```
https://cloud.google.com/storage/docs/request-endpoints#encoding
```

domZippilli: yeah IDK this is just copy-pasted from `_download`, I included that comment but never followed the link. It does look broken, so I can fix in both places while I'm under the hood

TheKevJames: Huh, yeah, looks like they changed the link on us. @jackton1 's link is indeed the one we meant to point to.

storage/gcloud/aio/storage/storage.py
TheKevJames: I think this conditional works fine -- often we would write this as eg.
```python
try:
    return response.iter_content
except AttributeError:
    return response.content
```
but I think this `if` check is clearer anyway. LGTM as-is :+1:

storage/gcloud/aio/storage/storage.py
```diff
@@ -228,6 +231,15 @@ async def download_metadata(self, bucket: str, object_name: str, *,
         metadata: Dict[str, Any] = json.loads(data.decode())
         return metadata
 
+    async def download_stream(self, bucket: str, object_name: str, *,
+                       headers: Optional[Dict[str, Any]] = None,
```
TheKevJames, line 235: To make the linter happy, please align the args, eg. such that `headers` starts at the same place as `self`.

TheKevJames (approved):
Overall this looks great to me! Happy to merge and release this once you fix up the build errors -- or please let me know if you need a hand with that.

jackton1:
@TheKevJames @domZippilli Any updates on this ??

domZippilli:
I fixed a couple small things from Kevin's last review just now. It should be ready to roll -- there's a small merge conflict to fix in the conditional imports.

I was going to take some more time to test this out, but it is actually a pretty small change and a new codepath, so I think the risk of bugs is not too awful if people are waiting on it.

domZippilli:
@jackton1 working on it... what's here now is pretty stable if you want to work from it. At least, it has tests now. I think we will end up with either the current commit or reverting a couple and going with the very slightly different design. See below...

@TheKevJames Ok, I started running pre-commit and locally and fixing things up. Also wrote integration tests to run with nox. In the course of doing this, I figured out another way to do this.

I think the approach I had before of a `StreamResponse` that was one type or another based on the `BUILD_GCLOUD_REST` variable wasn't going to work. It gave mypy all kinds of fits, and I couldn't get it to be happy. Basically I think I was running into [trying to use a type alias like a type variable](https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases). I needed an alias for annotations, but need control flow to set its type like a variable. Like a... dynamic alias. But I don't think we have that.

So, my next attempt was to just make a custom class instead of an alias or type variable. You can see [that commit](https://github.com/talkiq/gcloud-aio/pull/307/commits/a1f6f2c2a7b79468d5f3b112aeaa8acea048309f) here. This worked, and the test code looked pretty natural. The thing I didn't like about this is, the class doesn't extend an `io` base class, even though it looks like it's trying to. Mostly, that's because the `read` method is `async`. I didn't find a mirror image of, say, `io.BufferedIOBase` over in `asyncio`, though maybe I missed it. Perhaps if there is one, I can use an import alias to make this a dynamic subclass. My head hurts though!

So, my final attempt was to just ditch all this custom type/class business and simply have `download_stream` return a `Callable[[int],bytes]`. That lets me return either `iter_content` or `read` for `requests` and `aiohttp` sessions, respectively. So, it's quite a bit simpler code in `storage.py`. The only downside is that the interface is a bit weird, since I'm returning a Callable now, and not something with Callable members. You can see this in the test:

```python
        with io.BytesIO(b'') as downloaded_data:
            read = await storage.download_stream(bucket_name, res['name'])
            while True:
                chunk = await read(4096)
```

Specifically, `read` being a callable all by itself, and not the underlying response. 

Personally, I think I prefer having the custom class. I think it makes the end-user's code more idiomatic for buffered reading. But either of these approaches seem to satisfy the invariant that library users can write code that either `awaits` or doesn't, and it works either way. 

I think we're close now. WDYT?

storage/gcloud/aio/storage/storage.py
```diff
@@ -228,6 +229,36 @@ async def download_metadata(self, bucket: str, object_name: str, *,
...
+                              headers: Optional[Dict[str, Any]] = None,
+                              timeout: int = 10,
+                              session: Optional[Session] = None
+                              ) -> Callable[[int], bytes]:
+        """Download a GCS object in a buffered stream. Note the return value
+        is a reader function, not a stream. This is to support indirection
+        between async and synchronous HTTP libraries.
+
+        Args:
+            bucket (str): The bucket from which to dowonload.
```
domZippilli, line 242: s/dowonload/download

TheKevJames:
@domZippilli thanks for considering so many options!

I agree with you -- the custom class code shown [here](https://github.com/talkiq/gcloud-aio/pull/307/commits/a1f6f2c2a7b79468d5f3b112aeaa8acea048309f) looks fantastic IMO, and I'd be happy to go with that approach. Please additionally expose the `StreamReader` class via the [storage entrypoint file](https://github.com/talkiq/gcloud-aio/blob/master/storage/gcloud/aio/storage/__init__.py#L10), but otherwise all looks well there.

domZippilli:
OK, so I think we're good...

cc: @jackton1

jackton1:
Perfect

TheKevJames (approved):
Fantastic! Thanks again -- I'll get this all deployed presently.

TheKevJames:
@domZippilli @jackton1 your changes have now been released as part of [v5.7.0](https://github.com/talkiq/gcloud-aio/releases/tag/storage-5.7.0)

Note that I did have to make a small bugfix -- please check out [the additional changes](https://github.com/talkiq/gcloud-aio/compare/285db19afbd4396c701ff4a474daad360464dbf1..storage-5.7.0) before updating to make sure things are still working as you'd expect :)

PR #310, opened 2021-04-01
feat(storage): add ability to make patch request to stored objects (metadata)

eddiedialpad:
Summary:
• adds a method `Storage.patch_metadata()` that accepts an arbitrary dictionary to patch stored object metadata

Related:
• https://switchcomm.atlassian.net/browse/VAE-2317
• ✅ https://github.com/talkiq/gcloud-aio/pull/311 — move Session.patch() to gcloud-(aio|rest)-auth

Testing:
• ✅ tested locally w `pip install -e ./storage`

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -304,36 +304,32 @@ class Storage:
 
     async def _patch_metadata(
             self, bucket: str, object_name: str, metadata: Dict[str, Any],
-            *, params: Optional[Dict[str, str]] = None,
-            session: Optional[Session] = None,
+            params: Dict[str, str], headers: Dict[str, str],
+            *, session: Optional[Session] = None,
             timeout: int = 10) -> Dict[str, Any]:
+        # https://cloud.google.com/storage/docs/json_api/v1/objects/patch
         url = f'{API_ROOT}/{bucket}/o/{object_name}'
+        headers.update(await self._headers())
+        headers['Content-Type'] = 'application/json'
+
         s = AioSession(session) if session else self.session
-        headers = await self._headers()
-        headers.update({'Content-Type': 'application/json'})
-        resp = await s.patch(
-            url, data=json.dumps(metadata).encode('utf-8'),
-            headers=headers, params=params, timeout=timeout)
+        resp = await s.patch(url, data=json.dumps(metadata).encode('utf-8'),
+                             headers=headers, params=params, timeout=timeout)
         data: Dict[str, Any] = await resp.json(content_type=None)
         return data
 
     async def set_temporary_hold(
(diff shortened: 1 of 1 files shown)


storage/gcloud/aio/storage/storage.py
```diff
@@ -302,6 +302,38 @@ async def upload_from_filename(self, bucket: str, object_name: str,
             return await self.upload(bucket, object_name, file_object,
                                      **kwargs)
 
+    async def set_temporary_hold(
+            self, bucket: str, object_name: str, *, hold: bool = True,
+            params: Optional[Dict[str, str]] = None,
+            session: Optional[Session] = None, timeout: int = 10
+    ) -> Dict[str, Any]:
+
```
TheKevJames, line 310: Please toss a comment linking to https://cloud.google.com/storage/docs/json_api/v1/objects/patch here

storage/gcloud/aio/storage/storage.py
```diff
@@ -302,6 +302,38 @@ async def upload_from_filename(self, bucket: str, object_name: str,
             return await self.upload(bucket, object_name, file_object,
                                      **kwargs)
 
+    async def set_temporary_hold(
```
TheKevJames, line 305: It looks like there are a few dozen different aspects of an object which may be patched. What do you think of the relative merits of gradually exposing more and more of these `set_*()` methods vs, say, exposing `.patch(metadata: Dict[str, Any])` directly?

eddiedialpad: I like it — in that case, I can use say `def _set_metadata()` to indicate a private function for now, until it's clear that that the set of fields to patch has grown enough; alternatively, as this grows, we can also use an enum, named arguments etc., as a way to control the vocabulary of fields to those that are actually supported by each Google resource (starting here at Storage).

storage/gcloud/aio/storage/storage.py
eddiedialpad: (added to the private method instead.)

storage/gcloud/aio/storage/storage.py
TheKevJames: Hmm I guess I'm not sure we need to allowlist the available fields at all? Why not allow users to pass in `metadata={'temporaryHold': True}` directly? Then we won't need to manually build every new feature we want to be able to patch.

I think you can just do `s/_patch_metadata/patch_metadata/` with your current code (and maybe keep some of these defaults such as `headers = None`) and be in a good place -- does being overly proscriptive here win us something I'm missing?

eddiedialpad: No, that's fine — I misunderstood the direction you were going in the original comment.

TheKevJames (approved):

harsh-dialpad (approved):

PR #311, opened 2021-04-01
feat(auth): add http PATCH to BaseSession, AioSession, and SyncSession

eddiedialpad:
Add `patch` to `BaseSession`, `AioSession` and `SyncSession`.
• this is a pre-requisite to individually modifying metadata for Google Storage objects, but def belongs in the auth module, as it's probably general enough that we'll figure a need for it elsewhere.

Unblocks:
• https://github.com/talkiq/gcloud-aio/pull/310

### DIFF
--- a/auth/gcloud/aio/auth/session.py
+++ b/auth/gcloud/aio/auth/session.py
@@ -47,6 +47,12 @@ class BaseSession:
         pass
 
     @abstractmethod
+    async def patch(self, url: str, headers: Dict[str, str],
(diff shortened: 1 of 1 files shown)


eddiedialpad:
Just checked the circleci scripts — doesn't look like there's anything special beyond the usual string manipulations to get from aio -> rest — so I think we're done.

TheKevJames (approved):
LGTM!

PR #319, opened 2021-04-19
feat(bigquery): accept request params

shaundialpad:
Currently it is impossible for the user to paginate query results. Allowing one to pass through arbitrary query parameters will enable them to do so (see the full list here: https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults)

Example use case:

```python
    params = {'maxResults': 100}
    results = await job.get_query_results(params=params)
    rows = results['rows']
    page_token = results.get('pageToken')
 
    while page_token:
        params['pageToken'] = page_token
        results = await job.get_query_results(params=params)
        page_token = results.get('pageToken')
        rows.extend(results['rows'])
```

Optionally, we can add additional behavior to do the pagination for them, either aggregating over all results, or perhaps more ideally allowing direct iteration over a specified `maxResults`.

### DIFF
--- a/bigquery/gcloud/aio/bigquery/bigquery.py
+++ b/bigquery/gcloud/aio/bigquery/bigquery.py
@@ -146,7 +146,9 @@ class Job:
 
     # https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults
     async def get_query_results(self, session: Optional[Session] = None,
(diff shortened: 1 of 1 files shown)


amartsinovich-dialpad (approved):
looks good

harsh-dialpad (approved):
LGTM

PR #323, opened 2021-04-22
fix(storage): fixup blob signing

TheKevJames:
The previous implementation via the iamcredentials API seems to no longer
be working. Signing the blob via standard PEM->PKCS methods seems to work
just fine, fortunately!

Basically, rather than offloading the signing work, we just do this 
locally, so this should be a performance improvement as well. Since our 
previous implementation only worked for SERVICE_ACCOUNT tokens *anyway*
(due to a mistake -- theoretically, we could have pulled the email from the
metadata server for GCE_METADATA tokens...), this shouldn't break backwards
compatibility.

Note that at the same time as the iamcredentials API breaking, it seems the
canonical uri for blob signs also was updated (the bucket name is now
included in the host instead of as part of the path). I suspect the two are
related.

### DIFF
--- a/storage/gcloud/aio/storage/blob.py
+++ b/storage/gcloud/aio/storage/blob.py
@@ -1,19 +1,21 @@
 import binascii
 import collections
 import datetime
+import enum
 import hashlib
 import io
 from typing import Any
 from typing import Dict
 from typing import Optional
 from typing import TYPE_CHECKING
-from typing import Union
 from urllib.parse import quote
 
+import rsa
 from gcloud.aio.auth import BUILD_GCLOUD_REST  # pylint: disable=no-name-in-module
-from gcloud.aio.auth import decode  # pylint: disable=no-name-in-module
-from gcloud.aio.auth import IamClient  # pylint: disable=no-name-in-module
 from gcloud.aio.auth import Token  # pylint: disable=no-name-in-module
+from pyasn1.codec.der import decoder
+from pyasn1_modules import pem
+from pyasn1_modules.rfc5208 import PrivateKeyInfo
 
 # Selectively load libraries based on the package
 if BUILD_GCLOUD_REST:
@@ -26,6 +28,36 @@ if TYPE_CHECKING:
 
 
 HOST = 'storage.googleapis.com'
+
+PKCS1_MARKER = ('-----BEGIN RSA PRIVATE KEY-----',
+                '-----END RSA PRIVATE KEY-----')
+PKCS8_MARKER = ('-----BEGIN PRIVATE KEY-----',
+                '-----END PRIVATE KEY-----')
+PKCS8_SPEC = PrivateKeyInfo()
+
+
+class PemKind(enum.Enum):
+    """
+    Tracks the response of ``pem.readPemBlocksFromFile(key, *args)``>
+
+    Note that the specified method returns ``(marker_id, key_bytes)``, where
+    ``marker_id`` is the integer index of the matching ``arg`` (or -1 if no
+    match was found.
+
+    For example::
+
+        (marker_id, _) = pem.readPemBlocksFromFile(key, PKCS1_MARKER,
+                                                   PCKS8_MARKER)
+        if marker_id == -1:
+            # "key" did not match either type or was invalid
+        if marker_id == 0:
+            # "key" matched the zeroth provided marker arg, eg. PKCS1_MARKER
+        if marker_id == 1:
+            # "key" matched the zeroth provided marker arg, eg. PKCS8_MARKER
+    """
+    INVALID = -1
+    PKCS1 = 0
+    PKCS8 = 1
 
 
 class Blob:
@@ -56,12 +88,7 @@ class Blob:
     async def get_signed_url(  # pylint: disable=too-many-locals
             self, expiration: int, headers: Optional[Dict[str, str]] = None,
             query_params: Optional[Dict[str, Any]] = None,
-            http_method: str = 'GET',
-            iam_client: Optional[IamClient] = None,
-            service_account_email: Optional[str] = None,
-            service_file: Optional[Union[str, io.IOBase]] = None,
-            token: Optional[Token] = None,
-            session: Optional[Session] = None) -> str:
+            http_method: str = 'GET', token: Optional[Token] = None) -> str:
         """
         Create a temporary access URL for Storage Blob accessible by anyone
         with the link.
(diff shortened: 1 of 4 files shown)


TheKevJames:
Huh, Github/Circle seem to have fucked up the check statuses -- tests are passing on the latest commit.

storage/tests/integration/signed_url_test.py
```diff
@@ -26,9 +25,7 @@ async def test_gcs_signed_url(bucket_name, creds, data):
         bucket = Bucket(storage, bucket_name)
         blob = await bucket.get_blob(object_name, session=session)
 
-        iam_client = IamClient(service_file=creds, session=session)
-
-        signed_url = await blob.get_signed_url(60, iam_client=iam_client)
+        signed_url = await blob.get_signed_url(60)
```
eddiedialpad, line 28: errp ... didja wanna mock any of this?

TheKevJames: Nope, not for an integration test! Generally, you want to mock things in unit tests to ensure you test only your discrete component, but integration tests should not have anything about your app itself mocked. Since this method no longer requires a connection to GCP, there's nothing to mock there!

Also, this repo somewhat misnamed our integration test folders -- these tend to actually be smoke tests where we're literally hitting the GCP endpoints, since for a library whose only job is to access an API, that middle ground of integration testing where the API is mocked doesn't really test you all that much.

storage/gcloud/aio/storage/blob.py
```diff
@@ -73,23 +75,33 @@ async def get_signed_url(  # pylint: disable=too-many-locals
...
         datetime_now = datetime.datetime.utcnow()
         request_timestamp = datetime_now.strftime('%Y%m%dT%H%M%SZ')
         datestamp = datetime_now.strftime('%Y%m%d')
 
-        service_account_email = (service_account_email
-                                 or iam_client.service_account_email)
+        # TODO: support for GCE_METADATA and AUTHORIZED_USER tokens. It should
+        # be possible via API, but seems to not be working for us in some
+        # cases.
+        #
```
caseydialpad, line 88: How worthwhile would it be to reach out to GCP support?

storage/gcloud/aio/storage/blob.py
```diff
@@ -120,12 +132,30 @@ async def get_signed_url(  # pylint: disable=too-many-locals
...
-            session=session)
 
-        signature = binascii.hexlify(
-            decode(signed_resp['signedBlob'])).decode()
+        marker_id, key_bytes = pem.readPemBlocksFromFile(
+            io.StringIO(key), PKCS1_MARKER, PKCS8_MARKER)
+        # Given `pem.readPemBlocksFromFile(key, *args)`, returns `marker_id`
+        # as the index of the args which matched or -1 for an invalid/unmatched
+        # key.
+        if marker_id == -1:
```
caseydialpad, line 141: nit: What does `-1` (and `1` below) correspond with? If it means `PKCS1_MARKER_INVALID` or something something specific, a username might provide insight into what `marker_id` indicates.

storage/gcloud/aio/storage/blob.py
```diff
@@ -120,12 +132,30 @@ async def get_signed_url(  # pylint: disable=too-many-locals
...
+            key_info, remaining = decoder.decode(key_bytes,
+                                                 asn1Spec=PKCS8_SPEC)
+            if remaining != b'':
+                raise ValueError('could not read PKCS8 key: found extra bytes',
+                                 remaining)
+
+            private_key_info = key_info.getComponentByName('privateKey')
+            key_bytes = private_key_info.asOctets()
+
+        key = rsa.key.PrivateKey.load_pkcs1(key_bytes, format='DER')
```
caseydialpad, line 155: nit: The `key` variable was used above on line 93. Since this is related to cryptography and the same variable stores two different types of data, a different variable name may help make the code more clear.

caseydialpad (approved):
Overall looks good. Just some minor comments.

storage/gcloud/aio/storage/blob.py
```diff
@@ -73,23 +75,33 @@ async def get_signed_url(  # pylint: disable=too-many-locals
...
-                                 or iam_client.service_account_email)
+        # TODO: support for GCE_METADATA and AUTHORIZED_USER tokens. It should
+        # be possible via API, but seems to not be working for us in some
+        # cases.
+        #
+        # https://cloud.google.com/iam/docs/reference/credentials/rest/v1/
+        # projects.serviceAccounts/signBlob
+        token = token or self.bucket.storage.token
+        client_email = token.service_data.get('client_email')
+        key = token.service_data.get('private_key')
```
eddiedialpad, line 93: Oh cool — does this mean that the current service account is sufficient to perform signing, as long as it was granted permission to sign?

eddiedialpad (approved):
Approved.
• yay! I can't wait to use this in slicey slice!

storage/gcloud/aio/storage/blob.py
TheKevJames: Was the comment the line above this unclear? How can I improve it?

Basically, the function I'm calling works very roughly like:
```python
def readPemBlocksFromFile(key, *args):
    marker_id = -1
    for i, arg in args:
        if matches_type(key, arg):
            marker_id = i
            break
    return marker_id
```

storage/gcloud/aio/storage/blob.py
TheKevJames: I've opened the dialog with our TAM and will improve this function over time as per their response :+1:

storage/gcloud/aio/storage/blob.py
TheKevJames: Yup! The docs say your token needs the `urlSign` permissions and the associated `SCOPES` granted, but in practice I was able to get this to work just fine with a token with only `storage.read` and the standard scopes associated with `glcoud-aio-storage`'s token. I left the `token` argument in this method to let folks pass in a specific urlsign token if need be, but it seems unnecessary... that said, rather avoid blocking anyone if I can help it!

storage/gcloud/aio/storage/blob.py
caseydialpad: I somehow missed the comment. That is clear. I take back my `nit`.

I still think there is value with having a clear variable name (such as `marker_id == NO_MATCH` or `INVALID_KEY`). But the comment is very clear.

fbalboaDialpad:
Hi all! Reviving this old PR for some more discussion. This change broke our blob signing flow, because we are still using GCE_METADATA tokens, which don't have a private key, so we still need to use the IAM `signBlob` API to produce signed URLs. This method works fine for us when using an old version of this lib.

According to Google docs (https://cloud.google.com/storage/docs/xml-api/overview#endpoints), if using the XML API then both
```
BUCKET_NAME.storage.googleapis.com/OBJECT_NAME
storage.googleapis.com/BUCKET_NAME/OBJECT_NAME
```
are valid. And when using the JSON API then only the second option (https://cloud.google.com/storage/docs/json_api/v1).
I'm not sure which use case was breaking for you, but ideally we should:
1. Re-add support for using the IAM API to sign blobs, so we can unpin this dependency in our repo.
2. Use the `storage.googleapis.com/BUCKET_NAME/OBJECT_NAME` format, as this should be compatible with both JSON and XML APIs.
3. Keep the local signing implementation and use it when available, as it's a perf upgrade that at least some users can use.

I'll be working on a PR to do this, but I'll need some help with testing out your use cases that where previously failing, and also with info about how to try this out by myself (ie. generate a version with my changes and test it in our repo)

PR #324, opened 2021-04-23
feat(pubsub): Supporting context management protocol in SubscriberClient

caseydialpad:
I noticed there was no context manager protocol implemented in `SubscriberClient`. I might misunderstand the class, but this will allow the use of `async with` without relying on the following idiom:

```python
async with aiohttp.ClientSession() as session:
    subscriber_client = SubscriberClient(session=session)
```
Instead, we will be able to do:

```python
async with SubscriberClient() as client:
    ....
```

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -183,3 +183,12 @@ class SubscriberClient:
                            timeout=timeout)
         result: Dict[str, Any] = await resp.json()
         return result
+
+    async def close(self) -> None:
+        await self.session.close()
+
+    async def __aenter__(self) -> 'SubscriberClient':
+        return self
+
+    async def __aexit__(self, *args: Any) -> None:
+        await self.close()


TheKevJames (approved):
Ah, great catch, not sure why this one got missed.

jonathan-johnston:
@caseydialpad @TheKevJames One interesting point to note here is that if you create the `SubscriberClient` with an already created `session` object (i.e. it was constructed outside of the `SubscriberClient` constructor) then it will be closed. Is this a behaviour that is expected? The general rule is to not clean up objects that you were given, only the ones that you create yourself. For example, the programmer may be sharing this `session` object with other objects that need access to the session.

caseydialpad:
@jonathan-johnston 

That is a great point. I added this because `PublisherClient` supports the exact same functionality. I do think you have found a bug.

I would probably argue the correct place to correct this behaviour would be in the `AioSession` class. Actually, that might mean correcting this in the `BaseSession` class. Check if `session = None`, and set an internal flag `session_provided`. 

In `AioSession`, when `close` is called, check if `session_provided` is set. If so, then do nothing, otherwise `close`. This would fix the bug in `PublisherClient` too then.

I also notice there is `SyncSession` which is a subclass of `BaseSession` used for the REST clients. Both `await` and `async` must be removed during some intermediate processing because `SyncSession` has `async` definitions and a note saying they are there only for static analysis. It's possible we might be able to define the `close` once in the base class that correctly handles the bug you've brought up.

Naturally, we could just patch this bug in `PublisherClient/SubscriberClient`, but to me this seems like a bug in the `BaseSession` class.

TheKevJames:
I agree with @caseydialpad -- if we could solve this in `*Session`, that seems to be the right place.

Note that we've traditionally avoided this since it can be very difficult to determine for sure who "owns" the session and thus who should clean it up -- bear in mind that there are various points in the application life-cycle where we or the user might create a session; now that we have the base session classes, though, I think we can be sure those cases are limit to places where [this](https://github.com/talkiq/gcloud-aio/blob/92ca8ca6afad2d1f29b647dae032e7f7694b91ac/auth/gcloud/aio/auth/session.py#L110) or [this](https://github.com/talkiq/gcloud-aio/blob/92ca8ca6afad2d1f29b647dae032e7f7694b91ac/auth/gcloud/aio/auth/session.py#L186) method is called while `self._session` is false.

If we can be absolutely sure we're closing any session we create and not managing any session the user is expected to manage, that sounds ideal to me!

caseydialpad:
@jonathan-johnston,

I've corrected the bug you discovered with #367 .

PR #338, opened 2021-05-05
chore(pubsub): add backoff to pull

shaundialpad:
Adds backoff directly to the pull method of the `SubscriberClient` (done here to be consistent with other parts of `gcloud-aio`,  e.g. https://github.com/talkiq/gcloud-aio/blob/e410a9cbaa3304634ce1283195ea9b8b725c5e46/auth/gcloud/aio/auth/token.py#L208

Also used the generic `Exception` solely for consistency. Since we'll shutdown soon anyhow in the event of a non transient error, I think this is okay.

Currently, a single failed `pull` request (e.g. 502 or other transient errors) will shut down the subscriber. Let's give it a chance to succeed.

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -9,6 +9,7 @@ from typing import List
 from typing import Optional
 from typing import Union
 
+import backoff
 from gcloud.aio.auth import AioSession
 from gcloud.aio.auth import BUILD_GCLOUD_REST  # pylint: disable=no-name-in-module
 from gcloud.aio.auth import Token
@@ -93,6 +94,7 @@ class SubscriberClient:
         await s.delete(url, headers=headers, timeout=timeout)
 
     # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/pull
+    @backoff.on_exception(backoff.expo, Exception, max_tries=5)  # type: ignore
     async def pull(self, subscription: str, max_messages: int,
                    *, session: Optional[Session] = None,
                    timeout: Optional[int] = 30
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -94,7 +94,8 @@ class SubscriberClient:
         await s.delete(url, headers=headers, timeout=timeout)
 
     # https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/pull
-    @backoff.on_exception(backoff.expo, Exception, max_tries=5)  # type: ignore
+    @backoff.on_exception(
+        backoff.constant, Exception, interval=1, max_tries=5)  # type: ignore
     async def pull(self, subscription: str, max_messages: int,
                    *, session: Optional[Session] = None,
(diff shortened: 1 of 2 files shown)


amartsinovich-dialpad (changes requested):
I believe exponential backoff can grow up to `2 ** 5 = 32s`? That might affect performance quite a bit if a consumer  goes to sleep for half a minute with k8s controller not knowing something is wrong. Let's maybe use a faster strategy?

shaundialpad:
> I believe exponential backoff can grow up to `2 ** 5 = 32s`? That might affect performance quite a bit if a consumer goes to sleep for half a minute with k8s controller not knowing something is wrong. Let's maybe use a faster strategy?

Hmm It seems to be doing what we want. Do you have any suggestions? Maybe simply enforcing `max_time` to 10 seconds or so?

amartsinovich-dialpad:
@shaundialpad How about `constant` wait strategy? This way we can retry 5 times in 5 seconds + response latency. We don't particularly care about putting too much stress on google servers, so expo backoff won't do much, only increase time to recovery.

TheKevJames (changes requested):
I don't think this is the right place for this:

We have the backoff handling in the gcloud-aio-auth lib explicitly because Google has a line in its docs mentioning that all calls to that endpoint need to have the specified backoff policy.

In this case, though, the exact policy is incredibly dependent on the use-case, as evidenced by your discussion in PR. We can not and should not define a single strategy which all possible consumers of this library agree with; instead, this should be up to the application to specify. Instead, applications who want to implement backoff should re-wrap this method with their preferred arguments.

Even in our own systems I'm sure we'd have multiple different values we'd want (eg. realtime vs postcall process is almost certainly going to want to tradeoff speed for resilience!).

shaundialpad:
> We have the backoff handling in the gcloud-aio-auth lib explicitly because Google has a line in its docs mentioning that all calls to that endpoint need to have the specified backoff policy.

Good to know. On a related note, maybe we should have that commented?

> In this case, though, the exact policy is incredibly dependent on the use-case, as evidenced by your discussion in PR. We can not and should not define a single strategy which all possible consumers of this library agree with; instead, this should be up to the application to specify. Instead, applications who want to implement backoff should re-wrap this method with their preferred arguments.

Mmm yes I agree. Actually, the fact that the user passes in `SubscriberClient` did not occur to me. So would the expected approach that one should take look something like

```
class SubscriberClientWithBackoff(SubscriberClient):

    @backoff.on_exception(backoff.expo, Exception, max_time=10)
    async def pull(self, *args, **kwargs):
        return await super().pull(*args, **kwargs)
```

?

TheKevJames:
> Good to know. On a related note, maybe we should have that commented?

Yeah, good call. I'll dig up the relevant doc and toss it into a comment :+1:

> `SubscriberClientWithBackoff`

Yeah, that makes a lot of sense to me as you have it! I suppose this is another thing worth documenting to help out future folks who hit this.

> And as a followup, is complete shutdown of `subscribe` after a single failed pull as default behavior the correct approach?

That's harder to say. This sounds like another case where I think users would need to configure it -- I could imagine use-cases for failingfast or for retrying up to some timeout or...

shaundialpad:
Closing in favour of having the user implement their own handling and configuration.

PR #343, opened 2021-05-14
Have Query from_repr support the 'kind' format generated by to_repr

todddialpad:
Have Query.from_repr accept a list with a single dict in it as the format for the 'kind' key, the same as is produced by Query.to_repr.


TheKevJames (approved):
Looks great, thanks!

PR #344, opened 2021-05-15
fix(storage): make the storage upload functions accept custom metadata

gagphil1:


gagphil1:
With this fix, it is now possible to add a field  metadata={'key1": data1, "key2", data2, _etc_} to the metadata argument in upload function.
This way, we can create new custom metadata while saving a blob. No need to modify the download functions, they are already able to receive the custom metadata.

storage/gcloud/aio/storage/storage.py
```diff
@@ -490,11 +490,15 @@ async def _upload_multipart(self, url: str, object_name: str,
...
-                    for k, v in metadata.items()}
-        metadata['name'] = object_name
+        metadict = (metadata or {}).copy()
+        custom_metadata = metadict.pop('metadata', None)
+        metadata_content = {self._format_metadata_key(k): v
+                            for k, v in metadict.items()}
+
+        metadata_content['name'] = object_name
+        if custom_metadata:
+            metadata_content['metadata'] = custom_metadata
```
TheKevJames, line 501: Could you explain what this code change is meant to accomplish? It seems you're pulling `metadata` out of the dict to avoid running it through `_format_metadata_key()` and re-adding it later, but `_format_metadata_key('metadata') == 'metadata'` so I'm a bit confused as to why you need to do this?

storage/gcloud/aio/storage/storage.py
```diff
@@ -540,8 +544,13 @@ async def _initiate_upload(self, url: str, object_name: str,
         params['uploadType'] = 'resumable'
 
         metadict = (metadata or {}).copy()
-        metadict.update({'name': object_name})
-        metadata_ = json.dumps(metadict)
+        custom_metadata = metadict.pop('metadata', None)
```
TheKevJames, line 547: Same question, looks like this block could be copied from the above:
```python
metadict= {self._format_metadata_key(k): v
            for k, v in metadata.items()}
metadict['name'] = object_name
metadata_ = json.dumps(metadict)
```

TheKevJames (changes requested):
Thanks for the contribution @gagphil1! Really appreciate you adding support for custom metadata -- to be honest, I thought we'd already supported it. Added a couple questions for clarifying what's up here.

storage/gcloud/aio/storage/storage.py
gagphil1: First off all, thanks for this library.

And you're right, this is absolutely not necessary. This is an artefact of my experimentations, and I will remove it.

storage/gcloud/aio/storage/storage.py
gagphil1: Absolutely... I'll correct that.

storage/gcloud/aio/storage/storage.py
```diff
@@ -504,7 +504,7 @@ async def _upload_multipart(self, url: str, object_name: str,
 
         parts = [
             (metadata_headers, json.dumps(metadata).encode('utf-8')),
-            ({'Content-Type': headers['Content-Type']}, bytes_body),
+            ({'content-type': headers['Content-Type']}, bytes_body),
```
TheKevJames, line 507: Is the casing change here necessary, or just a dev artifact? I believe we're using Title-Cased headers throughout this lib, so if Google is doing weird things with them, I might need to look into updating this more generally.

TheKevJames (commented):
Just a question re the header casing change, otherwise the major code changes & tests look great!

storage/gcloud/aio/storage/storage.py
gagphil1: In fact the case is not necessary.

gagphil1: I just pushed a new commit where I "stringnify" every metadata.metadata fields. This seems necessary if there are list or dict fields (for simple types the conversion was automatic).

gagphil1:
I just pushed a new commit where I "stringnify" every metadata.metadata fields. This seems necessary if there are list or dict fields (for simple types the conversion was automatic).

TheKevJames (approved):
Awesome, and thanks for catching the stringification issue as well! It'll be nice to support  more metadata formats than just strings.

TheKevJames:
Released as [gcloud-aio-storage v6.1.0](https://github.com/talkiq/gcloud-aio/releases/tag/storage-6.1.0)!

PR #350, opened 2021-05-20
feat(datastore) Support creating a mutation from Value objects

### LINKED ISSUE #348: Create entity with excluded from indexes properties
Hi, team!

Please suggest what is a correct way to create entity with some properties excluded from indexes? I use this way but got error `NotImplementedError: <class 'gcloud.aio.datastore.value.Value'> is not a supported value type`.

```python
properties = {
    'active': True,
    'data' : Value(data, exclude_from_indexes=True),
}
key = Key(settings.GCLOUD_PROJECT_ID, [PathElement('MyEntity')])
await client.insert(key, properties)
```

Thank you!

KevinGDialpad:
Fixes #348

### DIFF
--- a/datastore/gcloud/aio/datastore/datastore.py
+++ b/datastore/gcloud/aio/datastore/datastore.py
@@ -120,11 +120,15 @@ class Datastore:
         if operation == Operation.DELETE:
             return {operation.value: key.to_repr()}
 
+        mutation_properties = {}
+        for k, v in (properties or {}).items():
+            value = v if isinstance(v, cls.value_kind) else cls.value_kind(v)
+            mutation_properties[k] = value.to_repr()
+
         return {
             operation.value: {
                 'key': key.to_repr(),
-                'properties': {k: cls.value_kind(v).to_repr()
-                               for k, v in (properties or {}).items()},
+                'properties': mutation_properties,
             }
         }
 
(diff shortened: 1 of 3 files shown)


datastore/tests/unit/datastore_test.py
```diff
@@ -1,5 +1,23 @@
-import gcloud.aio.datastore.datastore as datastore  # pylint: disable=unused-import
+import pytest
+from gcloud.aio.datastore import Datastore
+from gcloud.aio.datastore import Key
+from gcloud.aio.datastore import Operation
+from gcloud.aio.datastore import PathElement
+from gcloud.aio.datastore import Value
 
 
-def test_importable():
```
KevinGDialpad, line 4: I figured we were past the point of testing that the module imports correctly

datastore/gcloud/aio/datastore/datastore.py
```diff
@@ -120,11 +120,15 @@ def make_mutation(
         if operation == Operation.DELETE:
             return {operation.value: key.to_repr()}
 
+        mutation_properties = {}
+        for k, v in (properties or {}).items():
+            value = v if isinstance(v, cls.value_kind) else cls.value_kind(v)
+            mutation_properties[k] = value.to_repr()
+
         return {
```
KevinGDialpad, line 128: Eventually I'd like to deprecate this method and have the logic inside the Mutation object. But I figured that should be a separate change.

TheKevJames (approved):
Awesome, thanks @KevinGDialpad!

PR #356, opened 2021-06-08
Do a buffered read when uploading local file instead of a full read.

conet:
When uploading large files the current implementation uses a lot of memory without good reason, there is no need to read the complete file when upload can work with the file stream itself.


TheKevJames (approved):
You're absolutely correct! Good catch.

PR #361, opened 2021-06-23
feat(storage): Add data to copy endpoint.

langered:
Since gcloud supports a data body in its rewrite method, the copy
funcion can expose this to its public endpoint. This also enables the
enduser to set other metadata for the copied object.

https://github.com/talkiq/gcloud-aio/issues/362

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -148,7 +148,7 @@ class Storage:
 
     async def copy(self, bucket: str, object_name: str,
                    destination_bucket: str, *, new_name: Optional[str] = None,
-                   data: Optional[str] = None,
+                   metadata: Optional[str] = None,
                    params: Optional[Dict[str, str]] = None,
                    headers: Optional[Dict[str, str]] = None, timeout: int = 10,
                    session: Optional[Session] = None) -> Dict[str, Any]:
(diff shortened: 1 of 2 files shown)


storage/gcloud/aio/storage/storage.py
```diff
@@ -148,6 +148,7 @@ def get_bucket(self, bucket_name: str) -> Bucket:
 
     async def copy(self, bucket: str, object_name: str,
                    destination_bucket: str, *, new_name: Optional[str] = None,
+                   data: Optional[str] = None,
```
TheKevJames, line 151: What do you think about calling this `metadata` instead? `data` is fairly broad and it looks like [this endpoint only supports object metadata in the body](https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite) anyway.

TheKevJames (commented):
Hi @langered, Thanks for the contribution! I think it's fine to expose the entire body here, no worries about constraining that.

Overall implementation LGTM, just a question on naming.

storage/gcloud/aio/storage/storage.py
langered: Hey @TheKevJames I agree with you that object metadata is being should be set with this parameter. Therefore, I changed the naming to `metadata`. The original intention was to expose the `data` parameter from the post request, see here: https://github.com/langered/gcloud-aio/blob/908603ba96812d7fe88e7416d66a55ddb77a492d/storage/gcloud/aio/storage/storage.py#L198-L199

Also, I see the risk of confusion, because you still have to pass the metadata as a attribute inside the metadata object like: `{metadata: {...}}` as described here: https://cloud.google.com/storage/docs/json_api/v1/objects#resource

TheKevJames (approved):
LGTM, thanks! I'll get this released shortly :)

PR #367, opened 2021-07-06
bug(auth): Prevent closing shared sessions from AIO/REST Sessions

caseydialpad:
Related to https://github.com/talkiq/gcloud-aio/pull/324.

This change resolves the context manager bug detailed in the linked PR.


TheKevJames (approved):

TheKevJames (commented):
This LGTM, but note that the tests are currently only looking at the gcloud-aio flow; please make sure they compile gcloud-rest tests as well! See other places where we import and run differently depending on the `BUILD_GCLOUD_REST` value.

PR #368, opened 2021-07-07
feat(typing): expose type hints as per pep 561

jonathan-johnston:
We've been writing type hints but not even exposing them! (Relates to https://github.com/talkiq/gcloud-aio/issues/212)

When we run `mypy` via `pre-commit` we use `--ignore-missing-imports`, which suppresses this kind of error message:
```
error: Skipping analyzing "voiceai.gcloud.aio.pubsub": found module but no type hints or library stubs
```

In cases where no type hints are found, the assumed return value of any function in the library is `Any`. Since we are enforcing type hinting on our team we might as well use them and let our code use their type hints. In order to expose these type hints to the users of the package you have to follow PEP 561 (Distributing and Packaging Type Information):

> Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing.

https://www.python.org/dev/peps/pep-0561/#packaging-type-information

The official `mypy` documentation supports this as well as the solution to missing type hints:

> Mypy will not try inferring the types of any 3rd party libraries you have installed unless they either have declared themselves to be PEP 561 compliant stub package (e.g. with a py.typed file) or have registered themselves on typeshed, the repository of types for the standard library and some 3rd party libraries.

https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-type-hints-for-third-party-library

For packages built with `pip` this means including the `py.typed` file in `setup.py` but `poetry` automatically includes the file in the distribution without needing to declare it in the `pyproject.toml`. For `gcloud-aio` this means including `py.typed` in the `MANIFEST.in` file because we set `include_package_data=True` which delegates package data inclusion/exclusion lists to `MANIFEST.in`.

Bonus blog: https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/

### DIFF
--- a/auth/MANIFEST.in
+++ b/auth/MANIFEST.in
@@ -1,4 +1,4 @@
 include README.rst LICENSE requirements.txt
-recursive-include gcloud *.py
+recursive-include gcloud *.py py.typed
 recursive-include tests *.py
 global-exclude *.pyc __pycache__
--- a/bigquery/MANIFEST.in
+++ b/bigquery/MANIFEST.in
@@ -1,4 +1,4 @@
 include README.rst LICENSE requirements.txt
-recursive-include gcloud *.py
+recursive-include gcloud *.py py.typed
 recursive-include tests *.py
 global-exclude *.pyc __pycache__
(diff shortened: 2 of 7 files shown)


TheKevJames (approved):
Note to self: need to switch this over to poetry at some point.

Great find on all the py.typed stuff!

b0g3r:
Hi @TheKevJames! Could you please publish new versions of `gcloud-aio-*` libraries with py.typed?

PR #370, opened 2021-07-22
feat(pubsub): no ack deadline cache refresh by default

shaundialpad:
Each read of the deadline counts towards the GCP Admin Operations
quota, which has a hard limit of 6000 ops/min at the time of writing.
This can causes unexpected quota overages if there are many services
making use of the subscriber.

This change disables the ack deadline cache refresh behavior by
default, and users must now opt-in to enable it.

- [x] manually tested
- [x] update tests

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -31,10 +31,10 @@ else:
 
     class AckDeadlineCache:
         def __init__(self, subscriber_client: SubscriberClient,
-                     subscription: str, cache_timout: int):
+                     subscription: str, cache_timeout: float):
             self.subscriber_client = subscriber_client
             self.subscription = subscription
-            self.cache_timeout = cache_timout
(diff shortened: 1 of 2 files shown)


amartsinovich-dialpad (commented):
How will this work? Will it cache the value once and retain it forever or rather disable the short circuiting completely?

shaundialpad:
> How will this work? Will it cache the value once and retain it forever or rather disable the short circuiting completely?

Ah I just noticed this will never set the ack deadline, which we need regardless. I'll tweak it.

shaundialpad:
> How will this work? Will it cache the value once and retain it forever or rather disable the short circuiting completely?

@amartsinovich-dialpad It will attempt `refresh` until we successfully pull a value for `ack_deadline`. After this, we will never find `cache_outdated` to be truthy since `cache_timeout` is infinite.

amartsinovich-dialpad (approved):

TheKevJames (approved):

PR #380, opened 2021-09-17
feat(bigquery): implemented missing APIs for bigquery.

BoweiYuanDialpad:
Implemented following bigquery APIs:
1. insert job
2. synchronous query
3. create table via API call instead of DDL
4. patch table
5. list tables in a dataset
6. lsit datasets in a project
7. describe a dataset
8. insert a new dataset


bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -503,3 +527,64 @@ async def __aenter__(self) -> 'Table':
...
+
+        url = (f'{API_ROOT}/projects/{project}/datasets/'
+               f'{self.dataset_name}')
+        return await self._get_url(url, session, timeout, params=params)
+
+    # https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/insert
+    async def insert(self, dataset: Dict[str, Any],
+                     session: Optional[Session] = None,
+                     timeout: int = 60) -> Dict[str, Any]:
+        """List datasets in current project."""
```
TheKevJames, line 586: Please ensure docstrings match the function they're describing or omit them entirely -- accurate docs are fantastic, but missing docs would be better than misleading ones!

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -503,3 +527,64 @@ async def __aenter__(self) -> 'Table':
 
     async def __aexit__(self, *args: Any) -> None:
         await self.close()
+
+
+class Dataset(BigqueryBase):
+    def __init__(self, dataset_name: Optional[str] = None,
```
TheKevJames, line 533: It'd be great to make this `dataset_name` field required to avoid needing to have runtime validation in the bulk of it's methods; anywhere we can have schema failures occur in static analysis makes the library that much easier to debug.

What do you think about moving `list_datasets()` to `BigqueryBase` to accomplish this? Since `Dataset` inherits from it anyway, that means users would be able to do `Dataset().list_datasets()` if they need, but could also use the base client itself in cases where they don't have even a single named dataset to reference upfront.

Another option could be to expose this as a method: eg. something like `list_datasets(client: BigqueryBase, ...)`. I think I prefer having it be a member function, but wanted to present this as a valid option as well. Thoughts?

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -503,3 +527,64 @@ async def __aenter__(self) -> 'Table':
 
     async def __aexit__(self, *args: Any) -> None:
         await self.close()
+
+
+class Dataset(BigqueryBase):
```
TheKevJames, line 532: This file is starting to become quite the monolith with all these new classes. Would you mind splitting it apart? We should probably pull out the relevant classes to, at least, `job.py`, `table.py`, and `dataset.py` to help keep this understandable.

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -342,20 +345,46 @@ def _make_query_body(
...
+
+    # https://cloud.google.com/bigquery/docs/reference/rest/v2/tables/patch
+    async def patch(self, table: Dict[str, Any],
+                    session: Optional[Session] = None,
+                    timeout: int = 60) -> Dict[str, Any]:
+        """Patch an existing table specified by tableId from the dataset."""
+        project = await self.project()
+        url = (f'{API_ROOT}/projects/{project}/datasets/'
+               f'{self.dataset_name}/tables/{self.table_name}')
+        if 'tableReference' not in table:
```
TheKevJames, line 374: Interesting -- is there any usecase for the `tableReference` *not* matching the data you're providing here? As far as I can tell, the API docs say the `tableReference` must match the IDs listed in the URL. Since users can't modify the URL, we shouldn't let them modify the `tableReference` either (if I read that right!) since that's just a footgun waiting to happen. What do you think about overwriting `table['tableReference']` unconditionally?

bigquery/gcloud/aio/bigquery/bigquery.py
TheKevJames: Looks like this new class hasn't been exposed to users yet! Please check out the `__init__.py` file, where we define the public interface for users. You'll want to ensure this new `Dataset` class is included there!

TheKevJames (changes requested):
Thanks for cleaning up the duplicated post/get logic throughout this file -- great to make things a bit easier to read, though in the future I'd ask that you pull large refactors like that into a second PR; having both sets of changes in a cingle commit makes it much harder to reason about which lines were changed for which reason.

bigquery/gcloud/aio/bigquery/bigquery.py
BoweiYuanDialpad: Will do, thanks for spotting this copy-and-paste error :)

bigquery/gcloud/aio/bigquery/bigquery.py
BoweiYuanDialpad: One downside of this approach is that `list_datasets()` method will also be available in `Table` and `Job` class, which works but does not really make sense. 

I am not that against the idea of making `dataset_name` option. After all in class `Job` we have this pattern as well, a `Job` object does not need to work with a specific Job, similarly a `Dataset`object does not necessarily need to operate on a specific dataset either.

bigquery/gcloud/aio/bigquery/bigquery.py
BoweiYuanDialpad: will do.

bigquery/gcloud/aio/bigquery/bigquery.py
BoweiYuanDialpad: Overwriting `tableReference` makes sense, will do.

BoweiYuanDialpad: Done.

bigquery/gcloud/aio/bigquery/bigquery.py
BoweiYuanDialpad: Done.

bigquery/gcloud/aio/bigquery/bigquery.py
BoweiYuanDialpad: done.

bigquery/gcloud/aio/bigquery/bigquery.py
TheKevJames: Ah, good point -- I completely missed that, since it looks like we don't have explicit validation in the `Job` class as you've done here. Cool, looks good as is!

bigquery/gcloud/aio/bigquery/table.py
```diff
@@ -0,0 +1,315 @@
...
+
+    # https://cloud.google.com/bigquery/docs/reference/rest/v2/tables/insert
+    async def create(self, table: Dict[str, Any],
+                     session: Optional[Session] = None,
+                     timeout: int = 60) -> Dict[str, Any]:
+        """Create the table specified by tableId from the dataset."""
+        project = await self.project()
+        url = (f'{API_ROOT}/projects/{project}/datasets/'
+               f'{self.dataset_name}/tables')
+        if 'tableReference' not in table:
```
TheKevJames, line 143: Looks like this would be another place to overwrite `tableReference` unconditionally.

bigquery/gcloud/aio/bigquery/table.py
```diff
@@ -0,0 +1,315 @@
...
+                   session=self.session, token=self.token)
+
+    async def close(self) -> None:
+        await self.session.close()
+
+    async def __aenter__(self) -> 'Table':
+        return self
+
+    async def __aexit__(self, *args: Any) -> None:
+        await self.close()
```
TheKevJames, line 315: Please move these three methods to the new base class -- we want them to be defined in the same place we're describing what `self.session` is and will want them inherited into the new Dataset class.

TheKevJames (commented):
Thanks for splitting things apart, made it really clear to find the missing context manager inheritance. Once we've got that solved, this lgtm!

bigquery/gcloud/aio/bigquery/table.py
BoweiYuanDialpad: done.

bigquery/gcloud/aio/bigquery/table.py
BoweiYuanDialpad: done.

TheKevJames (approved):
Awesome, this looks great!

PR #395, opened 2021-11-04
fix(storage): Object name for patch_metadata

kubasaw:
Object name encoding fix: for any other methods, object name encoding is perfomed, but not for the Storage().patch_metadata() method


TheKevJames (approved):
Thank you, good catch -- I'd completely missed this!

Issue #382, opened 2021-09-23
Using `upload_from_filename` fails when used with async temporary file


dreki:
We're using `aiofiles.tempfile.NamedTemporaryFile` to stream data into a file, and then `storage.upload_from_filename` to upload that file into Google Cloud Storage. We experience an exception when, it appears, `gcloud-aio-storage` fails to recognize the file it opens as a valid reader.

Reproduction code:
```
async with httpx.AsyncClient() as http:
    async with http.stream('GET', url_preview.image) as response:
        response: httpx.Response
        async with aiofiles.tempfile.NamedTemporaryFile() as temp_file:
            temp_file: aiofiles.tempfile.NamedTemporaryFile
            [await temp_file.write(b) async for b in response.aiter_bytes()]
            result = await storage.upload_from_filename(
                settings.get('url_preview_image_bucket'),
                (f'{settings.get("url_preview_image_folder")}'
                    f'/{filename}'),
                filename=temp_file.name
            )
```

Using:
```
aiofiles==0.7.0
gcloud-aio-storage==6.1.1
```

Here's a stack trace:
![image](https://user-images.githubusercontent.com/563345/134539980-51bc7243-12bf-428a-9388-6450d6a42f9f.png)

dreki:
One result of this is that all files have to be fully loaded into memory before uploading.

rafiahmad16:
Having same issue

TheKevJames:
Solved in 899c8daf3586f5cae1f866b5fafba1972dae74da -- will get this fixed in the next release (v6.3.0, likely).

PR #397, opened 2021-11-09
fix(storage): fix update_from_filename method

lukaszKielar:
This is a fix for the [issue](https://github.com/talkiq/gcloud-aio/issues/382). Let me know if it's fine, so I can add some tests.

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -358,10 +358,11 @@ class Storage:
     async def upload_from_filename(self, bucket: str, object_name: str,
                                    filename: str,
                                    **kwargs: Any) -> Dict[str, Any]:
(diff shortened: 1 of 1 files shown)


TheKevJames:
Hi @lukaszKielar -- sorry, I did not see this PR (looks like I'm not currently getting notifications for PRs in the draft state... will fix that...). I've just fixed the issue [here](https://github.com/talkiq/gcloud-aio/commit/899c8daf3586f5cae1f866b5fafba1972dae74da) by reverting the original change which caused the regression. Will get that fix released as soon as I can.

lukaszKielar:
Hi @TheKevJames, thanks for response and fix! I saw it and I'm wondering what's the reason behind using read method on an async reader instead of using context manager? It could be problematic when dealing with large files on machines with limited memory.

PR #398, opened 2021-11-09
add timeout parameter to blob download

sanderland:
Ended up almost breaking prod when switching to this due to the low default timeout, which is hidden when using Blob.download. Here is a PR to add this parameter.

Not sure what the 'no name storage' in your test suite is about.


TheKevJames (approved):
Thanks for the contribution -- makes sense to expose that value here.

PR #401, opened 2021-11-11
fix(pubsub): gracefully handle worker cancellations

TheKevJames:
Graceful worker cancellations during subscriber shutdown should not be cause for alarm or logged at the error level. Additionally, it can be useful to track these separately, as cancellations and true worker errors are different issues.


shaundialpad (approved):
Makes sense to me!

PR #411, opened 2021-12-18
chore(auth): resolve usage of deprecated abc.abstractproperty

jackton1:
See: https://docs.python.org/3/library/abc.html#abc.abstractproperty

### DIFF
--- a/auth/gcloud/aio/auth/session.py
+++ b/auth/gcloud/aio/auth/session.py
@@ -2,7 +2,6 @@ import logging
 import threading
 from abc import ABCMeta
 from abc import abstractmethod
-from abc import abstractproperty
 from typing import Any
 from typing import Dict
 from typing import IO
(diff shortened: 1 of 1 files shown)


TheKevJames (commented):
Hi @jackton1 thank you for the contribution! Unfortunately, this project still supports python 2.7 for the `gcloud-rest-*` builds and (as far as I can see in the Python documentation) this approach would not work for those versions of Python.

If we want to use this syntax in `gcloud-aio`, we'd need to modify the py3to2 logic (ie. here: https://github.com/talkiq/gcloud-aio/blob/master/.circleci/config.yml#L142) to ensure we keep the backwards-compatible version in `gcloud-rest`.

TheKevJames:
Closing this for now, but happy to reopen if we can get a fix which will support py2.7 as well!

PR #425, opened 2022-02-07
feat(olly): enable prometheus metrics in pubsub

jonathan-johnston:
I have a proposal for configuring Prometheus within `gcloud-aio-pubsub`:

* We’d like the PubSub client code to work transparently, without having to configure Prometheus at all or even know it exists
* When a Prometheus metric is initialized, its _full name is also initialized_ (ex. initialized to `pubsub_batch_size` and you can’t change it later to `vi_ctm_pubsub_batch_size`)
* Our existing Datadog metrics agent implementation in this library allows for custom namespaces via injection of the metrics agent with its namespace feature (i.e. metric names are resolved later, not up-front)
* We might want to compare between different services' PubSub performance, therefore we should maintain the same metric `namespace` and `subsystem` for this library, regardless of which service is using it (i.e. these metric names are not configurable) even though this differs from the existing DD implementation
* `prometheus-client` is not available for Python < 3.6, and we only have metrics for the `gcloud-aio-pubsub` package, so I have limited the installation and usage to `gcloud-aio-pubsub` only

I have a question about one of the existing metrics: `pubsub.{acker,nacker,producer}.batch`.

This is a histogram which tracks the size of batches that are processed successfully by one of the components. Is there any reason we can’t use a counter for this instead? Is there a specific reason that we want to use a histogram for this? The reason I bring this up is because the buckets for this are sensitive to the batch size that you set when using the PubSub subscriber. Ex. some apps pull 1000 messages, some can only pull 1. Configuring the buckets for this is a little odd because of this variability. BTW the default buckets are tuned mostly for latency in seconds and max out around `10.0`.

In addition, I’d like to take this opportunity to ask about how we’d like to use `labels`. AFAIU, it is best practice to use `labels` instead of constructing many different metric names. I’ve done that in this PR and I’d love feedback about it.


jonathan-johnston:
Edit: I had a convo with Kevin about this, and we came to the conclusion that since we already have metric labels that are automatically added to the service metrics and disambiguate by app, we don't have to configure `namespace`, etc. here (since we can just filter by label). And also, it could make sense to do aggregations or comparisons across different apps with respect to their PubSub performance. This would be harder to execute if the metric names were different, and would be easier to execute if the metric names were the same but labels (ex. `kind`) were different.

pubsub/README.rst
```diff
@@ -139,6 +139,23 @@ any metrics agent that implements the same interface as ``MetricsAgent``
...
+the subscriber records Prometheus metrics of the form ``gcloud_aio_<metric>``,
+which will be ignored if you don't use Prometheus and will be automatically
+included if you do:
+
+- ``pubsub_consume`` (labels: ``outcome = {'succeeded', 'cancelled', 'failed',
+  'failfast'}``) - [counter] a consume operation has completed with a given
+  outcome
+- ``pubsub_consume_latency`` (labels: ``aspect = {'receive', 'runtime'}``) -
+  [histogram] how many seconds taken to receive a message, or callback runtime
+- ``pubsub_batch_failed`` (labels: ``component = {'acker', 'nacker'}``) -
```
TheKevJames, line 152: Thoughts on making this a generic `pubsub_batch` metric with `outcome = {'succeeded', 'failed'}`?

pubsub/README.rst
```diff
@@ -139,6 +139,23 @@ any metrics agent that implements the same interface as ``MetricsAgent``
   in a single request.
 
 
+If you prefer pull-based metrics like Prometheus you will be pleased to know that
+the subscriber records Prometheus metrics of the form ``gcloud_aio_<metric>``,
+which will be ignored if you don't use Prometheus and will be automatically
+included if you do:
+
+- ``pubsub_consume`` (labels: ``outcome = {'succeeded', 'cancelled', 'failed',
```
TheKevJames, line 147: Should these all have unit suffixes?

pubsub/gcloud/aio/pubsub/metrics_agent.py
```diff
@@ -12,3 +15,30 @@ def increment(self,
                   metric: str,
                   value: float = 1) -> None:
         pass
+
+
+class PrometheusMetrics:
```
TheKevJames, line 20: We might also want to use the `metrics.py` namespace approach here?

pubsub/gcloud/aio/pubsub/metrics_agent.py
```diff
@@ -12,3 +15,30 @@ def increment(self,
...
+            'Counter for failures to process PubSub message batches',
+            ['component'], namespace=namespace, subsystem=subsystem)
+
+        self.messages_processed = prometheus_client.Counter(
+            'pubsub_messages_processed',
+            'Counter of successfully processed messages',
+            ['component'], namespace=namespace, subsystem=subsystem)
+
+
+PROMETHEUS = PrometheusMetrics(namespace='gcloud', subsystem='aio')
```
TheKevJames, line 44: Please double check this gets converted to `subsystem='rest'` at build time.

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -215,22 +224,27 @@ async def _execute_callback(message: SubscriberMessage,
...
-            metrics_client.histogram('pubsub.consumer.latency.runtime',
-                                     time.perf_counter() - start)
+            with PROMETHEUS.consume_latency.labels(aspect='runtime').time():
+                start = time.perf_counter()
+                await callback(message)
+                await ack_queue.put(message.ack_id)
+                metrics_client.increment('pubsub.consumer.succeeded')
+                metrics_client.histogram('pubsub.consumer.latency.runtime',
+                                         time.perf_counter() - start)
+                PROMETHEUS.consume.labels(outcome='succeeded').inc()
```
TheKevJames, line 234: nit: now that we won't need to carry around the `start` values, I wonder if these sorts of metrics would fit better in an eg. `else:` block. We probably don't want the time it takes to write the "success" metric to be included in the "consume latency", for example.

pubsub/gcloud/aio/pubsub/metrics_agent.py
TheKevJames: Hmm, should this be `namespace='gcloud_aio', subsystem='pubsub'`? I'm trying to think of whether there would ever be non-pubsub metrics here -- probably not?

jonathan-johnston: I was thinking the same thing!

pubsub/README.rst
jonathan-johnston: I'm good with that!

pubsub/gcloud/aio/pubsub/subscriber.py
jonathan-johnston: Agreed, same goes for the receive latency (which was re-calculated for Prometheus). What about something like this?

```python
        try:
            start = time.perf_counter()
            with metrics.CONSUME_LATENCY.labels(aspect='runtime').time():
                await callback(message)
                await ack_queue.put(message.ack_id)
            metrics_client.histogram('pubsub.consumer.latency.runtime',
                                     time.perf_counter() - start)
            metrics_client.increment('pubsub.consumer.succeeded')
            metrics.CONSUME.labels(outcome='succeeded').inc()
        ...
```

pubsub/README.rst
```diff
@@ -139,6 +139,25 @@ any metrics agent that implements the same interface as ``MetricsAgent``
...
+  'runtime'}``) - [histogram] how many seconds taken to receive a message, or
+  callback runtime
+- ``pubsub_batch_total`` (labels: ``component = {'acker', 'nacker'}, outcome =
+  {'succeeded', 'failed'}``) - [counter] a batch has succeeded or failed to be
+  acked or nacked
+- ``pubsub_messages_processed_total`` (labels: ``component = {'acker',
+  'nacker', 'producer'}``) - [counter] the number of messages that were acked,
+  nacked, or retrieved from pubsub
+
+
```
TheKevJames, line 160: I think the above usages of labels vs metrics all make sense. Fundamentally, each metric has semantic meaning as far as I call tell: eg. `sum(pubsub_consume_total) by (outcome)` is something that would be absolutely useful to track on its own and to view the sratio of the different `outcome` values.

My only counter-example here would be `pubsub_consume_latency_seconds` -- as far as I can tell, there's some time unaccounted for there (between when we spawn the callback task and when it actually starts) -- we might need a `pubsub_consume_latency_seconds, aspect = "asyncio_waiting"` (naming tbd) for the fully aggregated metric to be entirely meaningful. Thoughts?

(Note: I actually _really like_ how you grouped these metrics for exactly the reason of making the above obvious. I hadn't even realized that there was time being unaccounted for in our metrics until I went to check on whether these metrics are fully meaningful as a sum across all labels!)

pubsub/README.rst
```diff
@@ -139,6 +139,25 @@ any metrics agent that implements the same interface as ``MetricsAgent``
   in a single request.
 
 
+If you prefer pull-based metrics like Prometheus you will be pleased to know that
+the subscriber records Prometheus metrics of the form ``gcloud_aio_<metric>``,
+which will be ignored if you don't use Prometheus and will be automatically
+included if you do:
+
+- ``pubsub_consume_total`` (labels: ``outcome = {'succeeded', 'cancelled',
```
TheKevJames, line 147: Note: I think you renamed these to eg. `pubsub_subscriber_*`, need to make sure the readme matches code.

pubsub/README.rst
TheKevJames: I also really like the naming of your labels -- `aspect`, `component`, and `outcome` might be nice to standardize across our other PRs.

pubsub/gcloud/aio/pubsub/metrics.py
```diff
@@ -0,0 +1,41 @@
...
+    CONSUME_LATENCY = prometheus_client.Histogram(
+        'subscriber_consume_latency',
+        'Histogram of PubSub message consume latencies',
+        ['aspect'],
+        namespace=_NAMESPACE,
+        subsystem=_SUBSYSTEM,
+        unit='seconds')
+
+    BATCH = prometheus_client.Counter(
+        'subscriber_batch',
```
TheKevJames, line 28: I might argue for giving this a `_status` suffix (eg. since it kind of matches the `job_status` we've been using in other PRs). Alternatively, for removing the `_status` suffix elsewhere. Thoughts? I'm mostly looking for places where we should encode a bit of consistency at this point.

pubsub/README.rst
allandialpad: curious why'd you pick `aspect`? I associate that with multiple perspectives on the same object. So like I'd assume it mean something like the 'reliability aspect' or 'latency aspect' whereas here its trying to label separate parts of a sequence... right?

how about `one_of([step, stage, phase])`

love `outcome` lets do that

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: Yeah, that seems reasonable to me!

re: deprecation. Yeah, in fact we should probably disable them sooner rather than later to give more folks time to migrate. Thoughts on adding a `warrnings.warn` for when clients pass in a `metrics_client` as part of this PR?

pubsub/README.rst
TheKevJames: I do definitely like `step` for something that has an implied ordering (eg. perfect for etl, for example). Less certain about it for places that aren't ordered, though I don't believe we've hit that case yet. I'm ok with `phase` as well.

Veto for `stage`, since we already use that as the `[canary, release, pre-release, ...]` label.

pubsub/gcloud/aio/pubsub/subscriber.py
jonathan-johnston: A deprecation warning sounds like a good idea to me!

pubsub/README.rst
jonathan-johnston: **re: Kevin**

Ah, good point, it would be good to measure the "queueing" time so to speak.

**re: Allan**

I picked `aspect` as it describes a different "angle" on the same metric and it provides a space to insert a descriptor for how the metric differs from others with the same name (ex. one aspect of latency is how long it took to process the thing, one aspect is how long it took for the thing to start being processed, one aspect could be the sum of the two, etc.).

I'm happy to choose any other name for `aspect` although like Kevin mentioned, `step` implies ordering which I don't aim to achieve with `aspect`. I would also argue that `stage` and `phase` also imply a kind of non-overlapping sequence of measurements, but that interpretation could just be my own bias. Although we could just force this to be a sequential, non-overlapping metric and use one of the terms you suggested instead. Thoughts?

pubsub/gcloud/aio/pubsub/metrics.py
jonathan-johnston: I do like batch status better than just `batch`. Something about `batch` was bothering me.

pubsub/README.rst
TheKevJames: > what do you think about including a "total" time including receive, runtime, async queueing, and ack time which we could use to calculate out the non-obvious processing times?

We might have to play around with this a bit to see how well it works. I think for aggregated metrics, it's always better to "add up" to a total instead of subtracting from it, though that might be bad memories of Datadog bleeding through for me rather than something more globally accurate.

TheKevJames (approved):
LGTM! I think the only pending conversation is the aspect vs step decision, which I'm content to see go either way.

pubsub/README.rst
caseydialpad: I would submit to you there may now be merit in providing a timestamp to perform subtractive operations on opposed to additive ones (if only to record the perceived time based from a user-provided timestamp).

Alternatively, we could utilize Google Cloud Spanner for high fidelity measurements to more precisely gauge our system and its performance metrics: https://cloud.google.com/spanner/docs/true-time-external-consistency.

It's been a while, but I studied Spanner in "Designing Data Intensive Applications".

jonathan-johnston: Interesting find @caseydialpad, I wonder if the Cloud Spanner TrueTime is applicable to nodes outside of Cloud Spanner? It looks like it uses atomic clocks as well as only allowing transactions on intervals. I found this explainer blog and this is an interesting block:

> It’s important to keep in mind that TrueTime does not guarantee perfectly synchronized clocks. Rather, TrueTime gives an upper bound for clock offsets between nodes in a cluster. Synchronization hardware helps minimize the upper bound. In Spanner’s case, Google mentions an upper bound of 7ms. That’s pretty tight; by contrast, using [NTP](https://en.wikipedia.org/wiki/Network_Time_Protocol) for clock synchronization is likely to give somewhere between 100ms and 250ms.

caseydialpad: I see what Allan is saying about `step` versus `aspect`. Formerly there were `receiving` and `runtime` that could be interpreted as steps or aspects. They could be viewed as steps if they are intended to be separate stages, but they could also be aspects if we say `runtime` includes `receiving`. I think the latter muddies intuitive interpretation.

Perhaps I would ask what the DE definitions of `aspect` and `step` are so we can consistently use them.

To me the intuitive use of aspect would be to in terms of perception as in the case of art, and user experience. The user-aspect would be metrics that reflect how a user perceives something. Another aspect might be how a speaker with a non-optimal decoder perceives transcription. Another still might be transformations on the data to reflect some system aspect of the system (such as the memory requirements of the chunks of audio transcription)  to help us perceive what the decoder experiences as it has different experiences digesting information.

For steps, the term to me denotes something more discrete with beginning and ends as opposed to sequences. A sequence is definitely composed of steps but steps may be composed of substeps (so we'd need to tighten up the terminology). Like a CPU many things may happen in parallel, but each step is discrete and exclusive to itself. I'd also argue steps are less a matter of perception and more a matter of impersonal measurement.

Returning to `runtime`, `queue` and `receiving`, these feel strongly of `steps` given the discreteness of their nature. Each have well defined beginning and ends.

caseydialpad: @jonathan-johnston, you're right. If I recall correctly, the timestamp received is some interval with a degree of confidence attached to it (>95% I think). I've never used it, but where high quality timing is required and no internally consistent reference is available, the existence of remote atomic clocks are about as good as it gets.

jonathan-johnston: Thanks for your thoughtful post Casey! After reading your post I feel like:

* `step` should be some kind of verb, and describes something that you have control over (like ETL where extract, transform, and load are all discrete steps which the code in question _actively does_) and of course is a series of steps which achieves some goal
* `phase` can be something that is happening which generally progresses toward a final state but it may not be something that the code in question _does_ (ex. it does not _ actively do_ `runtime` but phrased another way it could _actively do_ `processing` or `process`)
* `aspect` is as discussed in that it describes a facet of something and doesn't have to be sequential in nature

Overall I am leaning toward `phase` or `aspect` for `receive`, `queuing`, and `runtime` of the `consume_latency` metric.

I hope I'm interpreting this correctly, but it looks to me like there are no strong opinions for which word we should use. I feel like `step` should definitely be used for measurement of actively executed steps (ex. ETL) and `aspect` should be used for non-actively executed things (ex. `receive`).

TheKevJames: Those definitions make sense to me, I'd be comfortable using the above as our guidelines. Unless there are any strong objections, let's go with how Jon laid out those descriptions (and @jonathan-johnston could you please reflect this in https://docs.talkiq-echelon.talkiq.com/static/docs/guides/observability/metrics.html#common-names ?)

jonathan-johnston: @TheKevJames Added the recommendations here: https://github.com/talkiq/docs/pull/239

jonathan-johnston: FYI I have settled on `phase` for this metric since the `consume_latency` metric is progressing toward and "end state" of the incoming message being completely consumed but the metric is making passive observations (therefore, `step` is out of the question).

TheKevJames (approved):

eddiedialpad (approved):
Approved.
• add prometheus for gcloud-aio pubsub — i'm going to be leaning on some of these for the pubsub part of realtime-nlp + realtime-egress — it would let me cut down on at least eg `receive` in egress.

pubsub/README.rst
```diff
@@ -113,9 +111,35 @@ Here's a few hints:
...
+``gcloud_aio_pubsub_<metric>``, which will have no effect if you don't use
+Prometheus to scrape app metrics:
+
+- ``subscriber_consume_total`` (labels: ``outcome = {'succeeded', 'cancelled',
+  'failed', 'failfast'}``) - [counter] a consume operation has completed with a
+  given outcome
+- ``subscriber_consume_latency_seconds`` (labels: ``phase = {'receive',
+  'queueing', 'runtime'}``) - [histogram] how many seconds taken to receive a
+  message, while waiting for processing, or to complete the callback
+- ``subscriber_batch_status_total`` (labels: ``component = {'acker', 'nacker'},
```
shaundialpad, line 128: Something about the naming of `subscriber_batch_status_total` is not sitting quite right with me. I think it is that, without looking at the labels, one would (or at least, I would) tend to assume this relates to batches pulled by the subscriber itself, when really it is only pertaining to messages being acked and nacked. I wonder, is there a word which encapsulates both `acker` and `nacker` which we could use to make this clearer? Perhaps even `subscriber_batch_ack_status_total` which still includes both `acker` and `nacker` as components. Not sure how I feel about that just yet. Of course, the answer might just be "just always look at the labels", but it would be great if this was less misleading.

pubsub/gcloud/aio/pubsub/metrics.py
```diff
@@ -0,0 +1,41 @@
...
+        'subscriber_consume_latency',
+        'Histogram of PubSub message consume latencies',
+        ['phase'],
+        namespace=_NAMESPACE,
+        subsystem=_SUBSYSTEM,
+        unit='seconds')
+
+    BATCH_STATUS = prometheus_client.Counter(
+        'subscriber_batch_status',
+        'Counter for success/failure to process PubSub message batches',
```
shaundialpad, line 29: Related to my comment here https://github.com/talkiq/gcloud-aio/pull/425/files#r878185626, this doesn't say anything about being specific to the acker and nacker. You could say that it shouldn't be specific to only the acker and nacker components, but then something like summing over all of the components would not make sense if you had, for example, `producer` as a component to represent messages being pulled.

pubsub/README.rst
```diff
@@ -113,9 +111,35 @@ Here's a few hints:
...
+  outcome = {'succeeded', 'failed'}``) - [counter] a batch has succeeded or
+  failed to be acked or nacked
+- ``subscriber_messages_processed_total`` (labels: ``component = {'acker',
```
shaundialpad, line 131: In this case, a total sum over `subscriber_messages_processed_total` doesn't really represent anything meaningful, does it? IIUC, it should at most be 2x the total number of messages that were pulled during a given time period. Does this suggest the metric should be broken up? (based on the guidelines of our [own metric guide](https://docs.talkiq-echelon.talkiq.com/static/docs/guides/observability/metrics.html#metric-grouping)

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -201,36 +213,50 @@ async def maybe_nack(ack_id: str) -> None:
...
                 await nack_queue.put(message.ack_id)
             log.warning('Application callback was cancelled')
             metrics_client.increment('pubsub.consumer.cancelled')
+            metrics.CONSUME.labels(outcome='cancelled').inc()
         except Exception:
             if nack_queue:
                 await nack_queue.put(message.ack_id)
             log.exception('Application callback raised an exception')
             metrics_client.increment('pubsub.consumer.failed')
+            metrics.CONSUME.labels(outcome='failed').inc()
```
shaundialpad, line 259: I wonder, is it possible for applications to inject their own labels to extend these metrics? It would be wonderful to solely rely on this library for handling all of the metric capturing, but at the same time we often include additional information with latency measurements and such. For example in realtime-egress we break down the number of failures by `stage`, `domain` and `entity_type`, and in `realtime-nlp` we break down runtime latency by whether or not the model emitted any `events` (not to mention which model emitted them)

shaundialpad:
> I have a question about one of the existing metrics: pubsub.{acker,nacker,producer}.batch. This is a histogram which tracks the size of batches that are processed successfully by one of the components. Is there any reason we can’t use a counter for this instead? Is there a specific reason that we want to use a histogram for this?

In the past we have used `pubsub.producer.batch` to gauge how an application should be scaled. For example have an application configured to pull 100 messages but only 1-10 messages are ever being pulled, perhaps we should reduce the scale. Similarly, if all instances of the application are pulling 90-100 messages, perhaps its time to bump the scale. I think it was useful for example for the initial tuning of realtime-egress and debugging some asyncio latency issues in realtime-runner. Perhaps we could do without it?

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: Adding labels at runtime isn't really something prometheus supports; metrics need to have all their metrics declared upfront. I think the "default" way to do what you're describing with prometheus is to have your app layer just create _new_ metrics as needed.

shaundialpad: Yeah that's what I was thinking, which is already what we are currently doing and thus are essentially duplicating some of our metric capturing. I was just wondering if we could clean up some of our application code (the metric capturing portion) in apps which make use of this library.

pubsub/gcloud/aio/pubsub/metrics_agent.py
jonathan-johnston: FYI `gcloud_rest` is not supported, and this is within an `if rest` block so we're all covered.

Either way the metrics will turn out as `gcloud_aio_pubsub_subscriber` whether `subsystem` is `subscriber` or `pubsub`. I'm inclined to use `subsystem=pubsub` to allow this to be copy-pasted into other clients like `subsystem=storage`. IIRC the `namespace` and `subsystem` construction here is only relevant in code. It is not transmitted as part of the metric to Prometheus.

pubsub/README.rst
jonathan-johnston: That's a great point, I think that receiving the messages is an entirely different metric. With just the `acker/nacker` IMO this now makes sense since the total messages processed (either via `ack` or `nack`) through the PubSub client can be summed by using this metric. Thoughts?

pubsub/README.rst
jonathan-johnston: I'm not quite sure I agree, since the decision to `nack` or `ack` is also one that the user can decide when they write their callback. Either of the outcomes can be seen as "success" and can be considered "one more item that the PubSub client has processed" either successfully or unsuccessfully. Summing this metric should give the number of messages that were received in total (now that I've split out the "received" metric from the "processed" metric). Maybe the "processed" metric should die and this one should take its place instead? It is largely duplicating its efforts.

jonathan-johnston:
Alright, I sorted out a bunch of comments as well as the REST build issues I was experiencing. Now just to do a quick test and it should be ready for release.

@shaundialpad LMK if you think the custom buckets for the `batch_size` metric should be tuned.

pubsub/README.rst
jonathan-johnston: Actually, my mental context was out-of-date and `processed` should continue to live on since it counts the actual number of messages passed through the acker or nacker and the `batch_status` counts the number of batches (of one or more messages).

jonathan-johnston:
Okay, successfully tested this with some bootleg local code:

Set up emulator:
```
gcloud beta emulators pubsub start --host-port=0.0.0.0:8681
```

Run the test:
```
export PUBSUB_EMULATOR_HOST='0.0.0.0:8681'
poetry install
poetry run python test.py
```

Test the Prometheus metrics:
```
curl http://0.0.0.0:8080/metrics
```

And the actual test definition:
```python
import asyncio
import random

import aiohttp
import prometheus_client
from gcloud.aio.pubsub import PubsubMessage
from gcloud.aio.pubsub import PublisherClient
from gcloud.aio.pubsub import subscribe
from gcloud.aio.pubsub import SubscriberClient
from gcloud.aio.pubsub import SubscriberMessage

async def do_publish():
    async with aiohttp.ClientSession() as session:
        client = PublisherClient(session=session)

        topic = client.topic_path('voiceai-local', 'test-topic')
        try:
            await client.create_topic(topic)
        except aiohttp.client_exceptions.ClientResponseError as ex:
            if ex.code == 409:
                print('topic already exists, continuing...')

        messages = [
            PubsubMessage(b'payload', attribute='value'),
            PubsubMessage(b'other payload', other_attribute='whatever',
                          more_attributes='something else'),
        ]

        for _ in range(100):
            response = await client.publish(topic, 10 * messages)
            print('published:', response)


async def do_subscribe():
    client = SubscriberClient()
    subscription = 'projects/voiceai-local/subscriptions/test-subscription'

    try:
        await client.create_subscription(
            subscription,
            'projects/voiceai-local/topics/test-topic')
    except aiohttp.client_exceptions.ClientResponseError as ex:
        if ex.code == 409:
            print('subscription already exists, continuing...')

    async def handler(message):
        print('got message:', message)
        if random.randint(0, 1):
            # nack it
            raise Exception('failed to process')
        # ack it
        return

    await subscribe(
        subscription,
        handler,
        client,
        num_producers=1,
        max_messages_per_producer=100,
        ack_window=0.3,
        num_tasks_per_consumer=1,
        enable_nack=True,
        nack_window=0.3,
    )


async def main():
    await do_publish()
    await do_subscribe()

if __name__ == '__main__':
    prometheus_client.start_http_server(8080)
    asyncio.run(main())

```

pubsub/gcloud/aio/pubsub/metrics.py
```diff
@@ -0,0 +1,57 @@
...
+        buckets=(0, 1, 5, 10, 25, 50, 100, 150, 250, 500, 1000, 1500, 2000,
+                 5000, float('inf')))
+
+    CONSUME = prometheus_client.Counter(
+        'subscriber_consume',
+        'Counter of the outcomes of PubSub message consume attempts',
+        ['outcome'],
+        namespace=_NAMESPACE,
+        subsystem=_SUBSYSTEM,
+        unit='total')
```
TheKevJames, line 26: I believe we've discovered that the `unit` flag should not be provided to `Counter` types, as it leads to `<metricname>_total_total` naming.

TheKevJames (approved):
Just a potential metric unit/naming issue, otherwise this looks great!

pubsub/gcloud/aio/pubsub/metrics.py
jonathan-johnston: Fair point, I'll update this.

pubsub/README.rst
shaundialpad: Sorry all I meant was that `BATCH_STATUS` and `BATCH_SIZE` are not necessarily referring to the same batch, which is fine, just that it could be misleading, hence the suggestion for the rename.

shaundialpad:
> Alright, I sorted out a bunch of comments as well as the REST build issues I was experiencing. Now just to do a quick test and it should be ready for release.
> 
> @shaundialpad LMK if you think the custom buckets for the `batch_size` metric should be tuned.

I think this will largely depend on the application, unfortunately. Some applications will only ever pull a single message. I don't recall what the original issue was with using a gauge type metric for this :P

shaundialpad:
Awesome stuff! 🎉 Can't wait to make use of this in situations where we don't require additional labels :P

PR #428, opened 2022-02-16
auth: allow to pass credentials as text

orsinium:
1. Allow passing credentials as a `pathlib.Path` object.
2. Allow passing credentials as text content instead of a path to a file.
3. Propagate exceptions for an invalid JSON if it is explicitly passed, not only for IO errors. And all types of exceptions, really, because why not.
4. Allow `CLOUDSDK_CONFIG` to be a path to the directory, not only to the file.
5. Cover it all with tests.

### DIFF
--- a/auth/gcloud/aio/auth/token.py
+++ b/auth/gcloud/aio/auth/token.py
@@ -6,6 +6,7 @@ import enum
 import json
 import os
 import time
+from pathlib import Path
 from typing import Any
 from typing import AnyStr
 from typing import Dict
@@ -45,7 +46,7 @@ GCE_ENDPOINT_TOKEN = (f'{GCE_METADATA_BASE}/instance/service-accounts'
                       '/default/token?recursive=true')
 GCLOUD_TOKEN_DURATION = 3600
 REFRESH_HEADERS = {'Content-Type': 'application/x-www-form-urlencoded'}
-ServiceFile = Optional[Union[str, IO[AnyStr]]]
+ServiceFile = Optional[Union[str, IO[AnyStr], Path]]
 
 
 class Type(enum.Enum):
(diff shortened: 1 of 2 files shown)


orsinium:
Wow. Why there are [tests on CI for Python 2.7](https://app.circleci.com/pipelines/github/talkiq/gcloud-aio/1693/workflows/021d167d-94c7-4825-a4d9-f8a6b194e50a/jobs/34875) if the [minimal supported version by the package](https://github.com/talkiq/gcloud-aio/blob/master/auth/setup.py#L25) is 3.6? I see [zero downloads from Python 2.7 ever](https://pypistats.org/packages/gcloud-aio-auth). Am I missing something? 🤔

TheKevJames (commented):
@orsinium ah, yes, the Py2 compatibility here is certainly odd. As mentioned [in the readme](https://github.com/talkiq/gcloud-aio/blob/master/README.rst), this repo actually contains the codebase for two projects: both `gcloud-aio-*` (which uses asyncio and is thus 3.6+) and `gcloud-rest-*` (which is py2.7-compatible). You can see there are a few py2 downloads for that project (https://pypistats.org/packages/gcloud-rest-auth); we'll eventually drop support, but for now we have a couple legacy users we're still supporting.

The `gcloud-rest-*` version of the libraries can be debugged by grabbing 'em from CI here (https://app.circleci.com/pipelines/github/talkiq/gcloud-aio/1693/workflows/021d167d-94c7-4825-a4d9-f8a6b194e50a/jobs/34866/artifacts). The CI job for generating them could also theoretically be run locally, but it's unfortunately complex (`https://github.com/talkiq/gcloud-aio/blob/26a34a3a2ed1e22622235e16ee5c9c35a3e745c3/.circleci/config.yml#L142`).

I will review this code as soon as I can -- previously, this logic matched behaviour with the `google-auth` library, so I'll need to make sure all the same cases are covered and any differences are documented clearly.

PR #429, opened 2022-02-16
Allow timeout for AioSession to be ClientTimeout

orsinium:
The PR removes explicit wrapping of the passed timeout into `aihttp.ClientTimeout` (aiohttp does wrapping on its own if needed), allowing to pass `aihttp.ClientTimeout` instance as `timeout` as well as any `float`. This gives users more control over timeouts.

### DIFF
--- a/auth/gcloud/aio/auth/session.py
+++ b/auth/gcloud/aio/auth/session.py
@@ -39,30 +39,30 @@ class BaseSession:
 
     @abstractmethod
     async def post(self, url: str, headers: Dict[str, str],
-                   data: Optional[str], timeout: int,
+                   data: Optional[str], timeout: float,
                    params: Optional[Dict[str, str]]) -> Response:
         pass
 
     @abstractmethod
     async def get(self, url: str, headers: Optional[Dict[str, str]],
-                  timeout: int, params: Optional[Dict[str, str]],
+                  timeout: float, params: Optional[Dict[str, str]],
                   stream: bool) -> Response:
         pass
 
     @abstractmethod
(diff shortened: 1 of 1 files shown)


vidursatija:
Can we please have this feature? My application would very much enjoy user controlled timeout.

Thank you,
Vidur

auth/gcloud/aio/auth/session.py
```diff
@@ -201,7 +200,7 @@ def session(self) -> Session:
         # symbol ensures we match the base class's definition for static
         # analysis.
         async def post(self, url: str, headers: Dict[str, str],
-                       data: Optional[str] = None, timeout: int = 10,
+                       data: Optional[str] = None, timeout: Timeout = 10,
```
TheKevJames, line 203: It looks like you've only defined `Timeout` on L82 (eg. the `if not BUILD_GCLOUD_REST:` block), so it would not be available here. Since `requests` wouldn't support `aiohttp.Timeout` values anyway, having a `Timeout = float` definition in the `if BUILD_GCLOUD_REST:` might do the trick?

TheKevJames (changes requested):
Hi @orsinium, sorry for delays on reviewing this and thank you for the contribution! Looks like there's a potential issue here around usage in `gcloud-rest-auth`, but otherwise I love the idea of this changeset! :+1:

auth/gcloud/aio/auth/session.py
orsinium: Good catch, thank you. You can't redefine type aliases. I've set the timeout type for non-aio code to float directly. Check it now, please.

PR #430, opened 2022-02-16
Make StreamResponse a context manager

orsinium:
If `aiohttp.ClientResponse` is read without releasing it, the connection is never released until the whole session is closed. This can be mitigated by reading the response inside a context manager.

```
#bad
resp = await session.get(url)
return await resp.content.read()

# good
async with session.get(url) as resp:
  return await resp.content.read()
```

The PR allows using `StreamResponse` as a context manager.

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -111,6 +111,7 @@ class StreamResponse:
     """This class provides an abstraction between the slightly different
     recommended streaming implementations between requests and aiohttp.
     """
+
     def __init__(self, response: Any) -> None:
         self._response = response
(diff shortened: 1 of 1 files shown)


TheKevJames (changes requested):
This makes sense to me! Thank you for the contribution @orsinium . Would it be possible for you to fix the linting issues here? Just need to get type hints added for the new methods.

TheKevJames:
Rebased for testing with lint fixes [here](https://github.com/talkiq/gcloud-aio/commit/7f1a4b52aa4689b32c4671400f52b1fddbd599ed). Merging from there and will get this packed into a release shortly.

PR #432, opened 2022-02-17
Decrease the lower bound of compatible requests libs

egalpin:
This is done in hopes to widen the compatibility within others' projects.


TheKevJames:
The changelog shows lots of important bugfixes over time: https://pyup.io/changelogs/requests/ For v2.20, maybe we needed case-insensitive headers?

egalpin:
I attached a debugger to the BQ integration tests and found that the prior HTTP 404 errors did not show up while I was tracing. I also noted the sleeps in the test and assumed that was to allow an out-of-band process to complete.  I've doubled the sleep times in the BQ integration test in hopes of lowering the flakiness.

egalpin:
Doubling the sleep time worked:  https://app.circleci.com/pipelines/github/talkiq/gcloud-aio?branch=egalpin-pubsub-deps-requests-2.2.1&filter=all

PR #433, opened 2022-03-02
fix(bigquery): avoid nesting base session objects

jonathan-johnston:
@leanaha and I were digging into some issues in porting an app to `asyncio` and noticed an error stemming from `aiohttp.ClientTimeout` having an underlying timeout value of type `aiohttp.ClientTimeout` when using `gcloud-aio-bigquery`. This threw an exception within `aiohttp` code when it attempted to compare the timeout with an integer unsuccessfully (https://github.com/aio-libs/aiohttp/blob/v3.8.1/aiohttp/helpers.py#L648). Initially I thought this issue might exist somewhere in the `aiohttp` code itself, but after some static analysis of the source this didn't appear to be the case. To get more info I traced through the app and noticed that when using the BQ `Table` class to generate a `Job` without specifying a `session`, the resulting `Job` had a double-nested `AioSession` object such that you could only get to the `Session` object like `job.session.session.session`. Looking at the `gcloud-aio-bigquery` code I realized that the `Table` class was indeed passing a `AioSession` object to `Job` instead of the expected `Session`, causing the `Job` to wrap it like `AioSession(AioSession(Session()))`.

In fact, we already do this double-referencing in other modules here (non-exhaustive list) but we just missed them in our BQ module:
https://github.com/talkiq/gcloud-aio/blob/master/pubsub/gcloud/aio/pubsub/subscriber_client.py#L44
https://github.com/talkiq/gcloud-aio/blob/master/kms/gcloud/aio/kms/kms.py#L45
https://github.com/talkiq/gcloud-aio/blob/master/storage/gcloud/aio/storage/storage.py#L136

This changeset fixes this issue.

### DIFF
--- a/auth/gcloud/aio/auth/session.py
+++ b/auth/gcloud/aio/auth/session.py
@@ -33,8 +33,8 @@ class BaseSession:
         self._timeout = timeout
 
     @abstractproperty
-    def session(self) -> Session:
-        pass
+    def session(self) -> Optional[Session]:
+        return self._session
 
     @abstractmethod
     async def post(self, url: str, headers: Dict[str, str],
@@ -114,7 +114,6 @@ if not BUILD_GCLOUD_REST:
                 timeout = aiohttp.ClientTimeout(self._timeout)
                 self._session = aiohttp.ClientSession(connector=connector,
                                                       timeout=timeout)
-            assert self._session is not None
             return self._session
 
         async def post(self, url: str, headers: Dict[str, str],
@@ -196,7 +195,6 @@ if BUILD_GCLOUD_REST:
             if not self._session:
                 self._session = Session()
                 self._session.verify = self._ssl
-            assert self._session is not None
             return self._session
 
         # N.B.: none of these will be `async` in compiled form, but adding the


auth/gcloud/aio/auth/session.py
```diff
@@ -33,8 +33,8 @@ def __init__(self, session: Optional[Session] = None, timeout: int = 10,
         self._timeout = timeout
 
     @abstractproperty
-    def session(self) -> Optional[Session]:
-        return self._session
+    def session(self) -> Session:
+        pass
```
leanaha, line 38: Is it ok to `pass` here? did you do it this way to let subclasses of `BaseSession` to override this method?

jonathan-johnston: @leanaha Great question. Since this is an `abstractproperty` (https://docs.python.org/3.7/library/abc.html#abc.abstractproperty), it _must_ be overridden by inheriting classes otherwise those classes cannot be instantiated:

> A class that has a metaclass derived from [ABCMeta](https://docs.python.org/3.7/library/abc.html#abc.ABCMeta) cannot be instantiated unless all of its abstract methods and properties are overridden.
https://docs.python.org/3.7/library/abc.html#abc.abstractmethod

You can still manually call an abstract method or property using the `super()` method but in this case you'll already have defined a `session` property anyway.

auth/gcloud/aio/auth/session.py
```diff
@@ -114,6 +114,7 @@ def session(self) -> aiohttp.ClientSession:
                 timeout = aiohttp.ClientTimeout(self._timeout)
                 self._session = aiohttp.ClientSession(connector=connector,
                                                       timeout=timeout)
+            assert self._session is not None
```
TheKevJames, line 117: What's the motivation of these two checks?

jonathan-johnston: >It also modifies the typing for BaseSession to ensure that BaseSession.session returns a Session object for all subclasses, as a None value should be made impossible.

That being said, the typing changes I made to `gcloud-aio-auth` are not actually necessary. Unfortunately this is a case of `pre-commit` running quite differently on my machine than in CI as there are far more false positives of typing errors on my machine. Over the course of my changes I still held the believe that `mypy` had to be satisfied here, `mypy` being unsure that the object is not `Optional` as the parent class suggests and necessitating an explicit non-`None` check that `mypy` respects.

jonathan-johnston: (I have removed these typing changes to `gcloud-aio-auth`)

TheKevJames: Ah, I hadn't linked that comment to the `asserts`, that makes sense (TIL about [that](https://github.com/python/typing/issues/645), my guess for how to accomplish this would have been to [use `typing.cast` somehow](https://docs.python.org/3/library/typing.html#typing.cast)).

Definitely curious about the differences in usage between your machine and CI. I feel like we've been getting more reports of that sort of behaviour lately, which is troubling..

TheKevJames (approved):

PR #434, opened 2022-03-04
Updates version to 3.7.1 for patch release

egalpin:
Patch version update to indicate the change represented by https://github.com/talkiq/gcloud-aio/pull/432

### DIFF
--- a/auth/setup.py
+++ b/auth/setup.py
@@ -13,7 +13,7 @@ with open(os.path.join(PACKAGE_ROOT, 'requirements.txt')) as f:
 
 setuptools.setup(
     name='gcloud-aio-auth',
-    version='3.7.0',
+    version='3.7.1',
     description='Python Client for Google Cloud Auth',
     long_description=README,
     namespace_packages=[


eddiedialpad (approved):
Approved ... ?
• this looks like it's intended to widen compatibility 
• mmm this patch would make the changes in https://github.com/talkiq/gcloud-aio/pull/432 live ... but ...

⚠️ The only change is https://github.com/talkiq/gcloud-aio/pull/432/files#diff-b7a6e49386798ba76f54ca957a8a9e813dc41091c7445a1d5c234c2a6e937933R9

... Which is a commented out line in requirements.txt ...
```
# requests>=2.2.1,<3.0.0
```

Does this patch and the relevant change set really change anything?

egalpin:
@eddiedialpad Ya it's a bit odd looking, I'll give you that. In the CI processing, some commented sections are uncommented and swapped for other sections so as to build aio or rest libs from the "same" source.

eddiedialpad:
Hi Future Humans — in case you're wondering about when `requirements.txt` gets pre-processed ahead of installation, one such case is in the conversion of code back to Python2 from Python3 — an example of that change is this line:

https://github.com/talkiq/gcloud-aio/blob/master/.circleci/config.yml#L184

PR #435, opened 2022-03-07
Allow timeout in get_blob in storage

vector-ml:
Just a minor feature that lets the user provide timeout to `get_blob`.


TheKevJames (changes requested):
@vector-ml thanks for the contribution! Unfortunately, it looks like linting is failing. Could you please fix that up? Looks great to me otherwise.

vector-ml:
@TheKevJames it's fixed now. Thank you for the fast review!

TheKevJames (approved):
Perfect, thank you!

PR #441, opened 2022-03-11
fix(storage): fix warning uploading multipart data

### LINKED ISSUE #404: Warning when uploading large data using multipart
In aio.storage, in the `_upload_multipart()` function, the stream is converted to `bytes`. It cause aiohttp to use BytesPayload, which raises a warning if the data size is greater than 1MB. 

I think the `post` in `_upload_multipart()` should wrap body data into a `io.BytesIO` instead, to prevent that warning.

bruchar1:
Warp data in io.BytesIO when posting to ensure aiohttp do not use BytesPaylood. Fixes #404.

Not sure how to write a test for that, since it would require a >1MB data payload to trigger the warning...

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -539,8 +539,8 @@ class Storage:
         })
 
         s = AioSession(session) if session else self.session
-        resp = await s.post(url, data=BytesIO(body), headers=headers, params=params,
-                            timeout=timeout)
+        resp = await s.post(url, data=BytesIO(body), headers=headers,
+                            params=params, timeout=timeout)
         data: Dict[str, Any] = await resp.json(content_type=None)
         return data
 
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -17,6 +17,11 @@ from typing import Tuple
 from typing import Union
 from urllib.parse import quote
 
+try:
+    from cString import StringIO as BytesIO
+except ImportError:
+    from io import BytesIO
+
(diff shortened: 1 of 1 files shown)


storage/gcloud/aio/storage/storage.py
```diff
@@ -17,6 +17,11 @@
 from typing import Union
 from urllib.parse import quote
 
+try:
+    from cString import StringIO as BytesIO
```
TheKevJames, line 21: I'm not familiar with this library. By any chance, did you mean to refer to [`cStringIO`](https://docs.python.org/2/library/stringio.html#module-cStringIO) ?

bruchar1: As I understand, `str` in Python 2 is equivalent to `bytes` in Python 3, and `unicode` in Python 2 is equivalent to `str` in Python 3. Therefore, `StringIO` in Python 2 is equivalent to `BytesIO` in Python 3. `cStringIO` is only available in Python 2, and import will fail in Python 3. 

But rethinking about this, I think it would be better to only wrap data with when using aiohttp, and skip this with rest api. I will update the PR accordingly.

TheKevJames: Right, sorry, what I'd meant to call out was the literal name of the import.
```
~ » python2
>>> import cString
ImportError: No module named cString
>>> import cStringIO
>>> 
```

That said, your comment on the other approach makes sense to me, so I'll take a look at reviewing that instead :)

TheKevJames (approved):
Makes total sense with the explicit build check here :+1:

PR #442, opened 2022-03-15
fix(storage): upload retry results in IO error

bruchar1:
The retry mechanism in storage._do_upload() is broken because when you
perform a request on a stream in aiohttp, the stream is automatically
closed. Therefore, the retry raises an error: I/O Operation on closed
file.

This fix monkeypatch the close function on the stream to ensure
it remains open until all retries are performed.

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -263,8 +263,8 @@ class Storage:
     async def download_to_filename(self, bucket: str, object_name: str,
                                    filename: str, **kwargs: Any) -> None:
         async with file_open(  # type: ignore[attr-defined]
-            filename,
-            mode='wb+',
+                filename,
+                mode='wb+',
         ) as file_object:
             await file_object.write(
                 await self.download(bucket, object_name, **kwargs)
@@ -378,8 +378,8 @@ class Storage:
                                    filename: str,
                                    **kwargs: Any) -> Dict[str, Any]:
         async with file_open(  # type: ignore[attr-defined]
-            filename,
-            mode='rb',
+                filename,
+                mode='rb',
         ) as file_object:
             contents = await file_object.read()
             return await self.upload(bucket, object_name, contents,
@@ -612,17 +612,26 @@ class Storage:
                          timeout: int = 30) -> Dict[str, Any]:
         s = AioSession(session) if session else self.session
 
-        for tries in range(retries):
-            try:
-                resp = await s.put(session_uri, headers=headers,
-                                   data=stream, timeout=timeout)
-            except ResponseError:
-                headers.update({'Content-Range': '*/*'})
-                await sleep(2. ** tries)  # type: ignore[func-returns-value]
+        original_close = stream.close
+        original_position = stream.tell()
+        # Prevent the stream being closed if put operation fails
+        stream.close = lambda: None  # type: ignore[assignment]
+        try:
+            for tries in range(retries):
+                try:
+                    resp = await s.put(session_uri, headers=headers,
+                                       data=stream, timeout=timeout)
+                except ResponseError:
+                    headers.update({'Content-Range': '*/*'})
+                    stream.seek(original_position)
 
-                continue
-
-            break
+                    await sleep(  # type: ignore[func-returns-value]
+                        2. ** tries
+                    )
+                else:
+                    break
+        finally:
+            original_close()
 
         data: Dict[str, Any] = await resp.json(content_type=None)
         return data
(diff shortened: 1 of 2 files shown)


bruchar1:
I realised integration tests are not run on the CI... The test I wrote does not use a real cloud storage. Should I move it to unit tests?However, it needs `pytest-aiohttp` for the test server fixture, and woudl not work with rest. Should I add a `pytest.mark.skipif` for that?

TheKevJames (commented):
@bruchar1 thanks for the changeset! Overall this looks great.

We are aware of the issues in CI; we let things fall a bit behind in our configs there. I've got a changeset #443 which should fix this, and additionally which will ensure that test dependencies are available in both unit and integration tests rather than the incredibly manual and separate workflows we have now.

Hopefully that'll go in soon, which should unblock you from building this test case as a `unit` test. You can feel free to rebase off that branch if you'd like :) If it takes too long to get in or if you'd prefer, I'd be happy to make the relevant change on this branch for ya.

TheKevJames:
> However, it needs pytest-aiohttp for the test server fixture, and woudl not work with rest. Should I add a pytest.mark.skipif for that?

Yeah, any tests which do not work in `gcloud-rest` would need to be skipped. That said, it would be even better if we could write the tests to work in both cases -- would it be possible to have it use `requests` when being tested under `gcloud-rest` instead?

bruchar1:
Yes. I will use `http.server` instead. I will be more compatible with both implementations. 

But in your integration tests, how are you able to do `async with Session` when using the synchronous `requests.Session` ?

TheKevJames:
@bruchar1 some amount of code modification happens in the conversion stage, see the [CI conversion job here](https://github.com/talkiq/gcloud-aio/blob/7793ee034ab335da85fabfdc5c3169587e553e87/.circleci/config.yml#L142).

Note that this is a bit hard to read and can't be run locally: #443 fixes that as well by moving it to a `poetry run bin/build-rest` command.

storage/gcloud/aio/storage/storage.py
```diff
@@ -612,17 +612,26 @@ async def _do_upload(self, session_uri: str, stream: IO[AnyStr],
...
-                headers.update({'Content-Range': '*/*'})
-                await sleep(2. ** tries)  # type: ignore[func-returns-value]
-
-                continue
-
-            break
+        original_close = stream.close
+        original_position = stream.tell()
+        # Prevent the stream being closed if put operation fails
+        stream.close = lambda: None  # type: ignore[assignment]
```
eddiedialpad, line 618: 👁❗️ — okay.

storage/gcloud/aio/storage/storage.py
```diff
@@ -612,17 +612,26 @@ async def _do_upload(self, session_uri: str, stream: IO[AnyStr],
...
+            for tries in range(retries):
+                try:
+                    resp = await s.put(session_uri, headers=headers,
+                                       data=stream, timeout=timeout)
+                except ResponseError:
+                    headers.update({'Content-Range': '*/*'})
+                    stream.seek(original_position)
+
+                    await sleep(  # type: ignore[func-returns-value]
+                        2. ** tries
```
eddiedialpad, line 629: We're at 32s by the fifth try. This kinda bothers me because downstream users might end up with higher latency than their own apps are configured to tolerate, and have no recourse after they find out why they're timing out — we could potentially make the base `2.` an arg — but seeing as this is inherited, we can probably live with it.

eddiedialpad (approved):
approved: allow retries to actually complete, by preventing the streaming upload file object from closing before the final retry

TheKevJames:
Thanks for the contribution! And sorry for my very late follow-up on it.

Rebased for testing [[here](https://github.com/talkiq/gcloud-aio/commit/93211d11526c745005e34587b882a54a9064c2d6)](https://github.com/talkiq/gcloud-aio/commit/30b901a7243439591f837fea49a5861b12ac74be) with tests passing [here](https://app.circleci.com/pipelines/github/talkiq/gcloud-aio/2209/workflows/1af63dbb-8eb7-4880-bae8-929455d48c56) (not sure why tests are behaving oddly for external contributors at the moment, I'll figure that out when I can). Merging from there and will get this packed into a release shortly.

Note that I did have to modify the test fixture a bit (see `StoppableHTTPServer`) to avoid the tests hanging after completion, in case you're curious!

rad-pat:
I believe this is causing issue https://github.com/talkiq/gcloud-aio/issues/575 . Probably when the retries are exhausted, the ResponseError should be re-raised?

PR #444, opened 2022-03-17
fix(auth): coerce timeout to clienttimeout for clientsession

jonathan-johnston:
When sending requests, `aiohttp` intelligently converts timeout values from floats and ints into `ClientTimeout` but also accepts `ClientTimeout` (https://github.com/aio-libs/aiohttp/blob/v3.8.1/aiohttp/client.py#L440). Unfortunately, the `ClientSession` constructor is not so helpful and only accepts `ClientTimeout` values (https://github.com/aio-libs/aiohttp/blob/v3.8.1/aiohttp/client.py#L277).

Therefore when constructing the `ClientSession` object we must convert the `self._timeout` member variable to a `ClientTimeout` object before passing it in.

### DIFF
--- a/auth/gcloud/aio/auth/session.py
+++ b/auth/gcloud/aio/auth/session.py
@@ -111,13 +111,20 @@ if not BUILD_GCLOUD_REST:
 
     class AioSession(BaseSession):
         _session: aiohttp.ClientSession
+        _timeout: Timeout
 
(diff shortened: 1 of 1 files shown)


caseydialpad (approved):
I though `ClientSession` did this under the hood. What does it do?

PR #459, opened 2022-04-06
feat(datastore): add support for cursors on GQL queries

javierdialpad:
According to the [GQL Reference](https://cloud.google.com/datastore/docs/reference/gql_reference#clauses), cursors are available on the `LIMIT` and `OFFSET` clauses of GQL queries.

This PR gives support for using cursors as binding parameters of GQL queries. A parameter on a GQL query can either be a value of different types (already implemented) or a cursor (implemented on this PR). Check the syntax [here](https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/runQuery#gqlqueryparameter).

### DIFF
--- a/datastore/tests/integration/smoke_test.py
+++ b/datastore/tests/integration/smoke_test.py
@@ -4,6 +4,7 @@ import pytest
 from gcloud.aio.auth import BUILD_GCLOUD_REST  # pylint: disable=no-name-in-module
 from gcloud.aio.datastore import Datastore
 from gcloud.aio.datastore import Filter
+from gcloud.aio.datastore import GQLCursor
 from gcloud.aio.datastore import GQLQuery
 from gcloud.aio.datastore import Key
 from gcloud.aio.datastore import Operation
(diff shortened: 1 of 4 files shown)


datastore/tests/integration/smoke_test.py
```diff
@@ -245,6 +246,41 @@ async def test_gql_query(creds: str, kind: str, project: str) -> None:
         assert len(after.entity_results) == num_results + 3
 
 
+@pytest.mark.asyncio
+@pytest.mark.xfail(strict=False)
```
TheKevJames, line 250: Could you please explain why you've marked this as expecting to fail? The test makes sense to me!

TheKevJames (commented):
Overall looks great, just a question on the xfail!

datastore/tests/integration/smoke_test.py
javierdialpad: The output of this test depends on the records being inserted to be consistent before the reads. Given the eventual consistency property of Google's Datastore, the test could fail, event using the datastore emulator. I was not able to make it pass 100% of the times, even trying things like `consistency.STRONG` on the `runQuery` calls.

javierdialpad: Im sure the same happened with the test above this one, which I used to based my test.

TheKevJames: That's too bad. Man, datastore is such a pain to work with haha. I'll never understand why the world starting thinking not having ACID-complaince in your database was a good idea...

javierdialpad: Indeed! Took me a while to understand what was happening in this simple test.

PR #461, opened 2022-04-13
fix: Subscriber shutdown exception handling fix

nosammai:
In python3.8 CancelledError was changed to be a subclass of BaseException

This makes sure to catch CancelledError so that it will trigger the shutdown behavior on the subscriber


TheKevJames (approved):
Thanks @nosammai , good catch!

PR #482, opened 2022-06-13
Pass entity_result_type to EntityResult.from_repr and Entity.from_repr

nicolasazrak:
This change modifies `EntityResult.from_repr` and `Entity.from_repr` to receive `entity_result_type` as a parameter. We are currently using it to add some extra validations. For example: if the `entity_result_type` is `FULL` we validate that all required props of our model are set. However, if it is a `PROJECTION` we can ignore missing values. 

This might be a more common use case, so it might be a good idea to upstream it to this library.


TheKevJames:
Closing as we discussed offline.

PR #483, opened 2022-06-16
feat(bigquery): set job_id after job insertion

shaundialpad:
It seems not setting `self.job_id` after inserting a new job was an oversight. We do a similar thing in [Job.insert_via_query](https://github.com/talkiq/gcloud-aio/blob/b9821a5bb9103ce88771da4ff03f09d32658e181/bigquery/gcloud/aio/bigquery/job.py#L128) so I see no reason why we should not be doing it here. It would allow for the following pattern:

```python
job = Job()
await job.insert(job_config)
await job.get_query_results()
```

Note that `'jobId'` is not included in the response when one sets `'dryRun'` in the request. We could add the kwarg `dry_run: bool` to the method signature, however unlike the `Job.insert_via_query` method, the `Job.insert` method allows the user to specify the entire [Job object](https://cloud.google.com/bigquery/docs/reference/rest/v2/Job), which includes setting `dryRun` [in the job configuration](https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfiguration), so it seems illogical to have it set additionally as a kwarg.


bigquery/gcloud/aio/bigquery/job.py
```diff
@@ -103,7 +103,9 @@ async def insert(self, job: Dict[str, Any],
         project = await self.project()
         url = f'{API_ROOT}/projects/{project}/jobs'
 
-        return await self._post_json(url, job, session, timeout)
+        response = await self._post_json(url, job, session, timeout)
+        self.job_id = response['jobReference'].get('jobId')
```
jonathan-johnston, line 107: Since you can create a `Job` with a given `job_id`, this line will pave over that value with `None` if this is a dry run. Although generally you should use this `Job` object for a single job, I don't see any protection against re-use.

To align with other behaviour in this class (ex. `insert_via_query`), I think we should avoid setting `self.job_id` at all if the `jobId` field is not present as in the case of `dryRun` like you had mentioned.

jonathan-johnston (commented):
This makes sense to me to have the behaviour align with our other method of inserting a job for async execution.

bigquery/gcloud/aio/bigquery/job.py
shaundialpad: That's true, I was assuming the expected usage pattern was single use. I'm fine with setting only if it exists.

jonathan-johnston (approved):

PR #486, opened 2022-06-29
feat(bigquery): convert responses to dicts

TheKevJames:
Convenience method so client libs don't need to do the converting
themselves to get a slightly more useable dictionary format from their
BQ responses.

### DIFF
--- a/bigquery/gcloud/aio/bigquery/utils.py
+++ b/bigquery/gcloud/aio/bigquery/utils.py
@@ -1,8 +1,12 @@
 import datetime
+import logging
 from typing import Any
 from typing import Callable
 from typing import Dict
 from typing import List
+
+
+log = logging.getLogger(__name__)
 
 
 def flatten(x: Any) -> Any:
@@ -20,13 +24,7 @@ def flatten(x: Any) -> Any:
         # TODO: what if a user has stored a dictionary in their table and that
         # dictionary is shaped the same way as Google's response format?
         if 'f' in x:
-            print(x)
-            nested = [flatten(y['v']) for y in x['f']]
-            if len(nested) == 1:
-                # Nested data is always in a list form, even if it's a single
-                # value that cannot be REPEATED. Why? Who knows.
-                return nested[0]
-            return nested
+            return [flatten(y['v']) for y in x['f']]
 
         if 'v' in x:
             return flatten(x['v'])
@@ -37,7 +35,7 @@ def flatten(x: Any) -> Any:
     return x
 
 
-def parse(field: Dict[str, str], value: Any) -> Any:
+def parse(field: Dict[str, Any], value: Any) -> Any:
     """
     Parse a given field back to a Python object.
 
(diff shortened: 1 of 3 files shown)


kylepad (commented):
Some thoughts

bigquery/gcloud/aio/bigquery/utils.py
```diff
@@ -0,0 +1,83 @@
...
+      to flatten *first*, then convert.
+    """
+    f: Callable[[Any], Any] = {  # type: ignore[assignment]
+        'BOOLEAN': lambda x: x == 'true',
+        'FLOAT': float,
+        'INTEGER': int,
+        'RECORD': list,
+        'STRING': str,
+        'TIMESTAMP': lambda x: datetime.datetime.fromtimestamp(float(x)),
+    }[field['type']]
```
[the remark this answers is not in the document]
TheKevJames: I wasn't able to find a definitive list of which types existed or how to represent them. The [docs](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#TableFieldSchema) and [implementation](https://github.com/googleapis/python-bigquery/blob/76d88fbb1316317a61fa1a63c101bc6f42f23af8/google/cloud/bigquery/enums.py#L216) don't quite match.

That said, I can certainly support a few more low-hanging fruit and at least log debug information for capturing the rest!

bigquery/gcloud/aio/bigquery/utils.py
```diff
@@ -0,0 +1,83 @@
...
+    The API returns data of the form:
+
+        {'f': [{'v': ...}]}
+
+    to indicate groupings of fields and their potential for having multiple
+    values. We want those to just be plain old objects.
+    """
+    if isinstance(x, dict):
+        # TODO: what if a user has stored a dictionary in their table and that
+        # dictionary is shaped the same way as Google's response format?
```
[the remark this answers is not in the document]
TheKevJames: I'm not 100% sure this is even possible... but yeah, figured that it'd be worth including this here in case we run into Weird™ errors in the future.

bigquery/tests/unit/utils_test.py
```diff
@@ -0,0 +1,118 @@
...
+        'jobReference': {'projectId': 'sample-project',
+                         'jobId': 'job_Tlpl-66ca7a8e365a28084c39ffc52d402671',
+                         'location': 'US'},
+        'rows': rows,
+        'totalRows': '2',
+        'totalBytesProcessed': '0',
+        'jobComplete': True,
+        'cacheHit': True,
+    }
+    print(from_query_response(resp))
```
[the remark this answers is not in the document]
TheKevJames: Ah, yeah, I was originally only testing for failures and printing this out as a debug statement to help along the way. I should probably update this to assert on the return value anyway :+1:

bigquery/gcloud/aio/bigquery/utils.py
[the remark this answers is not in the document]
kylepad: idk those two examples do line up. Like the implementation detail you linked to normalized the SQL types to match the list in the docs. 
Like I think we could definitely add the following types pretty in failure straightforward manners:
`BYTES`
`DATE`
`TIME`
`DATETIME`

Might have to think a little bit about `NUMERAL` and `BIGNUMERAL`, as far as like should they be implemented using numpy or something else? or just left as strings.

The only type I'm really not sure how to handle is the `GEOGRAPHY` type, but there's probably a gold-standard GIS library that supports this, but I think that's niche enough we could leave it as a string with a comment like "if you think you what the right thing to do with this type is submit a PR"

We can safely ignore `JSON` and `INTERVAL` as they're currently marked as Pre-GA [here](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types)

My 2c at least

bigquery/gcloud/aio/bigquery/utils.py
```diff
@@ -0,0 +1,119 @@
...
+            'STRING': str,
+            'TIMESTAMP': lambda x: datetime.datetime.fromtimestamp(float(x)),
+        }[field['type']]
+    except KeyError:
+        # TODO: determine the proper methods for converting the following:
+        # BIGNUMERIC
+        # DATE
+        # DATETIME
+        # GEOGRAPHY
+        # TIME
```
harsh-dialpad, line 71: Can we leave them in the format returned by the API itself?

Maybe converting to a loss less datatype might be sufficient? For instance, int, str, str (both date times in ISO 8601), [unsure] , str?

bigquery/gcloud/aio/bigquery/utils.py
```diff
@@ -0,0 +1,119 @@
...
+    * NULLABLE fields should be handled specially, eg. so as not to
+      accidentally convert them to the schema type.
+    * REPEATED fields are nested a biot differently than expected, so we need
+      to flatten *first*, then convert.
+
+    `Field = Dict[str, Union[str, 'Field']]`, but wow is that difficult to
+    represent in a backwards-enough compatible fashion.
+    """
+    try:
+        f: Callable[[Any], Any] = {  # type: ignore[assignment]
```
harsh-dialpad, line 55: Can we call this variable something else? Makes it hard to comprehend as there is `f` being used in the list/dict unpacking as well.

bigquery/gcloud/aio/bigquery/utils.py
```diff
@@ -0,0 +1,119 @@
...
+
+    if field['mode'] == 'NULLABLE' and value is None:
+        return value
+
+    if field['mode'] == 'REPEATED':
+        if field['type'] == 'RECORD':
+            # TODO: The [0] and all this special-casing is suspicious. Is there
+            # a case I'm missing with overly nested RECORDS, perhaps?
+            # I suspect this entire block can get reduced down to a single case
+            # and then inserted into the dict of Callables above.
```
harsh-dialpad, line 85: Not sure at the moment!

harsh-dialpad (approved):
LGTM in essence, just a few nits. The test cases look good, thanks!

bigquery/gcloud/aio/bigquery/utils.py
TheKevJames: bignumeric -> int wouldn't work IIUC as bignumeric can hold larger values; I don't want to silently change folks' data on them. Converting these all to str imight be a decent idea, though... I guess I'm torn given `datetime` is probably the "right" answer. Is it better to convert them to string now and introduce a breaking change later once we have the sample data for converting 'em, or to introduce new support for additional fields via point releases? I'm leaning to the later.

bigquery/gcloud/aio/bigquery/utils.py
TheKevJames: Yeah, good call.

bigquery/gcloud/aio/bigquery/utils.py
TheKevJames: Ah, sorry, my mistake -- I was thinking of precision rather than max value. NUMERIC and BIGNUMERIC types have [38 and ~77 decimal places of precision](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types) respectively, whereas python floats are 18 places.

I guess that means I should use [decimal](https://docs.python.org/3/library/decimal.html) to do this properly.

kylepad (approved):
quality improvement 👍

PR #500, opened 2022-08-20
fix(auth): Guarantee we do not try to await for access token retrieval after the action has been cancelled

mykola-mokhnach:
We had a situation where after being canceled once all the following gcloud uploads and downloads started failing randomly with CancelledError. The stack trace helped to figure out that the exception is thrown inside of `ensure_token` call, in the following block of code:

```
        if self.acquiring:
            await self.acquiring # <<<
            return
``` 

The actual issue was that the future, which `self.acquiring` has been pointing to, has been already canceled. So every time we were awaiting it was throwing the same CancelledError, no matter how many times `ensure_token` has been called.

The fix makes sure `self.acquiring` does not get awaited if it was cancelled before

### DIFF
--- a/auth/gcloud/aio/auth/token.py
+++ b/auth/gcloud/aio/auth/token.py
@@ -145,19 +145,15 @@ class Token:
         return self.access_token
 
     async def ensure_token(self) -> None:
-        if self.acquiring:
+        if self.acquiring and not self.acquiring.cancelled():
             await self.acquiring
             return
 
-        if not self.access_token:
-            self.acquiring = asyncio.ensure_future(self.acquire_access_token())
-            await self.acquiring
-            return
-
-        now = datetime.datetime.utcnow()
-        delta = (now - self.access_token_acquired_at).total_seconds()
-        if delta <= self.access_token_duration / 2:
-            return
+        if self.access_token:
+            now = datetime.datetime.utcnow()
+            delta = (now - self.access_token_acquired_at).total_seconds()
+            if delta <= self.access_token_duration / 2:
+                return
 
         self.acquiring = asyncio.ensure_future(self.acquire_access_token())
         await self.acquiring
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
I am also in doubts whether I should change the `if self.acquiring and not self.acquiring.cancelled()` condition to `if self.acquiring and not self.acquiring.done()`. The second option would also allow to properly handle possible repeating exceptions inside `acquire_access_token` call, when `self.acquiring` is eventually never set to None.

TheKevJames:
Hi @mykola-mokhnach and sorry for the incredibly late response on this. Would you be able to tell me more about the situation in which the task was cancelled? Just trying to think about how we would want to fit this into error recovery.

The way I see it, there's _probably_ two main potential approaches we could take:
* assume the cancellation was user-driven / intentional / etc and propagate the error up: if the refresh task is cancelled, perhaps we should cancel the whole `Token` usage context?
* assume the cancellation was error-driven / unintentional / transient / etc and treat it basically as a failure/retry

It looks like you've aimed towards the second case, just want to make sure that makes sense holistically before approving.

There might also be another possible approach here, which is to replace the first block in that method with:
```python
if self.acquiring:
    if not self.acquiring.cancelled():
        await self.acquiring
        return

    self.acquiring = None
```
which could let us stick with the existing "create an acquiring future if our token is old" logic. Thoughts?

mykola-mokhnach:
In my case the cancellation was intentional, e.g. the scenario is: start file upload to GCS and in parallel start parsing file properties. If the latter fails then the upload gets cancelled as well.

Also, my current approach eliminates some redundancy which is present in the method

TheKevJames:
Ah, good call -- hadn't noticed the redundancy on first glance.

> In my case the cancellation was intentional, e.g. the scenario is: start file upload to GCS and in parallel start parsing file properties. If the latter fails then the upload gets cancelled as well.

Ok, interesting. So are you cancelling the refresh loop explicitly there? If you have a code link, that'd help me understand. Currently, as I understand it, you would be doing:
```python
task = storage.upload(..)

task.cancel()
storage.token.acquiring.cancel()
```
but perhaps I'm missing something with respect to how this specific task ends up cancelled?

mykola-mokhnach:
> Ah, good call -- hadn't noticed the redundancy on first glance.
> 
> > In my case the cancellation was intentional, e.g. the scenario is: start file upload to GCS and in parallel start parsing file properties. If the latter fails then the upload gets cancelled as well.
> 
> Ok, interesting. So are you cancelling the refresh loop explicitly there? If you have a code link, that'd help me understand. Currently, as I understand it, you would be doing:
> 
> ```python
> task = storage.upload(..)
> 
> task.cancel()
> storage.token.acquiring.cancel()
> ```
> 
> but perhaps I'm missing something with respect to how this specific task ends up cancelled?

Not really. Cancellation is asyncio works the way promises are canceled in cascade. So basically, in order to reproduce that issue you need to keep canceling uploads until the token expires, so then when the next upload is initiated then it waits until the token is refreshed first and there it gets cancelled. Basically, in my real life case the code was working fine for some hours or even days and then suddenly all calls to uploads were starting to fail. Took me a while to understand what the actual issue is. aiohttp handles CancelledError in a specific way, so one does not really see what the stack trace is without applying some custom hacks.

TheKevJames:
Gotcha! Ok, that helps me understand the issue here, thanks!

I've rebased this PR for testing [here](https://app.circleci.com/pipelines/github/talkiq/gcloud-aio/2246/workflows/497c8db0-d2e5-4839-8b21-ef1929064cc4) and [merged it directly](https://github.com/talkiq/gcloud-aio/commit/271229d2037dcb1a9a9b1d079d61db0b53e42df4). Thanks again for the fix! I'll get this released as soon as I can.

PR #503, opened 2022-09-03
Update cloud storage emulator host to use http

tdeg:
Currently, the storage emulator host is using HTTPs. This PR changes this to HTTP.


TheKevJames (approved):
Thanks for the contribution, this definitely makes sense! Noting to myself that this will be a breaking change, as it appears several emulators (such as [this one](https://github.com/fsouza/fake-gcs-server) do not run with `http` by default.

TheKevJames:
And [merged](https://github.com/talkiq/gcloud-aio/commit/8d20b2b15167c2b208a1e91842eb3dc9d49973f5). Thanks again!

PR #514, opened 2022-09-29
fix(ci): use nox image for everything

TheKevJames:
Poetry can't be installed in python 3.6. Rather than making that a
special case, use the nox image for everything.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,18 +1,13 @@
 version: 2.1
 
 orbs:
-  linter: talkiq/linter@2
-  poetry: talkiq/poetry@3
+  linter: talkiq/linter@3
+  poetry: talkiq/poetry@4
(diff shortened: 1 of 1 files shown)


leanaha (approved):
Looks good to me but I want to ask you, what is `thekevjames/nox:2022.8.7`.

TheKevJames:
@leanaha that is the base image for https://nox.thea.codes/en/stable/, which I happen to maintain as a personal project. We used to used `nox` as our test runner, but more recently switched away from it to avoid some complexity in local dev. The base image has all python minor versions installed side-by-side, so it's still a fairly convenient one to use in cases where you need multiple pythons. We do for this repo since poetry and various other tools only run on recent python versions, but we still want to be able to build for and test against various older versions for legacy reasons.

PR #518, opened 2022-10-12
fix(auth): Fix Token for stored credentials on Windows

hrichardlee:
Modifies `get_service_data` to be in sync with https://github.com/googleapis/google-auth-library-python/blob/73bc7e9e0e72b6c5057a13cdb4ac996b754ddb58/google/auth/_cloud_sdk.py#L47 I left out the "this should never happen but just in case" case, but happy to include that if desired.

Without this change, using any of the storage APIs on Windows while signed in via the Google Cloud CLI would fail because credentials were not found. With this change, this scenario now works.

Thank you!

### DIFF
--- a/auth/gcloud/aio/auth/token.py
+++ b/auth/gcloud/aio/auth/token.py
@@ -65,9 +65,13 @@ def get_service_data(
     service = service or os.environ.get('GOOGLE_APPLICATION_CREDENTIALS')
     if not service:
         cloudsdk_config = os.environ.get('CLOUDSDK_CONFIG')
-        sdkpath = (cloudsdk_config
-                   or os.path.join(os.path.expanduser('~'), '.config',
-                                   'gcloud'))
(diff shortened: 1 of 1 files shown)


TheKevJames (approved):
Looks great, thank you for the contribution! I'll get this merged and released right away :)

Thanks as well for linking to that method directly. Sounds like it might be a good idea to come up with a way we could test if it falls out of sync again. I'll add that to our backlog.

TheKevJames:
[Tested](https://app.circleci.com/pipelines/github/talkiq/gcloud-aio/2217/workflows/e881cdd0-9763-42d8-a75c-13e22189db2e) and [rebased](https://github.com/talkiq/gcloud-aio/commit/5ae60b22c1b3d66512466a471bf1691ee6c64189). Will release an update shortly :)

hrichardlee:
Thank you!

PR #522, opened 2022-10-19
Use aiohttp with contextmanager

nils-borrmann-y42:
I'm getting lots of errors like this in my logs: `Unclosed client session_client_session: <aiogoogle.sessions.aiohttp_session.AiohttpSession object at 0x7f1cb8188370>`

I believe this is because the aiohttp library is not used correctly. According to the docs, all get/post/.. request methods should always be called with a contextmanager. 

This is just a quick&dirty PR to demonstrate what I mean. Happy to provide a proper PR that changes this for all usages, if you agree with the change.

### DIFF
--- a/pubsub/gcloud/aio/pubsub/publisher_client.py
+++ b/pubsub/gcloud/aio/pubsub/publisher_client.py
@@ -132,10 +132,9 @@ class PublisherClient:
         headers['Content-Length'] = str(len(payload))
 
         s = AioSession(session) if session else self.session
-        resp = await s.post(url, data=payload, headers=headers,
-                            timeout=timeout)
-        data: Dict[str, Any] = await resp.json()
-        return data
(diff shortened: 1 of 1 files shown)


TheKevJames (changes requested):
The problem here would actually relate to session lifecycles rather than session usage. We've documented this [here](https://github.com/talkiq/gcloud-aio/tree/master/storage#usage), specifically:
```
async with Storage() as client:
    # closes the client.session on leaving the context manager

# OR

client = Storage()
# do stuff
await client.close()  # close the session explicitly
```

That said, it took me a while to dig this example up, seems we only included it in the README for _one_ of the packages here rather than all of them. I'll go update those -- all of the `gcloud-aio-*` libraries should be equivalent in this regard!

TheKevJames:
I've included a more detailed section on this in [our generated docs](https://talkiq.github.io/gcloud-aio/gcloud/aio.html#session-management).

PR #541, opened 2022-10-28
fix(datastore): Entity.from_repr(entity.to_repr()) fails if key is None

iamjagan:


iamjagan:
Will fix and resubmit it properly

caseydialpad:
@iamjagan,

Thank you for contributing. Can you elaborate on both the error and how this change fixes it?

TheKevJames:
@iamjagan Thank you for the contribution and LGTM! Merged via 88ed17f41b and will include in the next release.

PR #545, opened 2022-11-17
fallback to GOOGLE_CLOUD_PROJECT env var for project name on local dev

kevin-lee-dialpad:


kevin-lee-dialpad:
for services using bundled services and gcloud-rest-datastore at the same time, we are not allowed to set DATASTORE_PROJECT_ID because setting that will trigger bundled services to use py2 only lib for accessing datastore.

so we need to have other means of identifying the project name and GOOGLE_CLOUD_PROJECT env var does the job. that's the env var that google's dev_appserver.py sets when running app locally with bundled services.

TheKevJames:
Could you please expand on:

> we are not allowed to set DATASTORE_PROJECT_ID because setting that will trigger bundled services to use py2 only lib for accessing datastore.

As far as I understand, that would be intentional here: the `gcloud-aio-datastore` library is mostly aimed at replicating the interface of the various official Google libs and *should* act in the same way as them. Providing a method to make our library work differently than the `gcloud` ones is an anti-goal, as having unnecessary divergent behaviour will lead to confusion in usage.

kevin-lee-dialpad:
I don't think this has anything to do with functionality.
it's just that bundled services use DATASTORE_PROJECT_ID as an indicator to use Cloud Datastore v1.
https://github.com/GoogleCloudPlatform/appengine-python-standard/blob/main/src/google/appengine/datastore/datastore_rpc.py#L2735

and googledatastore client that it uses for Cloud Datastore v1 is python2-only lib.

kevin-lee-dialpad:
so this is a problem on dev because gcloud-rest-datastore uses DATASTORE_PROJECT_ID as the only env var to pull project id from, but for apps that use both gcloud-rest and bundled services(i.e. workflows service in FST), this creates problem because setting that env var will trigger bundled services to use py2 only lib and fail and not setting that would force gcloud-rest-datastore to default to 'dev' as project id, which is a different project id than the one bundled services uses.

TheKevJames:
Triggering the use of datastore v1 seems to be intentional, though? eg. I'm looking at the [datastore emulator docs](https://cloud.google.com/datastore/docs/tools/datastore-emulator#manually_setting_the_variables) and it seems that `DATASTORE_PROJECT_ID` variable is explicitly set when running the emulator ie. working in dev mode.

kevin-lee-dialpad:
well, that maybe have been their intent to set/use DATASTORE_PROJECT_ID for datastore emulator in dev mode, but gcloud-rest-datastore and bundled services run on a different process(aka application process) which has its own set of env vars and needs to be considered separately and in this case, google's bundled services running as part of application process certainly doesn't work on py3 when DATASTORE_PROJECT_ID is set.

we certainly don't know what exactly their intent was or it was their mistake.
it could be that they want DATASTORE_PROJECT_ID to be set for emulator process only OR they just forgot to add py3 support for the old googledatastore lib. I don't know.

either way, I don't think this change is too unreasonable.
GOOGLE_CLOUD_PROJECT is what google sets on both dev(when running app via dev_appserver.py) and appengine environment to indicate the project, so it should work as a fallback.

TheKevJames (approved):
Hmm, given we're only running this block of code when `DATASTORE_EMULATOR_HOST` is truthy anyway, you might be right on the fallback. Good point!

PR #548, opened 2022-11-24
ISSUE-547 Fix SSL verification parameter in auth session

pedrospdc:
The TCPConnector of all available versions of aiohttp expects a `verify_ssl` parameter and not an `ssl` parameter:
https://github.com/aio-libs/aiohttp/blob/v3.8.3/aiohttp/connector.py#L751

Only master has the "ssl" parameter and is unreleased; this breaks local emulators.

### DIFF
--- a/auth/gcloud/aio/auth/session.py
+++ b/auth/gcloud/aio/auth/session.py
@@ -117,7 +117,7 @@ if not BUILD_GCLOUD_REST:
         @property
         def session(self) -> aiohttp.ClientSession:
             if not self._session:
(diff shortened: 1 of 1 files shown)


TheKevJames:
@pedrospdc thank you for the fix! Not quite sure how we let this one slip through. I rebased this for [testing](https://app.circleci.com/pipelines/github/talkiq/gcloud-aio/2243/workflows/3f760518-4e05-4d5f-8362-5613eb47df0d) and got it merged [here](https://github.com/talkiq/gcloud-aio/commit/154e713f96a2cad6fca318cee54e6bcd15ea9698). Will get this released shortly!

PR #551, opened 2022-12-14
chore(pubsub): clean up log messages

TheKevJames:
Main goal: ensure that `hash(message)` is useful: eg. that log messages
can be relied on to be constant for cases where it is "the same error"
and different for cases where it is not. This allows for clients to do
things such as alert when the rate of a given error message is above a
given threshold.

While I was here, I cleaned up the other log strings in the file to make
'em all consistently typed/formatted/etc.

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -29,6 +29,7 @@ else:
                                            float]]
     else:
         MessageQueue = asyncio.Queue
+
     ApplicationHandler = Callable[[SubscriberMessage], Awaitable[None]]
     T = TypeVar('T')
 
@@ -52,8 +53,8 @@ else:
                     self.subscription)
                 self.ack_deadline = float(sub['ackDeadlineSeconds'])
             except Exception as e:
-                log.warning(
-                    'Failed to refresh ackDeadlineSeconds value', exc_info=e)
+                log.warning('failed to refresh ackDeadlineSeconds value',
+                            exc_info=e)
             self.last_refresh = time.perf_counter()
 
         def cache_outdated(self) -> bool:
@@ -91,9 +92,8 @@ else:
             # acknowledge endpoint limit is 524288 bytes
             # which is ~2744 ack_ids
             if len(ack_ids) > 2500:
-                log.error(
-                    'acker is falling behind, dropping %d unacked messages',
-                    len(ack_ids) - 2500)
+                log.error('acker is falling behind, dropping unacked messages',
+                          extra={'count': len(ack_ids) - 2500})
                 ack_ids = ack_ids[-2500:]
                 for _ in range(len(ack_ids) - 2500):
                     ack_queue.task_done()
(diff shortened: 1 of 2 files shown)


pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -438,9 +435,9 @@ async def subscribe(subscription: str,  # pylint: disable=too-many-locals
                                          return_when=asyncio.FIRST_COMPLETED)
             for task in done:
                 task.result()
-            raise Exception('A subscriber worker shut down unexpectedly!')
+            raise Exception('a subscriber worker shut down unexpectedly')
         except (asyncio.CancelledError, Exception) as e:
-            log.info('Subscriber exited', exc_info=e)
+            log.info('subscriber exited', exc_info=e)
```
shaundialpad, line 440: While we're here, thoughts on bumping this to `WARNING`? It has been very useful in the past for debugging shutdowns.

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -124,8 +123,8 @@ async def maybe_ack(ack_id: str) -> None:
                         asyncio.ensure_future(maybe_ack(ack_id))
                     ack_ids = []
 
-                log.warning(
-                    'Ack request failed, better luck next batch', exc_info=e)
+                log.warning('ack request failed, better luck next batch',
```
shaundialpad, line 126: Do you think `better luck next batch` is really necessary? 😂 Perhaps more importantly, it is difficult to infer the meaning of it just from the wording (e.g. are messages dropped or not?). Looks like it is the result of a retryable error, maybe we can make this more clear with a message such as `ack request failed and will be retried` ?

shaundialpad: Actually I guess it is difficult to know the behavior, and it is not the same request that will be retried by the looks of it - depends on how many messages are in the batch. So my wording above also sounds misleading. Maybe fine to leave.

shaundialpad (approved):

pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -105,17 +105,16 @@ async def acker(subscription: str,
...
+                                  'messages may be dropped: %s', e)
 
                     async def maybe_ack(ack_id: str) -> None:
                         try:
                             await subscriber_client.acknowledge(
                                 subscription,
                                 ack_ids=[ack_id])
                         except Exception as ex:
-                            log.warning('Ack failed for ack_id=%s', ack_id,
+                            log.warning('ack failed', extra={'ack_id': ack_id},
```
shaundialpad, line 117: Looks like tests are using `caplog` to grab the specific message, will have to be updated.

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: Yeah, good call. Probably shouldn't be swallowing this stack trace!

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: Yeah, for now I'll leave it since I don't quite understand all the circumstances where wed retry / not / etc, but agreed this could get further improved.

PR #552, opened 2022-12-20
Add Fetch table data method to fetch table rows

parth-crest:
This PR adds support for fetching table rows from the BQ.
https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/list


parth-crest:
Hey @caseydialpad @shaundialpad, We have made this change to enhance the user experience and It would be better if you can start review since we require this change in our use case as well.

TheKevJames (approved):
Thanks for the contribution! Looks good to me.

Rebased for testing [here](https://github.com/talkiq/gcloud-aio/commit/30b901a7243439591f837fea49a5861b12ac74be) (not sure why tests are behaving oddly for external contributors at the moment, I'll figure that out when I can). Merging from there and will get this packed into a release shortly.

PR #554, opened 2022-12-20
feat(deps): drop support for py35 in gcloud-rest

TheKevJames:


eddiedialpad (approved):
approved: drop python `3.5` support for `gcloud-rest` half of the library —

PR #555, opened 2022-12-20
feat(deps): drop support for py36

TheKevJames:


eddiedialpad (approved):
approved: no more support for python `3.6`

PR #539, opened 2022-10-28
Check PUBSUB_EMULATOR_HOST at runtime instead of import time

adriangb:

PR #557, opened 2022-12-22
feat: load emulator hosts at runtime

### LINKED ISSUE #202: variable STORAGE_EMULATOR_HOST
Hey folks,

I'm looking into how to use https://github.com/fsouza/fake-gcs-server to test both googles standard storage client `google.cloud.storage.Storage` and the `gcloud.aio.storage.Storage` client at the same time. `google.cloud.storage.Storage` just made it possible to pass `client_options=ClientOptions(api_endpoint='https://your_storage_emulator_address')` to the client at initialization time while the glcoud-aio relies on a [module wide global](https://github.com/talkiq/gcloud-aio/blob/master/storage/gcloud/aio/storage/storage.py#L40). Would you be open to adding a similar initialization option for the `gcloud.aio.storage.Storage` client? The module wide global is problematic because there's no way to monkeypatch it safely in an async context which means you can only talk to one API_ENDPOINT at a time. Would be happy to help with implementation on this if this sounds reasonable.

Thanks in advance!

TheKevJames:
Solves a couple related goals in a consistent way across all projects:

* allows users to provide an api host via code instead of only via env
  var
* load that value at time of constructing the client rather than import
  time

Thanks to @josipbudzaki in #542 and @adriangb in #539 for providing a
baseline approach.

Fixes #202.

### DIFF
--- a/bigquery/gcloud/aio/bigquery/bigquery.py
+++ b/bigquery/gcloud/aio/bigquery/bigquery.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import os
 from enum import Enum
 from typing import Any
@@ -6,6 +7,7 @@ from typing import AnyStr
 from typing import Dict
 from typing import IO
 from typing import Optional
+from typing import Tuple
 from typing import Union
 
 from gcloud.aio.auth import AioSession  # pylint: disable=no-name-in-module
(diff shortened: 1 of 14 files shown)


eddiedialpad (approved):
approved: emulator URLs are now defined during runtime instead of via env vars

PR #558, opened 2022-12-23
fix(DeprecationWarning): thrown by aiohttp in python3.11

kylepad:
Python 3.11 seems to raise a bunch of warnings as errors that were previously being suppressed such as this one:
```
/usr/local/lib/python3.11/site-packages/gcloud/aio/auth/session.py:121: in session
    connector = aiohttp.TCPConnector(verify_ssl=self._ssl)
/usr/local/lib/python3.11/site-packages/aiohttp/connector.py:776: in __init__
    self._ssl = _merge_ssl_params(ssl, verify_ssl, ssl_context, fingerprint)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

ssl = None, verify_ssl = False, ssl_context = None, fingerprint = None

    def _merge_ssl_params(
        ssl: Union["SSLContext", bool, Fingerprint, None],
        verify_ssl: Optional[bool],
        ssl_context: Optional["SSLContext"],
        fingerprint: Optional[bytes],
    ) -> Union["SSLContext", bool, Fingerprint, None]:
        if verify_ssl is not None and not verify_ssl:
>           warnings.warn(
                "verify_ssl is deprecated, use ssl=False instead",
                DeprecationWarning,
                stacklevel=3,
            )
E           DeprecationWarning: verify_ssl is deprecated, use ssl=False instead

/usr/local/lib/python3.11/site-packages/aiohttp/client_reqrep.py:153: DeprecationWarning
```

While this warning should have showed up under any version of python it is causing 2 tests to break when run under python 3.11, so best to address this now.

### DIFF
--- a/auth/gcloud/aio/auth/session.py
+++ b/auth/gcloud/aio/auth/session.py
@@ -135,7 +135,7 @@ if not BUILD_GCLOUD_REST:
         @property
         def session(self) -> aiohttp.ClientSession:  # type: ignore[override]
             if not self._session:
-                connector = aiohttp.TCPConnector(verify_ssl=self._ssl)
+                connector = aiohttp.TCPConnector(ssl=self._ssl)
 
                 if isinstance(self._timeout, aiohttp.ClientTimeout):
                     timeout = self._timeout
--- a/auth/pyproject.rest.toml
+++ b/auth/pyproject.rest.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "gcloud-rest-auth"
-version = "4.1.2"
+version = "4.1.3"
 description = "Python Client for Google Cloud Auth"
 readme = "README.rst"
 
--- a/auth/pyproject.toml
+++ b/auth/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "gcloud-aio-auth"
-version = "4.1.2"
+version = "4.1.3"
 description = "Python Client for Google Cloud Auth"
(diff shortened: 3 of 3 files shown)


TheKevJames:
Huh, we literally just switched in the other direction in #548, which called out that the `ssl` flag is unreleased.

Looks like we're going to need to do a bit of bisecting in https://github.com/talkiq/gcloud-aio/blob/43bc597c2e96c25d3a277d194c3c1229a27803c7/auth/pyproject.toml#L24 and figure out what version of `aiohttp` _actually_ made the switch over before we can know for sure how to tackle this...

kylepad:
🤔 only master branch?
It's been in their code base for 4 year (see below)

Screen Shot 2023-01-03 at 9 59 16 AM

Although this date appears to just be the refactor that added type hints, I checked the version 3.3.0 tag and it shows it being 5 years old (added in 2017):

Screen Shot 2023-01-03 at 10 02 17 AM

So the `ssl` argument is supported by all versions of `aiohttp` we support

kylepad:
@TheKevJames to be specific the first version tag containing the `ssl` flag is `v3.0.0`: https://github.com/aio-libs/aiohttp/blob/v3.0.0/aiohttp/connector.py#L605

TheKevJames (approved):

PR #564, opened 2023-01-05
fix: use `ssl` instead of `verify_ssl` in aiohttp.TCPConnector

juanamari94:
`aiohttp` is throwing deprecation warnings due to us using `verify_ssl` instead of `ssl`. We're now propagating the parameter of `verify_ssl` into `aiohttp.TCPConnector`'s `ssl` argument.


juanamari94:
Closed in favor of https://github.com/talkiq/gcloud-aio/pull/558

PR #566, opened 2023-01-06
chore(poetry-publish): use pypi-token env variable for `poetry publish`

juanamari94:
Since our `gcloud-aio-*` are considered PyPI critical packages, we need
to switch the way we publish them to use a PyPI token instead of
username and password.

We're adding the names of the environment variables to the matrix that's
used to create the CircleCI jobs and removing the usage of `--username`
and `--password` in `poetry publish`, preferring to set the
environemnt variable `POETRY_PYPI_TOKEN_PYPI` instead. See
[here](https://python-poetry.org/docs/repositories/#publishable-repositories)
for more information.

This change requires setting the following environment variables:
```
- POETRY_PYPI_TOKEN_PYPI_AUTH
- POETRY_PYPI_TOKEN_PYPI_BIGQUERY
- POETRY_PYPI_TOKEN_PYPI_DATASTORE
- POETRY_PYPI_TOKEN_PYPI_KMS
- POETRY_PYPI_TOKEN_PYPI_PUBSUB
- POETRY_PYPI_TOKEN_PYPI_STORAGE
- POETRY_PYPI_TOKEN_PYPI_TASKQUEUE
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,10 +115,16 @@ jobs:
       - attach_workspace:
           at: /rest
       - run:
+          command: |
+            suffix=$(if [ <<parameters.basedir>> = "/rest" ]; then echo "REST"; else echo "AIO"; fi)
+            token=$(printenv | grep -i PYPI_TOKEN_GCLOUD_$(echo $suffix)_<<parameters.cwd>> | awk -F= '{ print $2 }')
+            echo 'export POETRY_PYPI_TOKEN_PYPI=${token}' >> "$BASH_ENV"
+          working_directory: <<parameters.basedir>>/<<parameters.cwd>>
+      - run:
           command: poetry build
           working_directory: <<parameters.basedir>>/<<parameters.cwd>>
       - run:
-          command: poetry publish -u $TWINE_USERNAME -p $TWINE_PASSWORD
+          command: poetry publish
           working_directory: <<parameters.basedir>>/<<parameters.cwd>>
 
   docs:


.circleci/config.yml
```diff
@@ -114,11 +114,17 @@ jobs:
       - checkout
       - attach_workspace:
           at: /rest
+      - run:
+          command: |
+            lower=PYPI_TOKEN_GCLOUD_AIO_<<parameters.cwd>>
```
TheKevJames, line 119: Needs to be different tokens for both `aio` and `rest` as well -- `basedir == /rest` for `gcloud-rest-*` tokens and `basedir == /root/project` for `gcloud-aio-*` ones.

kylepad:
@juanamari94 please lmk when this is deployed so I can retag + redeploy auth 4.1.2 (or I guess you can do it if you want 😛)

.circleci/config.yml
juanamari94: Yeah I just noticed that there's also rest 🤦

juanamari94: Added code to contemplate this

.circleci/config.yml
```diff
@@ -114,11 +114,18 @@ jobs:
       - checkout
       - attach_workspace:
           at: /rest
+      - run:
+          command: |
+            suffix=$(if [ <<parameters.basedir>> = "/rest" ]; then echo "REST"; else echo "AIO"; fi)
+            lower=PYPI_TOKEN_GCLOUD_$(echo $suffix)_<<parameters.cwd>>
+            upper=$(echo $lower | awk '{ print toupper }')
+            export POETRY_PYPI_TOKEN_PYPI=$(printenv | grep $upper | awk -F= '{ print $2 }')
```
jonathan-johnston, line 122: Exporting this variable to the shell env will not persist beyond this `run` block. If it needs to persist into the next `run` blocks, as it should for the `poetry publish` step, then it needs to be pushed into the shell dotfile via `BASH_ENV`:

https://circleci.com/docs/set-environment-variable/#set-an-environment-variable-in-a-shell-command

jonathan-johnston, line 122: You could also write this as a grep with case insensitivity for simplicity:

```shell
suffix=$(if [ <<parameters.basedir>> = "/rest" ]; then echo "REST"; else echo "AIO"; fi)
token=$(printenv | grep -i PYPI_TOKEN_GCLOUD_${suffix}_<<parameters.cwd>> | awk -F= '{ print $2 }')
echo 'export POETRY_PYPI_TOKEN_PYPI=${token}"' >> "$BASH_ENV"
```

jonathan-johnston (changes requested):

.circleci/config.yml
juanamari94: Ohhh I didn't think of using case insensitivity, good idea!

jonathan-johnston (approved):

PR #571, opened 2023-01-10
fix: bad interpolation of token in BASH_ENV

juanamari94:
Turns out that doing `echo '$asd'` with single quotes prints `asd`, whereas using `echo "$asd"` will interpolate the value of `asd` and thus print the value within the variable. We're currently using single quotes instead of double quotes, so we gotta make them single quotes double quotes for that sweet interpolation to work.


kylepad (approved):
good catch!

PR #589, opened 2023-02-24
Add newTransaction option for lookups

nicolasdialpad:
Datastore supports an option in `ReadOptions` called `newTransaction` that, if specified will start a new transaction on the operation. This is a performance optimization that allows to avoid the `beginTransaction` RPC call. This PR implements the possibility to do that for lookups with the respective test.


datastore/gcloud/aio/datastore/options.py
```diff
@@ -0,0 +1,31 @@
...
+        return {'previousTransaction': self.previous_transaction}
+
+
+class TransactionOptions:
+    def __init__(
+            self,
+            read_write: Optional[ReadWrite] = None,
+            read_only: Any = None):
+        self.read_write = read_write
+        self.read_only = read_only  # Type not documented
```
TheKevJames, line 21: Given the following comment in the docs:
> Union field mode can be only one of the following:

I suspect we'll want to make this an `enum.Enum` to avoid needing to handle the neither/both error cases.

I'm thinking something roughly along the lines of this:
```python
class ReadOnly:
    def to_repr(self) -> Dict[str, str]:
        return {}

class ReadWrite:
    def __init__(self, previous_transaction: str):
        self.previous_transaction = previous_transaction

    def to_repr(self) -> Dict[str, str]:
        return {'previousTransaction': self.previous_transaction}

class TransactionOptions:
    def __init__(self, option: Union[ReadOnly, ReadWrite]) -> None:
        self._option = option

    def to_repr(self) -> Dict[str, Dict[str, str]]:
        if isinstance(self.option, ReadOnly):
            return {'readOnly': self.option.to_repr()}
        if isinstance(self.option, ReadWrite):
            return {'readWrite': self.option.to_repr()}
        raise ValueError(f'invalid TransactionOptions {self.option}')
```

Thoughts?

---

Note that in other cases where the docs do not show a JSON representation, that has meant Google requires an empty dictionary as input. That said, it would be fantastic if we could test that assumption!

datastore/gcloud/aio/datastore/datastore.py
```diff
@@ -336,6 +343,22 @@ async def lookup(
                 for k in data.get('deferred', [])
             ],
         }
+        if 'transaction' in data:
+            new_transaction: str = data['transaction']
+            result['transaction'] = new_transaction
+        return result
+
+    def build_read_options(self,
```
TheKevJames, line 351: Hmm, we should probably expose this as a `class ReadOptions` now that there are so many config options here. That said, that'd lead to a breaking change, so I think it's reasonable to implement as-is for now and improve it later. Would you mind adding a note on this in this function? eg.
```
# TODO: expose ReadOptions directly to users
# See https://cloud.google.com/datastore/docs/reference/data/rest/v1/ReadOptions
```
Thanks!

For now, please also rename this method with a leading underscore (eg. to `def _build_read_options()`) to ensure folks don't use this method directly as a public interface.

datastore/gcloud/aio/datastore/options.py
nicolasdialpad: Just as a counter point, it would be little bit confusing when there is some nesting. For example if you were to combine it with `ReadOptions` (which is not exposed right now) and you want a new transaction it would be:

```python
ReadOptions(TransactionOptions(ReadWrite()))
```  

instead of something that could be more clear:

```python
ReadOptions(new_transaction=TransactionOptions(read_write=ReadWrite()))
```

anyway it needs a validation that you don't provide both of them.

TheKevJames (approved):
LGTM, just one minor type hint improvement.

datastore/gcloud/aio/datastore/datastore.py
```diff
@@ -320,9 +324,12 @@ async def lookup(
...
         )
 
-        data: Dict[str, List[Any]] = await resp.json()
+        data: Dict[str, Any] = await resp.json()
 
-        return {
+        return self._build_lookup_result(data)
+
+    def _build_lookup_result(self, data: Dict[str, Any]) -> LookUpResult:
+        result: Dict[str, Union[str, List[Union[EntityResult, Key]]]] = {
```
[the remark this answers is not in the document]
TheKevJames: (fixed this in my cherry-pick https://github.com/talkiq/gcloud-aio/commit/58520efc1ca3a2be64c74e41289e06a20411decc)

TheKevJames:
Merged (https://github.com/talkiq/gcloud-aio/commit/58520efc1ca3a2be64c74e41289e06a20411decc), target release v7.4.0.

PR #593, opened 2023-03-10
feat(pubsub): improve logging

TheKevJames:
Improve some logging:
* move some info logs to debug, as they're generally just noise for most
  users
* avoid logging requested shutdowns as warnings, and log unexpected ones
  as errors


shaundialpad (approved):
Makes sense in theory! These were definitely nice to have when we were debugging issues with GKE pods receiving multiple kill signals and the subscriber getting doubly cancelled (doesn't appear to be happening any longer from what I can tell) and other oddities, but if we ever need to dig into this again I suppose we can always manually enable it.

PR #602, opened 2023-03-23
(storage) Adjustments to how signed urls are generated

kmorey:
- fix the sorting of headers to use the proper lowercase keys
- support domain-named buckets which can't be referenced as subdomains like normal buckets

If a publicly available domain-named bucket was created for testing, we could also test the domain-named buckets, in addition to the currently used `dialpad-oss-public-test` bucket.

### DIFF
--- a/storage/gcloud/aio/storage/blob.py
+++ b/storage/gcloud/aio/storage/blob.py
@@ -150,7 +150,8 @@ class Blob:
         headers = headers or {}
         headers['host'] = f'{self.bucket.name}.{HOST}'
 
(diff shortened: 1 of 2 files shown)


TheKevJames:
@kmorey thanks for the contribution! Re-pushed the branch to double-check the tests were correctly working with our integration creds (https://app.circleci.com/pipelines/github/talkiq/gcloud-aio/2472/workflows/d67c75c0-55b6-4cc6-acfa-a1679ef01efe).

LGTM! Merged in b4620e72a2ff and will get this wrapped up into the next release.

PR #605, opened 2023-04-03
feat(pubsub): support subscription patch operation

TheKevJames:


leanaha (approved):

eddiedialpad (approved):
approved: add pubsub subscription patch method

PR #617, opened 2023-05-11
feat(list-buckets): adding a new method returning buckets from a project

leanaha:
## Summary

- Added `get_buckets()` method in `Storage` class
- Added an integration test using `dialpad-oss` project

I tried to keep all the functionality that the API google docs shows [here](https://cloud.google.com/storage/docs/json_api/v1/buckets/list#parameters)

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -173,21 +173,36 @@ class Storage:
             'Authorization': f'Bearer {token}',
         }
 
-    async def get_buckets(
-        self, project: str, timeout: int = DEFAULT_TIMEOUT,
+    # This method makes the following API call:
+    # https://cloud.google.com/storage/docs/json_api/v1/buckets/list
+    async def list_buckets(
+        self, project: str, *,
+        params: Optional[Dict[str, str]] = None,
+        headers: Optional[Dict[str, Any]] = None,
+        session: Optional[Session] = None,
+        timeout: int = DEFAULT_TIMEOUT,
     ) -> Set[Bucket]:
-        url = ('https://storage.googleapis.com/'
-               f'storage/v1/b?project={project}')
+        url = f'{self._api_root_read}?project={project}'
+        headers = headers or {}
+        headers.update(await self._headers())
+        if not params:
+            params = {'pageToken': ''}
+        s = AioSession(session) if session else self.session
+        buckets = set()
 
-        headers = await self._headers()
-        resp = await self.session.get(url,
-                                      timeout=timeout, headers=headers)
+        while True:
+            resp = await s.get(url, headers=headers,
+                               params=params or {},
+                               timeout=timeout)
 
-        data = await resp.json(content_type=None)
-        result = set()
-        for item in data['items']:
-            result.add(Bucket(self, item['id']))
-        return result
+            content: Dict[str, Any] = await resp.json(content_type=None)
+            for item in content.get('items', []):
+                buckets.add(Bucket(self, item['id']))
+
+            params['pageToken'] = content.get('nextPageToken', '')
+            if not params['pageToken']:
+                break
+        return buckets
 
     def get_bucket(self, bucket_name: str) -> Bucket:
         return Bucket(self, bucket_name)
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -173,16 +173,17 @@ class Storage:
         }
 
     async def list_buckets(
-        self, project: str, timeout: int = DEFAULT_TIMEOUT
+        self, project: str, timeout: int = DEFAULT_TIMEOUT,
     ) -> List[str]:
         token = await self.token.get()
-        url = f'curl -X GET -H "Authorization: Bearer {token}"'
-        url += f' https://storage.googleapis.com/storage/v1/b?project={project}'
+        url = (f'curl -X GET -H "Authorization: Bearer {token}"'
+               ' https://storage.googleapis.com/'
+               f'storage/v1/b?project={project}')
 
-        s = AioSession(session) if session else self.session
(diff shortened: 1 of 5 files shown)


storage/gcloud/aio/storage/storage.py
```diff
@@ -172,6 +173,22 @@ async def _headers(self) -> Dict[str, str]:
             'Authorization': f'Bearer {token}',
         }
 
+    async def get_buckets(
```
TheKevJames, line 176: Looks like you're calling the "list bucket" operation here. `list_buckets()` might be a more accurate name.

storage/gcloud/aio/storage/storage.py
```diff
@@ -172,6 +173,22 @@ async def _headers(self) -> Dict[str, str]:
...
+        self, project: str, timeout: int = DEFAULT_TIMEOUT,
+    ) -> Set[Bucket]:
+        url = ('https://storage.googleapis.com/'
```
TheKevJames, line 179: Please add a comment linking to the docs here, eg. just the following is fine:
```
# https://cloud.google.com/storage/docs/json_api/v1/buckets/list
```

storage/gcloud/aio/storage/storage.py
```diff
@@ -172,6 +173,22 @@ async def _headers(self) -> Dict[str, str]:
...
+               f'storage/v1/b?project={project}')
```
TheKevJames, line 180: We build the API root in the contructor of this class so we can eg. support the GCS emulator and such. Please make use of the pre-built variable:
```suggestion
        url = f'{self._api_root_read}?project={project}'
```

storage/gcloud/aio/storage/storage.py
```diff
@@ -172,6 +173,22 @@ async def _headers(self) -> Dict[str, str]:
             'Authorization': f'Bearer {token}',
         }
 
+    async def get_buckets(
+        self, project: str, timeout: int = DEFAULT_TIMEOUT,
```
TheKevJames, line 177: Looks like there are [several extra optional params here](https://cloud.google.com/storage/docs/json_api/v1/buckets/list#parameters), it'd be awesome to get those included.

storage/gcloud/aio/storage/storage.py
```diff
@@ -172,6 +173,22 @@ async def _headers(self) -> Dict[str, str]:
...
+
+        headers = await self._headers()
+        resp = await self.session.get(url,
+                                      timeout=timeout, headers=headers)
+
+        data = await resp.json(content_type=None)
+        result = set()
+        for item in data['items']:
+            result.add(Bucket(self, item['id']))
+        return result
```
TheKevJames, line 190: The [docs](https://cloud.google.com/storage/docs/json_api/v1/buckets/list#parameters) mention that this AIP is paginated and that the `nextPageToken` value must be checked to get all results.

You should be able to steal the code from our `list_blobs` method [here](https://github.com/talkiq/gcloud-aio/blob/0dc6f57ac6af6f5859c9feeec3d75703ddf39572/storage/gcloud/aio/storage/bucket.py#L64-L82).

storage/gcloud/aio/storage/storage.py
leanaha: The thing with that name is that I'm not returning a `List[]` object. That's why I changed for `get_buckets()`. Are we ok with no returning a `List[]`?

storage/gcloud/aio/storage/storage.py
leanaha: DONE!

storage/gcloud/aio/storage/storage.py
leanaha: Will do

storage/gcloud/aio/storage/storage.py
leanaha: Thanks for the example. I changed it accordingly.

storage/gcloud/aio/storage/storage.py
TheKevJames: Yup, that's fine IMO, though I welcome anyone else's thoughts! You are still calling "the bucket listing API method" in this function, regardless of the exact data type you're returning it as.

storage/gcloud/aio/storage/storage.py
```diff
@@ -172,6 +173,37 @@ async def _headers(self) -> Dict[str, str]:
...
+        params: Optional[Dict[str, str]] = None,
+        headers: Optional[Dict[str, Any]] = None,
+        session: Optional[Session] = None,
+        timeout: int = DEFAULT_TIMEOUT,
+    ) -> Set[Bucket]:
+        url = f'{self._api_root_read}?project={project}'
+        headers = headers or {}
+        headers.update(await self._headers())
+        if not params:
+            params = {'pageToken': ''}
```
TheKevJames, line 189: I'm not 100% sure on the semantics of this API from reading the docs, but given you're setting the pageToken here explicitly, it seems like perhaps this value needs to be provided?

As such, there might be a bug here where if the user provides `params` but the `params` _don't_ include a pageToken, we end up in a broken state?

As such, we might want to prefer doing something like this:
```suggestion
        params = params or {}
        if not params.get('pageToken'):
            params['pageToken'] = ''
```

TheKevJames (approved):

storage/gcloud/aio/storage/storage.py
leanaha: Good reasoning. It convinced me. I will return a `List[]` object which is the one retuned by the google API.

storage/gcloud/aio/storage/storage.py
leanaha: I had to initialize the `params` dict  and set `params[pageToken]` to be `''` because I couldn't  ask for:
```
if not params['pageToken']
```
I'll use your three lines to have a more elegant solution

PR #618, opened 2023-05-17
feat(list-buckets): prereleasing list_buckets() method

leanaha:
## Summary

Prereleasing latest version of gcloud.aio lib which cointains a new method in `storage` class, called `list_buckets()`. This method retrieve bucket names from a gcp project according to this [API call](https://cloud.google.com/storage/docs/json_api/v1/buckets/list)


eddiedialpad (approved):
approved: pre-release `list_buckets()` method — thanks!

juanamari94 (approved):

PR #626, opened 2023-06-19
refactor: migrate to PEP 420 namespace packages

TheKevJames:
## Summary

Finally upgrade to PEP420 namespace packages and remove pkg_resources. Uses
importlib-metadata as a py37 backport stepping-stone.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -95,7 +95,6 @@ repos:
         - backoff
         - types-cryptography
         - types-jwt
-        - types-pkg_resources
         - types-requests
         args:
         - --show-error-codes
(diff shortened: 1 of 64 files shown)


leanaha (approved):
Looks good. One question:
Is It a good idea to have a record of repos supporting both py37 and py38+? Because we will need to remove support for py37 later. Removing support for py37 will tidy up the code.

shaundialpad (approved):

TheKevJames:
@leanaha when I opened this PR, py3.7 was still a non-EOL'd python version; if I'd have had the time to finish it up then, I'd have definitely wanted to leave py3.7 support in -- given this is an OSS library, it's worth supporting additional use-cases and dependencies that we ourselves don't rely on, where it's not a major problem.

That said, given I'm only coming back to this now and py3.7 is EOL... yeah, it'd probably be worth rebasing this after nuking py3.7, as it'll be simpler and we want to drop support for outdated pythons anyway.

PR #628, opened 2023-06-27
fix(auth): reset token acquisition future after failure

shaundialpad:
## Summary

If `self.acquire_access_token()` fails for any reason other than task
cancellation, we currently never detect it and will repeatedly raise
the exception upon waiting for the token.

See https://github.com/talkiq/gcloud-aio/pull/500 for further details of
where this issue was first surfaced but only partially fixed.

### DIFF
--- a/auth/gcloud/aio/auth/token.py
+++ b/auth/gcloud/aio/auth/token.py
@@ -195,7 +195,7 @@ class Token:
         return self.access_token
 
     async def ensure_token(self) -> None:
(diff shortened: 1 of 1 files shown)


all2ham (approved):

TheKevJames:
Aha! Ok, yeah, this makes sense with the current context. Good find!

TheKevJames (approved):

PR #630, opened 2023-07-10
bump(storage-8.3.0): from 8.2.1a0 to 8.3.0

leanaha:
## Summary
I only bump the `storage` version to 8.3.0 in pyproject.toml.

The correct way to do this should be to create an `alpha` version from my branch (not from `master` as I did). After testing it, merge my branch to `master` and create the `minor` tag/release.
In this particular case, I've already merged my branch to master and create the `alpha` version from `master`. Thus, the only thing left to do is to change the pyproject.toml version.

If you want to see what I added to `master` at that time, please [see this link](https://github.com/talkiq/gcloud-aio/commit/d68dcbf62fea7125fc11044142e85a32dc04a999)

### DIFF
--- a/storage/poetry.lock
+++ b/storage/poetry.lock
@@ -105,7 +105,7 @@ python-versions = ">=3.7"
 
 [[package]]
 name = "charset-normalizer"
-version = "3.1.0"
+version = "3.2.0"
 description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
 category = "main"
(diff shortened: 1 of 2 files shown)


jonathan-johnston (changes requested):
In my opinion the creation of a new feature (i.e. `list_buckets`) warrants a minor version bump from `8.2.0` to `8.3.0`.

jonathan-johnston (approved):
Thanks!

Issue #631, opened 2023-08-07
Mishandling of `STORAGE_EMULATOR_HOST` env variable


guneemwelloeux:
Hello,

I've been using the `google-cloud-storage` library for a while with the Firebase Storage emulator, and for that purpose, I've set my `STORAGE_EMULATOR_HOST` env variable to `http://localhost:9000`.

As can be seen here: https://github.com/googleapis/python-storage/blob/dd3d5c850f9f50550d546e77014037caa1f05cfb/google/cloud/storage/_helpers.py#L75
(and following the variable names), it turns out that the official library expects the protocol to be present in the variable (the `http://` part).

However `gcloud-aio-storage` requires to `http://` to be left out, and always add it: https://github.com/talkiq/gcloud-aio/blob/b170cf916f0ad52357bd3457ce5b905dd9170132/storage/gcloud/aio/storage/storage.py#L52

As a result, I cannot use both libraries in my project and have both connected to the same emulator, because they both rely on the same env variable, but expect this variable to be in different and incompatible formats.

Could we change the use of `STORAGE_EMULATOR_HOST` to use the variable as it needs to be defined in the official library?

PR #632, opened 2023-08-07
Accept the official format for `STORAGE_EMULATOR_HOST` env var

guneemwelloeux:
## Summary
In order for this library to be compatible with the official python Google Cloud Storage library, it needs to accept the same value for configuring the `STORAGE_EMULATOR_HOST` environment variable. This will allow users to rely on both libraries in a single project.

Fixes: https://github.com/talkiq/gcloud-aio/issues/631

### DIFF
--- a/storage/gcloud/aio/storage/storage.py
+++ b/storage/gcloud/aio/storage/storage.py
@@ -5,6 +5,7 @@ import json
 import logging
 import mimetypes
 import os
+import warnings
 from typing import Any
 from typing import AnyStr
 from typing import Dict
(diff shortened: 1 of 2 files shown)


storage/gcloud/aio/storage/storage.py
```diff
@@ -50,7 +50,7 @@ def init_api_root(api_root: Optional[str]) -> Tuple[bool, str]:
 
     host = os.environ.get('STORAGE_EMULATOR_HOST')
     if host:
-        return True, f'http://{host}'
+        return True, f'http://{host.removeprefix("http://")}'
```
leanaha, line 53: Why adding `http://` prefix and then removing it from `host`? Can we leave this line as:

```
return True, host
```

leanaha (changes requested):

storage/gcloud/aio/storage/storage.py
guneemwelloeux: Some users might already be using `STORAGE_EMULATOR_HOST` in the "wrong" format, so the change as it is proposed would allow those users to upgrade transparently to the new version.

Otherwise the emulator would not work anymore for these users.

But if that is an acceptable break in compatibility, then I agree, there is no need for that.

leanaha:
Thank you! :-)

guneemwelloeux:
Hello,

When could we expect this to be merged and a new version of the package to be released on PyPI?

Thanks

PR #633, opened 2023-08-16
Unpins `cryptography` in order to expedite CVE patches

chrisguidry:
## Summary

There have been a couple of `cryptography` CVEs (the most recent being
CVE-2023-38325), which often requires upgrading to a new _major_ version of
`cryptography` due to their somewhat quirky versioning scheme.  Having an
upstream dependency like `gcloud-aio-*` that pins an upper version of a
security-critical library like `cryptography` is very inconvenient, and often
requires users to manually install a newer version or wait to patch a CVE.

I'm proposing that we make an exception to the typical requirements style of the
`gcloud-aio` packages (to pin an upper major version) to expedite these kinds of
things.  Anecdotally, I can't recall ever seeing breaking changes in a major
version bump of `cryptography`, especially at its current state of maturity.

### DIFF
--- a/auth/pyproject.rest.toml
+++ b/auth/pyproject.rest.toml
@@ -24,7 +24,7 @@ python = ">= 3.7, < 4.0"
 # aiohttp = ">= 3.3.0, < 4.0.0"
 backoff = ">= 1.0.0, < 3.0.0"
 chardet = ">= 2.0, < 6.0"
-cryptography = ">= 2.0.0, < 42.0.0"
+cryptography = ">=42.0.1"
 pyjwt = ">= 1.5.3, < 3.0.0"
 requests = ">= 2.2.1, < 3.0.0"
 setuptools = ">= 66.0.0, < 67.0.0"  # TODO: upgrade to PEP 420 namespace packages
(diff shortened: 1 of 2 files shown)


TheKevJames:
We generally want to pin to majors to avoid potential breaking changes -- I see from [the cryptography docs](https://cryptography.io/en/latest/api-stability/#deprecation) that they do breaking changes on a rolling three release basis.

My ideal solution would be to pin our upper bound to the major version three releases ahead, which should be a good mix of not causing problems with CVEs without also getting us into trouble with incompatibilities. I'll do this manually for now (eg. pin to < v44.x), but it would be fantastic if that could be automated... our existing dependency update tool doesn't (as far as I know!) support their quirky scheme.

chrisguidry:
@TheKevJames thank you for that, I think that's a good compromise!

chrisguidry:
@TheKevJames Just wanted to thank you again for addressing this, we got notified of another CVE for `cryptography` ([CVE-2024-26130](https://nvd.nist.gov/vuln/detail/CVE-2024-26130)) and I was able to immediately pick up the patch!

Issue #578, opened 2023-01-24
Add support for Workload Identity Federation


jawnsy:
**Summary**

Add support for the `external_account` service account type, which is used in Workload Identity Federation.

**Background**

When running in GitHib Actions, we can use Workload Identity Federation to exchange our GitHub Actions identity token for a corresponding Google Cloud Service Account credential. The credential looks like this:

```json
{
  "type": "external_account",
  "audience": "//iam.googleapis.com/projects/PROJECT_ID/locations/global/workloadIdentityPools/POOL_NAME/providers/PROVIDER_NAME",
  "subject_token_type": "urn:ietf:params:oauth:token-type:jwt",
  "token_url": "https://sts.googleapis.com/v1/token",
  "service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/service-account-name@project-name.iam.gserviceaccount.com:generateAccessToken",
  "credential_source": {
    "url": "https://pipelines.actions.githubusercontent.com/REDACTED_ID/00000000-0000-0000-0000-000000000000/_apis/distributedtask/hubs/Actions/plans/REDACTED_UUID/jobs/REDACTED_UUID/idtoken?api-version=2.0&audience=https%3A%2F%2Fiam.googleapis.com%2Fprojects%2FPROJECT_ID%2Flocations%2Fglobal%2FworkloadIdentityPools%2FPOOL_ID%2Fproviders%2FPROVIDER_ID",
    "headers": {
      "Authorization": "Bearer BIG_JWT_TOKEN"
    },
    "format": {
      "type": "json",
      "subject_token_field_name": "value"
    }
  }
```

[Google describes this as "pluggable credentials"](https://github.com/googleapis/google-auth-library-python/blob/a83af399fe98764ee851997bf3078ec45a9b51c9/google/auth/pluggable.py) and this requires support in your software package.

[Google supports six types for file-based credentials](https://github.com/googleapis/google-auth-library-python/blob/a83af399fe98764ee851997bf3078ec45a9b51c9/google/auth/_default.py#L34-L40):

```python
# Valid types accepted for file-based credentials.
_AUTHORIZED_USER_TYPE = "authorized_user"
_SERVICE_ACCOUNT_TYPE = "service_account"
_EXTERNAL_ACCOUNT_TYPE = "external_account"
_EXTERNAL_ACCOUNT_AUTHORIZED_USER_TYPE = "external_account_authorized_user"
[...]

PR #637, opened 2023-09-14
feat(auth): implement iap auth token

jonathan-johnston:
## Summary

Implement an IAP auth token class `IapToken` which acts in a similar way to our OAuth 2.0 access token class `Token`. I've refactored out the basic components of the two as `BaseToken`, which includes:

* Top-level methods like `get`
* Reading in and interpreting service files
* Refreshing the token before the stated expiry
* General token management
* Handling the `Session`, ex. context manager dunder methods, `close`, etc.

In terms of the `IapToken` class, its general behaviour is:

* Manage tokens for a single IAP-secured service, since IAP is secured on a per-service level
* When running in the Google infrastructure, directly fetch the identity token
* When running with a SA file, fetch the IAP audience (client ID) then fetch the identity token for that IAP audience
* When running as a user, similarly fetch the IAP audience (client ID) before impersonating a service account to generate the identity token

Other changes include:

* Implementing a `HEAD` method on `Session` and piping through `allow_redirects`, for this specific application it aids in explicitly stopping at the `3xx` response to parse out the `location` header
* Implementing a `TokenResponse` dataclass to handle differences between the HTTP response formats of various requests (ex. GCE metadata server for IAP vs. GCE metadata server for OAuth) so that they can all be handled similarly by the `BaseToken` class' token management

Checklist:
- [x] Included references to GCP docs in comments and docstrings
- [x] Tested this manually via alpha version (except for `IapToken` against GCE metadata server; however I verified that the URL is correct through manual testing)


jonathan-johnston:
@TheKevJames Thoughts on building out a thin IAP-secured service on `dialpad-oss` to allow us to write an integration test for `IapToken`?

auth/gcloud/aio/auth/session.py
```diff
@@ -167,6 +175,7 @@ async def get(  # type: ignore[override]
             timeout: Timeout = 10,
             params: Optional[Mapping[str, Union[int, str]]] = None,
             stream: Optional[bool] = None,
+            allow_redirects: bool = False,
```
TheKevJames, line 178: Changing the default for GETs (afaik, this defaults to true in both aiohttp and requests) might cause some breaking changes which could be hard to debug or test for. Thoughts on having this be truthy by default?

Looks like you only need to disable redirects for HEADs, do we need/want to expose this at all here?

auth/gcloud/aio/auth/token.py
```diff
@@ -132,12 +143,20 @@ def get_service_data(
         return {}
 
 
-class Token:
+@dataclass
+class TokenResponse:
+    value: str
+    expires_in: int = GCLOUD_TOKEN_DURATION
```
TheKevJames, line 149: Looks like there are fairly limited places where this default gets used, all within the new IAP flow. It might be a bit easier to reason about if this value is passed in explicitly in that case, since (if I'm reading this correctly!) it looks like there might be two values at play here: previously, we passed this value in and Google created tokens of the designated duration. With IAP, though, it seems like we're finding the tokens have this duration even though we aren't passing this value in (for gce metadata and authorized users)?

auth/gcloud/aio/auth/token.py
```diff
@@ -267,18 +326,158 @@ async def acquire_access_token(self, timeout: int = 10) -> None:
...
+    async def _refresh_authorized_user(
+        self, iap_client_id: str,
+        timeout: int,
+    ) -> TokenResponse:
+        """
+        Fetch IAP ID token by impersonating a service account.
+
+        https://cloud.google.com/iap/docs/authentication-howto#obtaining_an_oidc_token_in_all_other_cases
+        """
+        async def get_access_token(timeout: int) -> str:
```
TheKevJames, line 390: nit: consider de-nesting this method to allow us to more easily test it on its own.

auth/gcloud/aio/auth/token.py
```diff
@@ -267,18 +326,158 @@ async def acquire_access_token(self, timeout: int = 10) -> None:
...
+        return TokenResponse(value=content['id_token'],
+                             expires_in=expiry - int(time.time()))
+
+    async def refresh(self, *, timeout: int) -> TokenResponse:
+        if self.token_type == Type.AUTHORIZED_USER:
+            iap_client_id = await self._get_iap_client_id(timeout=timeout)
+            resp = await self._refresh_authorized_user(
+                iap_client_id, timeout=timeout)
+        elif self.token_type == Type.GCE_METADATA:
+            resp = await self._refresh_gce_metadata(timeout=timeout)
```
TheKevJames, line 475: Huh, how interesting that the client ID isn't necessary in this flow!

TheKevJames (commented):
Please update the [usage docs](https://github.com/talkiq/gcloud-aio/blob/ea736ce8b9b7dda5f07c845a2e75fef5e28a2c7c/auth/gcloud/aio/auth/__init__.py#L14) to include the IapToken workflow.

TheKevJames:
> Thoughts on building out a thin IAP-secured service on dialpad-oss to allow us to write an integration test for IapToken?

I like the idea in principle, but this repo has a problem with integration tests that we've not yet solved: external contributors can't auth correctly, so tests fail for their PRs. It'd be nice if we could avoid introducing more "failures" like that until we can solve the underlying problem, I think...

The long term solution is probably to solve #578 and use workload identity via CircleCI to authenticate so it isn't tied to org secrets, but that's a much bigger task here. I wonder if mocking out the integration pieces would be a good enough test in the meantime?

That said, I guess e2e tests which only work for us are still better than no e2e tests at all...

auth/gcloud/aio/auth/session.py
jonathan-johnston: That's a good point and thanks for catching that, for some reason when I read the documentation I thought it stated that `GET`s have `allow_redirects=False` by default, but that's only for `HEAD`.

Agreed, I'm okay removing `allow_redirects` for `GET` since we don't have any use case for it so KISS applies here.

auth/gcloud/aio/auth/session.py
```diff
@@ -79,6 +79,14 @@ async def delete(
     ) -> Response:
         pass
 
+    @abstractmethod
+    async def head(
+        self, url: str, headers: Optional[Mapping[str, str]],
+        timeout: float, params: Optional[Mapping[str, Union[int, str]]],
+        allow_redirects: bool,
+    ) -> Response:
+        pass
```
caseydialpad, line 88: Perhaps https://docs.python.org/3/library/exceptions.html#NotImplementedError?

auth/gcloud/aio/auth/session.py
```diff
@@ -219,6 +229,21 @@ async def delete(  # type: ignore[override]
             await _raise_for_status(resp)
             return resp
 
+        async def head(  # type: ignore[override]
+            self, url: str,
+            headers: Optional[Mapping[str, str]] = None,
+            timeout: Timeout = 10,
+            params: Optional[Mapping[str, Union[int, str]]] = None,
+            allow_redirects: bool = False,
+        ) -> aiohttp.ClientResponse:
```
caseydialpad, line 238: This is imported as `from aiohttp import ClientResponse as Response`. Perhaps change?

auth/gcloud/aio/auth/token.py
```diff
@@ -132,12 +143,20 @@ def get_service_data(
...
+@dataclass
+class TokenResponse:
+    value: str
+    expires_in: int = GCLOUD_TOKEN_DURATION
+
+
+class BaseToken:
+    """GCP auth token base class."""
     # pylint: disable=too-many-instance-attributes
+    __metaclass__ = ABCMeta
```
caseydialpad, line 155: Do you have any good references on __metaclass__ and working with metaclasses in Python? Interest only.

auth/gcloud/aio/auth/token.py
```diff
@@ -209,27 +222,72 @@ async def ensure_token(self) -> None:
             self.acquire_access_token())
         await self.acquiring
 
-    async def _refresh_authorized_user(self, timeout: int) -> Response:
+    @abstractmethod
+    async def refresh(self, *, timeout: int) -> TokenResponse:
+        pass
```
caseydialpad, line 227: Same as https://docs.python.org/3/library/exceptions.html#NotImplementedError.

auth/gcloud/aio/auth/token.py
caseydialpad: I am with Kevin here, unless there are compelling reasons not to.

caseydialpad (commented):
I have added my comments inline.

auth/gcloud/aio/auth/token.py
jonathan-johnston: I'll go with the majority on this one and de-nest this one, but I personally find inline functions to be great for readability when a chunk of code doesn't deserve its own method but represents a block of functionality which serves a specific purpose. I.e. it can be represented as an input/output contract via a function.

auth/gcloud/aio/auth/token.py
jonathan-johnston: Agreed!

auth/gcloud/aio/auth/token.py
jonathan-johnston: For some reason the responses from IAP token requests are all over the place and they don't provide you with an expiry. Maybe there's some way to force them to, but in practice the tokens do expire within 1 hour.

I've moved this default TTL to the class-level and removed the default from this `TokenResponse` dataclass. LMK what you think!

auth/gcloud/aio/auth/token.py
```diff
@@ -267,18 +326,158 @@ async def acquire_access_token(self, timeout: int = 10) -> None:
...
+class IapToken(BaseToken):
+    """An OpenID Connect ID token for a single IAP-secured service."""
+
+    default_token_ttl = 3600
+
+    def __init__(
+        self, app_uri: str,
+        service_file: Optional[Union[str, IO[AnyStr]]] = None,
+        session: Optional[Session] = None,
+        service_account: Optional[str] = None,
```
shaundialpad, line 341: noob question, why do we need both `service_file` and `service_account` here?

shaundialpad: Ah sorry I see:

```
`service_account`: an optional string denoting a GCP service account which
  takes the form of an email address. Only valid (and required!) for
  authentication with a project's authorized users. [Impersonating a service
  account][impersonating-sa] is required when generating an ID token in this
  case.
  ```

Which service account should be supplied here?

**Review by shaundialpad — APPROVED** [review]
Looks pretty good to me! More tests would be phenomenal but I saw Kevin's comment re: integration tests and I see we don't have much at the moment. As long as this had been tested again for the various auth modes for both Token and IapToken then I'm happy with this.

File: auth/gcloud/aio/auth/token.py (hunk 9 above)
Comment (TheKevJames) [reply auth/gcloud/aio/auth/token.py:341]: nit: would renaming this to `impersonate_service_account` or similar be clearer?

**Review by TheKevJames — COMMENTED** [review]
doc updates LGTM, but you'll need to md->rst them and move the README updates into the __init__.py file as well once you rebase. Had to switch that as part of the pdoc->sphinx changeover.

File: auth/gcloud/aio/auth/token.py (hunk 9 above)
Comment (jonathan-johnston) [reply auth/gcloud/aio/auth/token.py:341]: Funny you mention this, I originally had this argument named `impersonating_service_account` 😛

File: auth/gcloud/aio/auth/session.py (hunk 5 above)
Comment (jonathan-johnston) [reply auth/gcloud/aio/auth/session.py:88]: That could definitely work, although I would like to limit the sprawl of my changes and this is an extension of existing form. Since this is an abstract base class and this is an abstract method, you can't instantiate an object of this class so raising is redundant:

```python
>>> import abc
>>> class Base(abc.ABC):
...     def fn(self):
...         return 42
...     @abc.abstractmethod
...     def abstract(self):
...         pass
... 
>>> b = Base()
TypeError: Can't instantiate abstract class Base with abstract method abstract
```

auth/gcloud/aio/auth/session.py
jonathan-johnston: Same as the previous comment about wanting to avoid changing unrelated parts of the code.

@TheKevJames might be able to tell you more about this choice.

auth/gcloud/aio/auth/token.py
jonathan-johnston: The only place that has served me well on this has been the official Python docs unfortunately! I'm happy to field questions if you have specific ones :)

https://docs.python.org/3/library/abc.html

auth/gcloud/aio/auth/token.py
jonathan-johnston: Ditto for my comment above.
https://github.com/talkiq/gcloud-aio/pull/637#discussion_r1333407708

caseydialpad (approved):

PR #646, opened 2023-10-11
fix(auth): use iap aud for gce metadata server

jonathan-johnston:
## Summary

The official documentation is wrong on this, we need to fetch the ID token from the GCE metadata server using the typical audience value instead of the URI as stated here:
https://cloud.google.com/docs/authentication/get-id-token#metadata-server


jonathan-johnston:
@TheKevJames I believe that was actually a different use case, using the user auth method with SA impersonation. I reproduced the failure in staging with the metadata server ID token fetch, so at least it _should_ fail consistently.

shaundialpad (approved):
classic GOOG

PR #653, opened 2023-11-15
perf(auth): avoid unnecessary token refresh on get_project()

TheKevJames:
## Summary

Perf optimization: for GCE_METADATA tokens, there's no hit the token refresh
endpoint when we're only looking to call `get_project()` and have the value set
locally.


shaundialpad (approved):
LGTM!

cphoward (approved):

PR #663, opened 2023-12-18
feat(storage): blob upload with content type

dodeca-6-tope:
Propagate content type when uploading using a blob object.


dodeca-6-tope:
image

TheKevJames (approved):
Sorry for the delay! Thanks, looks good, will release ASAP.

PR #665, opened 2023-12-22
feat(auth): add support for service account impersonation

### LINKED ISSUE #421: Support for Impersonated Credentials
I am interested in using gcloud-aio-storage, but I also need to use impersonated service account credentials. My current code works with the non-asyncio google-cloud-storage package, which in turn utilizes the google-auth package. Here is a direct reference to the bit of that auth lib I am currently using: https://google-auth.readthedocs.io/en/master/reference/google.auth.impersonated_credentials.html

My question here is 1) is there a way to utilize SA impersonation today with gcloud-aio-storage? and 2) if not, would it be possible to add support for this given the current gcloud-aio library architecture?

m1racoli:
## Summary

Added the arguments `target_principal` and `delegates` which optionally define an impersonation chain for the given target principal to the Token class and related clients.

The request to generate the impersonated access token is made via the IAM credentials API method [projects.serviceAccounts.generateAccessToken](https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken).

Test skeletons have been added. Those tests are currently skipped as
they require additional setup of the testing environment with additional
service accounts which are to be impersonated:

- `sa_a` to be impersonated by the base service account
- `sa_b` to be impersonated by `sa_a` via delegation

closes #421

### DIFF
--- a/bigquery/gcloud/aio/bigquery/bigquery.py
+++ b/bigquery/gcloud/aio/bigquery/bigquery.py
@@ -6,6 +6,7 @@ from typing import Any
 from typing import AnyStr
 from typing import Dict
 from typing import IO
+from typing import List
 from typing import Optional
 from typing import Tuple
 from typing import Union
@@ -70,12 +71,16 @@ class BigqueryBase:
             service_file: Optional[Union[str, IO[AnyStr]]] = None,
             session: Optional[Session] = None, token: Optional[Token] = None,
             api_root: Optional[str] = None,
+            target_principal: Optional[str] = None,
+            delegates: Optional[List[str]] = None,
     ) -> None:
         self._api_is_dev, self._api_root = init_api_root(api_root)
         self.session = AioSession(session)
         self.token = token or Token(
             service_file=service_file, scopes=SCOPES,
             session=self.session.session,  # type: ignore[arg-type]
+            target_principal=target_principal,
+            delegates=delegates,
         )
 
         self._project = project
--- a/auth/gcloud/aio/auth/__init__.py
+++ b/auth/gcloud/aio/auth/__init__.py
@@ -63,6 +63,15 @@ The ``Token`` constructor accepts the following optional arguments:
   ``Token.close()`` method to ensure the session is cleaned up appropriately.
 * ``scopes``: an optional list of GCP `scopes`_ for which to generate our
   token. Only valid (and required!) for `service account`_ authentication.
+* ``target_principal``: The service account to generate the access token for.
+  The **roles/iam.serviceAccounts.getAccessToken** role on that service account
+  is required.
+* ``delegates``: The sequence of service accounts in a delegation chain.
+  This field is required for delegated requests.
+  Each service account must be granted the **roles/iam.serviceAccountTokenCreator**
+  role on its next service account in the chain. The last service account in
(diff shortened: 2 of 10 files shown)


m1racoli:
I understand that I added changes beyond the `auth` package. Let me know if you'd like this PR to focus only on the `auth` package.

shahar1 (approved):
LGTM, This PR is highly anticipated in the Apache Airflow community :)

m1racoli:
Thanks! :) Do we have an ETA for this to get merged or is something particular blocking the PR?

Lee-W:
Hi @TheKevJames , we're wondering whether gcloud-aio is interested in this feature and if so, is there any blocker of suggestions in this PR? Thanks!

bigquery/gcloud/aio/bigquery/bigquery.py
```diff
@@ -70,12 +71,16 @@ def __init__(
             service_file: Optional[Union[str, IO[AnyStr]]] = None,
             session: Optional[Session] = None, token: Optional[Token] = None,
             api_root: Optional[str] = None,
+            target_principal: Optional[str] = None,
```
melugoyal, line 74: do we need a corresponding change in bigquery.Job (and the other BigQuery classes) to pass in these values to the BigqueryBase constructor? https://github.com/talkiq/gcloud-aio/blob/fcfb54895de44f385a31bfd3de2bb4be2bea3fb7/bigquery/gcloud/aio/bigquery/job.py#L22-L32

kaxil:
Gentle nudge @TheKevJames @shaundialpad

TheKevJames:
Hey all, sorry for the radio silence on this one, taking a look now! Thanks for posting this.

TheKevJames:
Ok, I've merged this ([here](https://github.com/talkiq/gcloud-aio/commit/19b0c27c95b9d1a92a7b4ea2334ba2097029470a)) and will be releasing shortly. Thanks for the contribution! Overall this looked great, just needed to tweak some linting things. Thank you especially for the integration test skeleton, that was very easy to work with.

TheKevJames:
Oh, just to add -- I needed to pull out the changes to other libraries to avoid order-of-operations, but will introduce them later. For now, you can build a Token instance with the delegation and pass it to any other library class, you just can't let it pass through those values.

Now that the Token class has gotten so many params, I'm starting to think it might be worth making it mandatory to have users create the Token, frankly -- especially with how many places we would need to plumb everything, as you pointed out for BQ Jobs. I'll look into this and see about pushing out a follow-up.

m1racoli:
Thank you

> Oh, just to add -- I needed to pull out the changes to other libraries to avoid order-of-operations, but will introduce them later. For now, you can build a Token instance with the delegation and pass it to any other library class, you just can't let it pass through those values.

Yeah, that makes sense.

> Now that the Token class has gotten so many params, I'm starting to think it might be worth making it mandatory to have users create the Token, frankly -- especially with how many places we would need to plumb everything, as you pointed out for BQ Jobs. I'll look into this and see about pushing out a follow-up.

Yes, I had similar thoughts. You either provide a token or assume the default credentials. That would align with the methodology in google libs where the credentials object is a container for every aspect of the credentials and needs to be configured as such. I also wonder if the project ID needs to be part of the token.

Issue #584, opened 2023-02-06
Unable to read location parameter other then 'US'  for the Job object


VladaZakharova:
Hi Team!
After some testing, it seems like it is not possible to construct Job object as such (https://github.com/talkiq/gcloud-aio/blob/master/bigquery/gcloud/aio/bigquery/job.py#L21) with location other then 'US'. 

The Job object was created from the existing job_id, so later i can use this Job object for async calls, as such:

```
Job(job_id=job_id, project=project_id, service_file=f, session=cast(Session, session))
```
However, trying to create a Job for the datasets with location other then 'US' and then starting this Job using method result() from the standard library (https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.job.QueryJob#google_cloud_bigquery_job_QueryJob_result) fails with error:

```
404, message='Not Found: {\n  "error": {\n    "code": 404,\n    "message": "Not found: Job ***-system-tests-303516:AAA_c5e382f968ca9480f7408e1167672ca8",\n    "errors": [\n      {\n        "message": "Not found: Job ***-system-tests-303516:AAA_c5e382f968ca9480f7408e1167672ca8",\n        "domain": "global",\n        "reason": "notFound"\n      }\n    ],\n    "status": "NOT_FOUND"\n  }\n}\n', url=URL('https://www.googleapis.com/bigquery/v2/projects/***-system-tests-303516/jobs/AAA_c5e382f968ca9480f7408e1167672ca8')
```

After some debugging i can see that there is no possibility to specify location directly when creating Job object. Could you please add some documentation what exactly location will be used? The docs for the BigQuery says that the location for the jobs should be taken from the dataset, for which the Job was created:
https://cloud.google.com/bigquery/docs/locations#specify_locations

In comparison, for the sync version of Job (QueryJob, LoadJob and etc) there is a possibility to specify location parameter while creating:
https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.job.QueryJob#google_cloud_bigquery_job_QueryJob_location

Possible resolution:
As a solution, I would suggest adding location parameter to the url that you are using for the API call here -> https://github.com/talkiq/gcloud-aio/blob/master/bigquery/gcloud/aio/bigquery/job.py#L67
So that new link would be as follows:

```
"https://www.googleapis.com/bigquery/v2/projects/project_id/jobs/job_id?location=asia-south1"
```

[...]

PR #658, opened 2023-11-29
fix(bigquery) Add location for Big Query job

manhld0206:
### What changed
- Include `params` for `_post_json()` and `_delete()` function of `BigqueryBase` class
- Include `location params` for `Job` class
  - Add utils function `_config_params` to add `location` parameter if needed
  - Include `location` parameter for BigQuery API if needed

### Why
- Related Issue: https://github.com/talkiq/gcloud-aio/issues/584
- The following BigQuery API cloud require `location` parameter if job's location is not US or EU multiple region
  - https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/get#query-parameters
  - https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults#query-parameters
  - https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/cancel#query-parameters
  - https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/delete#query-parameters

PR #666, opened 2023-12-24
fix(bigquery) Add location param to BigQuery Job

spencertollefson:
Bigquery Job has been missing the location param.
Sometimes this param is required, as can be seen
in the link above the get_job() func definition.

## Summary
I added a small code change here, which, if I understand correctly, should fix the issue described in #584 .

I tried to make some unit tests, and frankly felt a little over my head so I stopped (between lack of other tests, asyncio, and making network calls to GCP). I thought I'd get the PR up and could start a discussion if we can try merging this in without unit tests - let me know.

EDIT: I was also thinking this PR could be used to expand the Job() class to allow for all of the query parameters that Google has as options in their API for the various Job methods. Let me know if that is something to be implemented here, or saved for another PR.

Originally I was notified of this issue by this issue in the Airflow repository:
https://github.com/apache/airflow/issues/35833

### DIFF
--- a/bigquery/gcloud/aio/bigquery/job.py
+++ b/bigquery/gcloud/aio/bigquery/job.py
@@ -23,9 +23,10 @@ class Job(BigqueryBase):
             self, job_id: Optional[str] = None, project: Optional[str] = None,
             service_file: Optional[Union[str, IO[AnyStr]]] = None,
             session: Optional[Session] = None, token: Optional[Token] = None,
-            api_root: Optional[str] = None,
+            api_root: Optional[str] = None, location: Optional[str] = None,
     ) -> None:
         self.job_id = job_id
+        self.location = location
         super().__init__(
             project=project, service_file=service_file,
             session=session, token=token, api_root=api_root,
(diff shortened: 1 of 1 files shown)


bigquery/gcloud/aio/bigquery/job.py
```diff
@@ -65,6 +66,8 @@ async def get_job(
 
         project = await self.project()
         url = f'{self._api_root}/projects/{project}/jobs/{self.job_id}'
+        if self.location:
+            url += f'?location={self.location}'
 
         return await self._get_url(url, session, timeout)
```
moiseenkov, line 72:
```suggestion
        params = {"location": self.location} if self.location else None:

        return await self._get_url(url, session, timeout, params)
```

moiseenkov:
@overset , @kmorey , @bhearsum , hi,
Could you please take a look at this PR and merge it if possible?

spencertollefson:
@moiseenkov  - I appreciate the review, but I think @manhld0206 has a better PR here addressing the same issue:
https://github.com/talkiq/gcloud-aio/pull/658

I'll close my PR, and steer you and the other PRs to that one.

Issue #45, opened 2018-11-20
Upload as gzip


JensMadsen:
gzipping files before upload could lead to a significant performance boost. see: 
- https://cloud.google.com/storage/docs/json_api/v1/how-tos/performance
- https://cloud.google.com/storage/docs/transcoding

- [ ] `Content-Encoding: gzip` must set  in headers
- [ ] when use `gzip = True` -> zip data before upload (https://docs.python.org/3/library/gzip.html or use built-in gzip of aiohttp?)
- Not sure how this would work with streams?

TheKevJames:
Good idea! Just like in [this comment](https://github.com/talkiq/gcloud-aio/issues/41#issuecomment-440078798) I think we'd want to provide a:
```
Storage.upload(..., zipped: Optional[bool] = None)
```
where:
- `zipped = False` - current behaviour
- `zipped = True` - run gzip, set `content-encoding`, etc
- `zipped = None` (default) - allow us to chose, which we might do based on file size etc.

TheKevJames:
Support has been added in `gcloud-*-storage` v9.3.0!

PR #698, opened 2024-03-15
feat(storage): Add support for gzipping files before uploading them

fbalboaDialpad:
Adds support for gzipping files before uploading them to GCS. Fixes https://github.com/talkiq/gcloud-aio/issues/45


auth/gcloud/aio/auth/session.py
```diff
@@ -335,7 +338,9 @@ async def get(
             timeout: float = 10,
             params: Optional[Mapping[str, Union[int, str]]] = None,
             stream: bool = False,
+            auto_decompress: Optional[bool] = None,
         ) -> Response:
+            del auto_decompress  # Not used in SyncSession
```
TheKevJames, line 343: We're generally very cautious on introducing functionality which varies between the sync and aio versions of this library. It looks like both `requests` and `aiohttp` automatically decompress by default, but only the latter lets you disable the functionality.

We should do one of the following:
* remove this flag, eg. such that both libraries act the same and always decompress
* throw an explicit error to users if `auto_decompress != True` in the sync version of the library

I am leaning quite heavily toward the former of those options, but if you've got a strong use-case for needing the libraries to act differently, we could consider the latter. Thoughts?

storage/gcloud/aio/storage/storage.py
```diff
@@ -485,18 +490,58 @@ def _get_stream_len(stream: IO[AnyStr]) -> int:
...
+        else:
+            raise TypeError(f'unsupported upload type: "{type(data)}"')
 
-        if isinstance(data, bytes):
-            return io.BytesIO(data)
-        if isinstance(data, str):
-            return io.StringIO(data)
-        if isinstance(data, io.IOBase):
-            return data  # type: ignore[return-value]
+        if gzip_compress:
```
TheKevJames, line 535: You should be able to keep the type hint breakages simpler by doing this as a second pass: eg. instead of doing this block within `_preprocess_data()`, we can just do it as a second step in `upload()` and avoid modifying this function at all.

So instead of how you have:
```
stream = self._preprocess_data(file_data, gzip_compress=zipped)
```
We can do:
```
stream = self._preprocess_data(file_data)
if zipped:
    stream = self._compress_file_in_chunks(stream)
```

storage/gcloud/aio/storage/storage.py
```diff
@@ -485,18 +490,58 @@ def _get_stream_len(stream: IO[AnyStr]) -> int:
             stream.seek(current)
 
     @staticmethod
-    def _preprocess_data(data: Any) -> IO[Any]:
+    def _compress_file_in_chunks(input_stream: Union[IO[AnyStr], io.IOBase],
+                                 output_stream: io.IOBase,
```
TheKevJames, line 494: Rather than passing in a mutable object, please create this locally and return it, eg.
```
def _compress_file_in_chunks(input_stream: Union[IO[AnyStr], io.IOBase], chunk_size: int = 8192) -> None:
    output_stream = io.BytesIO()
    # your existing method
    return output_stream
```

storage/gcloud/aio/storage/storage.py
```diff
@@ -485,18 +490,58 @@ def _get_stream_len(stream: IO[AnyStr]) -> int:
...
+        with gzip.open(output_stream, 'wb') as f_out:
+            while True:
+                chunk = input_stream.read(chunk_size)
+                if not chunk:
+                    break
+                if isinstance(chunk, str):
+                    chunk_bytes = chunk.encode('utf-8')
+                else:
+                    chunk_bytes = chunk
+                # At this point we assume that chunk is a bytes object
```
TheKevJames, line 510: Rather than assuming, we can verify this by pre-declaring our type hint.

```suggestion
                chunk_bytes: bytes
                if isinstance(chunk, str):
                    chunk_bytes = chunk.encode('utf-8')
                else:
                    chunk_bytes = chunk
```

TheKevJames (changes requested):

storage/gcloud/aio/storage/storage.py
fbalboaDialpad: Even with the type hint, it would still be an assumption. This is because the `io.IOBase` interface doesn't have a `read` method with a type hint that specified that the return type is either `str` or `bytes`, even though all the implementations I'm aware of return one of those 2. I find this part of Python's typing system and its inconsistency with the `typing.IO` classes very confusing.

From the docs:

> Even though [IOBase](https://docs.python.org/3/library/io.html#io.IOBase) does not declare read() or write() because their signatures will vary, implementations and clients should consider those methods part of the interface. Also, implementations may raise a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) (or [UnsupportedOperation](https://docs.python.org/3/library/io.html#io.UnsupportedOperation)) when operations they do not support are called.

auth/gcloud/aio/auth/session.py
fbalboaDialpad: From a client's standpoint, their code doesn't need to change, because of the `if not auto_decompress and BUILD_GCLOUD_REST:` clause introduced in `_download`. This let's you access the compressed bytes when using the `requests` lib. I verified with the integration tests and I also manually and it works as expected. By passing `auto_decompress` to `blob.download()` you are just indicating to our lib that you want the compressed file.

I think this is a totally valid use case, you could very well want to grab a gzip file from GCS and store it zipped somewhere else for serving or later processing. Automatically decompressing just to compress it again is wasteful.

storage/gcloud/aio/storage/storage.py
fbalboaDialpad: By moving `_compress_file_in_chunks` out of `_preprocess_data` I was able to change the typing hints so `input_stream` is now a `IO[AnyStr]`, which does bound the return type of `read`

auth/gcloud/aio/auth/session.py
TheKevJames: Aha, I had missed that `requests` avoids decompressing if you do a streaming download, in which case you are entirely correct that we can coerce the behavior into being the same, so this looks good as-is.

TheKevJames:
Overall looks good! Split this into #713 and #714 so we can release auth before storage, and applied a couple minor changes (headers can be lowercased, some type hints were a bit broad / incorrectly disabled, added a warnings for auto_decompress and stream mismatching in gcloud-rest-auth) and am working to get it merged and released. Thanks!

fbalboaDialpad:
> Overall looks good! Split this into #713 and #714 so we can release auth before storage, and applied a couple minor changes (headers can be lowercased, some type hints were a bit broad / incorrectly disabled, added a warnings for auto_decompress and stream mismatching in gcloud-rest-auth) and am working to get it merged and released. Thanks!

Awesome, thanks for taking care of that!

PR #713, opened 2024-04-04
feat(auth): support auto-decompress argument passthrough

TheKevJames:
## Summary

Subset of talkiq/gcloud-aio#698, split to let me release `auth` before
`storage`.

Modified to be explicit on usage warning for `auto_decompress` and `stream`
flag mismatch in `gcloud-rest-auth`.


auth/gcloud/aio/auth/session.py
```diff
@@ -335,7 +339,17 @@ async def get(
             timeout: float = 10,
             params: Optional[Mapping[str, Union[int, str]]] = None,
             stream: bool = False,
+            auto_decompress: bool = True,
         ) -> Response:
+            if not auto_decompress and not stream:
+                warnings.warn(
+                    'the requests library always decompresses responses when '
+                    'outside of streaming mode; when audo_decompress is '
```
shaundialpad, line 347: s/audo_decompress/auto_decompress/

shaundialpad: Also can you elaborate on why we want a warning here and not raising an error? Perhaps to maintain backwards compatibility?

PR #718, opened 2024-04-06
fix(storage): avoid stomping over session-level auto_decompress value

TheKevJames:
## Summary

Blocked on #717


storage/gcloud/aio/storage/storage.py
```diff
@@ -591,12 +591,14 @@ async def _download(
         # aiohttp and requests automatically decompress the body if this
         # argument is not passed. We assume that if the Accept-Encoding header
         # is present, then the client will handle the decompression
-        auto_decompress = 'accept-encoding' not in {k.lower() for k in headers}
+        auto_decompress = None  # inherit the Session default setting
+        if 'accept-encoding' not in {k.lower() for k in headers}:
```
fbalboaDialpad, line 595:
```suggestion
        if 'accept-encoding' in {k.lower() for k in headers}:
```

eddiedialpad (approved):
approved: avoid overriding auto_decompress value

storage/gcloud/aio/storage/storage.py
cphoward: I believe we want want Kevin has here. If `'accept-encoding' in {k.lower() for k in headers}` is `True`, that means that the session passes `auto_decompress` value implicitly in `session`, and we want `None`. We only want to set `auto_decompress` if `'accept-encoding' in {k.lower() for k in headers}` is False, that is `'accept-encoding'` is not in the headers.

cphoward (approved):

juanamari94:
:(

PR #723, opened 2024-04-11
Allow external account type

### LINKED ISSUE #722: `external_account` is not a valid Token enum value
I'm using the gcloud-aio-kms package locally and in production without any issues, thanks for that!

I'm now also trying to run my integration tests in Github Actions. i am using OIDC to authenticate via my github repo to act as a specific service account with permissions that are required. It looks like this approach returns some unexpected `Token` values. `external_account` is apparently not a valid enum value.

Thought I'd at least make you aware of the issue right now. I hope I will find the time soon to provide a PR to fix this, if this is still an issue by that time!

```
 File "/home/runner/work/payments/payments/.venv/lib/python3.11/site-packages/gcloud/aio/auth/token.py", line 166, in __init__
    |     self.token_type = Type(self.service_data['type'])
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/nix/store/2m6gdzz77ikkc5xwp3iiqi901339frp7-python3-3.11.8-env/lib/python3.11/enum.py", line 712, in __call__
    |     return cls.__new__(cls, value)
    |            ^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/nix/store/2m6gdzz77ikkc5xwp3iiqi901339frp7-python3-3.11.8-env/lib/python3.11/enum.py", line 1135, in __new__
    |     raise ve_exc
    | ValueError: 'external_account' is not a valid Type
```

bjgbeelen:
## Summary
This should fix #722 , but not sure if it is enough

### DIFF
--- a/auth/gcloud/aio/auth/token.py
+++ b/auth/gcloud/aio/auth/token.py
@@ -74,6 +74,7 @@ class Type(enum.Enum):
     AUTHORIZED_USER = 'authorized_user'
     GCE_METADATA = 'gce_metadata'
     SERVICE_ACCOUNT = 'service_account'
+    EXTERNAL_ACCOUNT = 'external_account'
 
 
 def get_service_data(


TheKevJames (changes requested):
Appreciate the contribution, but unfortunately this won't be enough -- we actually need to build the code which knows how to authenticate as an external account. See, for example, the `Token.refresh()` method, which would need to be updated to include a new workflow for refreshing said new type of token.

It sounds like (from reading #722) you've got a test case set up? You should be able to install `gcloud-aio-auth` from this branch for testing. Alternatively, if you could point us in the direction of that test case, it would be great if we could add that approach to our integration tests directly.

PR #760, opened 2024-06-10
feat(datastore): Support "IN" query filter

onurdialpad:
## Summary
Add support for "IN" filter so that Firestore in Datastore mode would be able to execute single query with IN filter instead of multiple parallel queries.

### DIFF
--- a/datastore/gcloud/aio/datastore/constants.py
+++ b/datastore/gcloud/aio/datastore/constants.py
@@ -53,6 +53,7 @@ class PropertyFilterOperator(enum.Enum):
     LESS_THAN_OR_EQUAL = 'LESS_THAN_OR_EQUAL'
     NOT_EQUAL = 'NOT_EQUAL'
     UNSPECIFIED = 'OPERATOR_UNSPECIFIED'
+    IN = 'IN'
 
 
 class ResultType(enum.Enum):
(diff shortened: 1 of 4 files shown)


datastore/gcloud/aio/datastore/constants.py
```diff
@@ -53,6 +53,7 @@ class PropertyFilterOperator(enum.Enum):
     LESS_THAN_OR_EQUAL = 'LESS_THAN_OR_EQUAL'
     NOT_EQUAL = 'NOT_EQUAL'
     UNSPECIFIED = 'OPERATOR_UNSPECIFIED'
+    IN = 'IN'
```
TheKevJames, line 56: Please update the comment at the top of this class (if it's trivial to add `NOT_IN` as well, that'd be appreciated!)

nit: please keep this enum sorted.

datastore/gcloud/aio/datastore/filter.py
```diff
@@ -113,8 +115,13 @@ def from_repr(cls, data: Dict[str, Any]) -> 'PropertyFilter':
...
 
     def to_repr(self) -> Dict[str, Any]:
-        return {
+        rep = {
             'op': self.operator.value,
-            'property': {'name': self.prop},
-            'value': self.value.to_repr(),
+            'property': {'name': self.prop}
         }
+        # Temporary workaround for handling arrayValue with PropertyFilter
```
TheKevJames, line 122: Could you please explain what's temporary about this?

onurdialpad: It looks we handle `to_repr` in 2 different ways for GQL and standard filter cases. The way we handle GQL case adds `arrayValue` by itself [here](https://github.com/talkiq/gcloud-aio/blob/master/datastore/gcloud/aio/datastore/value.py#L81) which looks better to me. Here I add a basic type check but It might be better to call appropriate one _maybe_  doing type inferring (I haven't investigated it deeply). That's the reason why I call this as "workaround"

TheKevJames:
(force merging -- tests pass on branch, but not for the PR -- [details](https://app.circleci.com/pipelines/github/talkiq/gcloud-aio?branch=in_filter))

PR #765, opened 2024-06-13
fix(auth): use the full format for the JWT payload in IAPToken

juanamari94:
## Summary

Some of the GCP products when retrieving the JWT token from the GCE
Metadata Server come with the `email` claim, but some have recently
started not returning that claim as part of the JWT payload. Adding the
`format=full` query parameter to the GCE Metdata Server API seems to
return the `email` claim for these cases, along with other GCE Metadata
which for the time being we do not need and thus, we discard.

[See these docs for more
information.](https://cloud.google.com/compute/docs/instances/verifying-instance-identity#request_signature)

### DIFF
--- a/auth/pyproject.toml
+++ b/auth/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "gcloud-aio-auth"
-version = "5.3.1"
+version = "5.3.2"
 description = "Python Client for Google Cloud Auth"
 readme = "README.rst"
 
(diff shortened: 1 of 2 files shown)


auth/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "gcloud-aio-auth"
-version = "5.3.1"
+version = "5.3.2a0"
```
shaundialpad, line 3: Friendly non alpha reminder

shaundialpad (approved):
Nice find!

auth/pyproject.toml
eddiedialpad: ➕ HOSTILE PLUS ONE REPLY COMMENT

eddiedialpad (approved):
✅ approved: fix authentication

TheKevJames (approved):
Good find!

PR #815, opened 2024-09-19
fix(pubsub): fix unhandled exception in subscriber task

w-miller:
## Summary

We have seen unhandled exceptions making their way to the root unhandled exception handler of the event loop in our application that uses pubsub subscribe(), as in the below Traceback. Whilst our application catches the relevant aiohttp errors when awaiting the subscribe() call, this exception remained unhandled. This is because subscribe was ignoring the tasks returned from the calls to asyncio.wait(), meaning any exception results on these tasks would not be handled and the tasks would get deleted when subscribe() returned.

Fix this by gathering the results of all worker tasks before returning from subscribe(). Given that we're about to raise
asyncio.CancelledError() and return, there's no point in raising the internal errors here, so just log them.

```
  File "/usr/lib/python3/dist-packages/gcloud/aio/pubsub/subscriber.py", line 403, in producer
    new_messages = await asyncio.shield(pull_task)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3/dist-packages/gcloud/aio/pubsub/subscriber.py", line 426, in producer
    new_messages += await asyncio.wait_for(pull_task, 5)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/tasks.py", line 479, in wait_for
    return fut.result()
           ^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/gcloud/aio/pubsub/subscriber_client.py", line 157, in pull
    resp = await s.post(
           ^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/gcloud/aio/auth/session.py", line 190, in post
    resp = await self.session.post(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/aiohttp/client.py", line 560, in _request
    await resp.start(conn)
  File "/usr/lib/python3/dist-packages/aiohttp/client_reqrep.py", line 899, in start
    message, payload = await protocol.read()  # type: ignore[union-attr]
                       ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/aiohttp/streams.py", line 616, in read
    await self._waiter
aiohttp.client_exceptions.ServerDisconnectedError: Server disconnected
```

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -542,14 +542,33 @@ else:
 
         for task in producer_tasks:
             task.cancel()
-        await asyncio.wait(producer_tasks, return_when=asyncio.ALL_COMPLETED)
+        producer_done, _ = await asyncio.wait(
+            producer_tasks,
+            return_when=asyncio.ALL_COMPLETED,
+        )
 
         for task in consumer_tasks:
             task.cancel()
-        await asyncio.wait(consumer_tasks, return_when=asyncio.ALL_COMPLETED)
+        consumer_done, _ = await asyncio.wait(
+            consumer_tasks,
+            return_when=asyncio.ALL_COMPLETED,
+        )
 
         for task in acker_tasks:
             task.cancel()
-        await asyncio.wait(acker_tasks, return_when=asyncio.ALL_COMPLETED)
+        acker_done, _ = await asyncio.wait(
+            acker_tasks,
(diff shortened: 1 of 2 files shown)


TheKevJames:
Good catch, this makes sense! Thank you for the contribution.

Issue #685, opened 2024-02-23
Unable to use impersonated credentials with `gcloud.aio.storage`


and3rson:
I need to impersonate a SA in order to be able to sign GCS URLs locally as follows:

```sh
gcloud auth application-default login --impersonate-service-account=EMAIL
```

However, instantiating `gcloud.aio.storage.Client` with such credentials fails:

```py
  File "/REDACTED/google.py", line 163, in __init__
    self.client = Storage()
                  ^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/gcloud/aio/storage/storage.py", line 168, in __init__
    self.token = token or Token(
                          ^^^^^^
  File "/usr/local/lib/python3.11/site-packages/gcloud/aio/auth/token.py", line 265, in __init__
    super().__init__(service_file=service_file, session=session)
  File "/usr/local/lib/python3.11/site-packages/gcloud/aio/auth/token.py", line 166, in __init__
    self.token_type = Type(self.service_data['type'])
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/enum.py", line 712, in __call__
    return cls.__new__(cls, value)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/enum.py", line 1128, in __new__
    raise ve_exc
ValueError: 'impersonated_service_account' is not a valid Type
```

TheKevJames:
We currently support impersonation via the `target_principal` and `delegates` Token arguments ([see docs](https://talkiq.github.io/gcloud-aio/autoapi/auth/index.html)), but I'm not familiar offhand with the workflow which leads to the `service_data` containing the impersonation details. Would you be able to upload a (redacted, of course) copy of the SA json key file you're attempting to load? I suspect we'll just need to parse out a few different fields in that case.

zintrepid:
I had this exact issue, and I made a PR with a proposed solution: https://github.com/talkiq/gcloud-aio/pull/831.

zintrepid:
For reference, here is what the redacted SA json key looks like:
```json
[...]

PR #831, opened 2024-11-06
feat(auth): support service account impersonation from gcloud json file

zintrepid:
Added support for type 'impersonated_service_account' with fields 'source_credentials' and 'service_account_impersonation_url' which gcloud will use when you login with the flag --impersonate-service-account

Proposed solution for https://github.com/talkiq/gcloud-aio/issues/685


zintrepid:
This feature would be very useful for my workflow! I'm not an expert in python or GCP, so any suggestions or edits are welcomed. I appreciate your time reviewing this PR.

TheKevJames (approved):
Thanks for the fix! Overall, this looks great -- we'll aim to include it in the next `auth` release. Much appreciated!

PR #834, opened 2024-12-02
feat(auth): support the standard gce metadata host env

lanbochen-anyscale:
## Summary
We have the use case such that we want to override the default gce compute metadata and proxy to other compute nodes (outside of gce) for them to be able to access gcloud APIs. Add `GCE_METADATA_HOST` and `GCE_METADATA_ROOT` support which is google's standard env vars (see https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/compute_engine/_metadata.py#L45).


TheKevJames (approved):
Awesome, thanks for the contribution!

PR #835, opened 2024-12-03
feat(datastore): Support plain old list arg with IN filter

onurdialpad:
## Summary

This PR adds support to IN filters with plain old list such as
`query.filter('names IN', ['aaa', 'bbb'])`


TheKevJames (approved):
This was already supported via the Array type, eg. since we generally try to match with the Google builtin types.

That said, I don't see any particular reason we couldn't support this syntactic sugar, sure thing.

PR #840, opened 2024-12-18
Adding support for and id_token result using a service account

alexdialpad:
- ensure_token will now not block everything after 50% of TTL.
- Token class now able to deal with an `id_token` return payload.

### DIFF
--- a/auth/gcloud/aio/auth/token.py
+++ b/auth/gcloud/aio/auth/token.py
@@ -160,7 +160,18 @@ class BaseToken:
     def __init__(
         self, service_file: Optional[Union[str, IO[AnyStr]]] = None,
         session: Optional[Session] = None,
+        cache_preempt_after: float = 0.5,
+        cache_refresh_after: float = 0.95,
     ) -> None:
+        if cache_preempt_after <= 0 or cache_preempt_after > 1:
+            raise ValueError('cache_preempt_after must be a value between 0 and 1')
+        if cache_refresh_after <= 0 or cache_refresh_after > 1:
+            raise ValueError('cache_refresh_after must be a value between 0 and 1')
+        # Portion of TTL after which a background refresh would start
+        self.cache_preempt_after = cache_preempt_after
+        # Portion of TTL after which a cached token is considered invalid
+        self.cache_refresh_after = cache_refresh_after
+
         self.service_data = get_service_data(service_file)
         if self.service_data:
             self.token_type = Type(self.service_data['type'])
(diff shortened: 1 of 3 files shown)


auth/gcloud/aio/auth/token.py
```diff
@@ -324,8 +355,9 @@ async def _refresh_service_account(self, timeout: int) -> TokenResponse:
             timeout=timeout,
         )
         content = await resp.json()
-        return TokenResponse(value=str(content['access_token']),
-                             expires_in=int(content['expires_in']))
+        token = str(content.get('access_token') or content.get('id_token'))
```
TheKevJames, line 358: I think we want a KeyError if neither of these exist, since AFAIK that would still be invalid. Something like this?
```suggestion
        token = str(content.get('access_token') or content['id_token'])
```

auth/gcloud/aio/auth/token.py
```diff
@@ -238,6 +264,11 @@ async def acquire_access_token(self, timeout: int = 10) -> None:
         self.access_token_duration = resp.expires_in
         self.access_token_acquired_at = datetime.datetime.now(
             datetime.timezone.utc)
+        base_timstamp = self.access_token_acquired_at.timestamp()
+        self.access_token_preempt_after = int(
+            base_timstamp + (resp.expires_in * self.cache_preempt_after))
+        self.access_token_refresh_after = int(
+            base_timstamp + (resp.expires_in * self.cache_refresh_after))
```
TheKevJames, line 271: nit typo:
```suggestion
        base_timestamp = self.access_token_acquired_at.timestamp()
        self.access_token_preempt_after = int(
            base_timestamp + (resp.expires_in * self.cache_preempt_after))
        self.access_token_refresh_after = int(
            base_timestamp + (resp.expires_in * self.cache_refresh_after))
```

auth/gcloud/aio/auth/token.py
```diff
@@ -160,7 +160,20 @@ class BaseToken:
     def __init__(
         self, service_file: Optional[Union[str, IO[AnyStr]]] = None,
         session: Optional[Session] = None,
+        cache_preempt_after: float = 0.5,
+        cache_refresh_after: float = 0.95,
```
TheKevJames, line 164: nit: I think "cache" is the wrong word here. What do you think about `background_refresh_after` and `force_refresh_after`, maybe?

Also occurs to me that in the `gcloud-rest` version of this library, there is no background/preempt version, so whichever is the lower of these values just becomes the forced/immediate refresh. So maybe it'd be even better to name accordingly: `async_refresh_after` and `sync_refresh_after`.

Thoughts? Trying to make usage as clear as possible in both cases, but I'm kind of blanking on wording which makes sense for both flows.

TheKevJames (commented):
Overall flow looks great, only some small/nit comments. Feel free to tackle whatever subset of the nits you'd like, I can also futz around with naming/etc after the fact.

auth/gcloud/aio/auth/token.py
alexdialpad: I like the first suggestion, and specifically `background_refresh_after` is spot on imo. However, in the `gcloud-rest` case, it make zero sense.
Having said that, having two thresholds doesn't make any sense in that case altogether, since it's whichever is first forces a blocking refresh. No idea how to name intricate details of a mechanism that is pure obfuscation in `gcloud-rest`.

TheKevJames (approved):

juanamari94:
hello

PR #869, opened 2025-02-18
fix(storage): Storage copy for large files

deeuu:
## Summary

Issue: Storage copies fail for large files, generally occurring when objects are transferred across buckets in different regions or across GCP projects, i.e. where the transfer is costly, requiring multiple calls to `rewriteTo`.  Both timeout and server disconnection errors are common.

Summary of changes: Added missing `metadata_` when making subsequent posts to the `rewriteTo` API, as part of `aio.storage.Storage.copy`.


TheKevJames (approved):
Thank you!

PR #875, opened 2025-03-19
feat(pubsub): clean up logging

TheKevJames:
## Summary

Clean up our general approach to logging:
* include an `exc_message` extra wherever an exception is relevant to
  the log, to allow for message-level aggregation (eg. grouping errors
  by raised exception)
* avoid using string templates in exception messages, to make
  aggregation easier (`message` is always a constant string for any
  given error case, so it can be alerted on)
* downgrade application callback exceptions to a warning, since that can
  be a normal part of user flows -- if users want it logged at the error
  level, they can always do that themselves, we shouldn't be enforcing
  it

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber.py
@@ -63,7 +63,9 @@ else:
                 log.warning(
                     'failed to refresh ackDeadlineSeconds value',
                     exc_info=e,
+                    extra={'exc_message': str(e)},
                 )
+
             self.last_refresh = time.perf_counter()
 
         def cache_outdated(self) -> bool:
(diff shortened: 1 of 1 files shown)


pubsub/gcloud/aio/pubsub/subscriber.py
```diff
@@ -149,8 +157,9 @@ async def maybe_ack(ack_id: str) -> None:
                     ack_ids = []
 
                 log.warning(
-                    'ack request failed, better luck next batch',
+                    'ack request failed',
```
eddiedialpad, line 160: 💟 bahahaah yaaaa ... it was a bit ... passive aggressive 🇨🇦🇨🇦🇨🇦 sorry

eddiedialpad (approved):
✅ Approved: logging practices in pubsub

pubsub/gcloud/aio/pubsub/subscriber.py
TheKevJames: Hahaha yeah, some of this was just "lets make our error messages a bit less cutesy" :P

cphoward (approved):

PR #883, opened 2025-04-02
Fix 'Token' object has no attribute 'scopes' error

tgwizard:
If we don't do this, Token won't always have a scopes attribute, and sometimes (e.g. for service account keys) we'd get errors like

```
'Token' object has no attribute 'scopes'
```

Better to get an error that insufficient scopes are used (if that's the error).


TheKevJames (approved):
Ah, great catch, thanks!

TheKevJames:
Lint issue, so I had to cherry-pick along with a lint fix. Now in mainline, will be released as a patch bump shortly! Thanks again :)

PR #898, opened 2025-05-23
feat(pubsub): allow disabling ack deadline fetch

shaundialpad:
## Summary
When the user uses our `subscribe` feature, we will always attempt to fetch the
ack deadline configured for that subscription (even if the first call fails, we
will do this repeatedly until we succeed). This operation counts towards the
Administrator Operations quota, which can potentially cause issues if the user
is running many other instances.

Here, we introduce a new optional parameter ``ack_deadline`` which the user can
specify (in conjunction with ``ack_deadline_cache_timeout = float('inf')`` which
is the default) to avoid the library making any attempts to fetch or refresh the
specified deadline.

Example:

```python
    await subscribe(
        'projects/<my_project>/subscriptions/<my_subscription>',
        handler,
        subscriber_client,
        ack_deadline=10,
    )
```

will not attempt to fetch the configured value for the ack deadline from the
subscription.

### DIFF
--- a/pubsub/pyproject.rest.toml
+++ b/pubsub/pyproject.rest.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "gcloud-rest-pubsub"
-version = "6.1.0"
+version = "6.2.0"
 description = "Python Client for Google Cloud Pub/Sub"
 readme = "README.rst"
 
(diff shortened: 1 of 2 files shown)


eddiedialpad (approved):
✅ approved: expose ack_deadline to be user configurable

tonydialpad (approved):
LGTM

mauroparis (approved):

PR #906, opened 2025-06-06
feat(auth): external credential type in auth

rakshasa-1729:
## Summary
Related to https://github.com/talkiq/gcloud-aio/issues/578

Adds support for the external_account service account type, which is used in Workload Identity Federation.

### DIFF
--- a/auth/gcloud/aio/auth/token.py
+++ b/auth/gcloud/aio/auth/token.py
@@ -1,6 +1,7 @@
 """
 Google Cloud auth via service account file
 """
+
 import datetime
 import enum
 import json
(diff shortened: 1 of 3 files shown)


rakshasa-1729:
@TheKevJames Could use some help triggering the integration tests once

auth/gcloud/aio/auth/token.py
```diff
@@ -409,47 +478,178 @@ async def _refresh_service_account(self, timeout: int) -> TokenResponse:
...
+            expires_in=data.get('expires_in', self.default_token_ttl),
+        )
+
+    async def _get_subject_token(
+        self, credential_source: Dict[str, Any], timeout: int
+    ) -> str:
+        """Get the subject token from the credential source."""
+        source_type = credential_source.get('type')
+        if not source_type:
+            raise ValueError('Credential source missing type')
```
cagataygurturk, line 587: Hi, the source_type property is not mandatory as far as I understand. When we download a configuration file, this is what we get:

````yaml
{
  "universe_domain": "googleapis.com",
  "type": "external_account",
  "audience": "WORKLOAD_IDENTITY_PROVIDER_ID",
  "subject_token_type": "urn:ietf:params:oauth:token-type:jwt",
  "token_url": "https://sts.googleapis.com/v1/token",
  "credential_source": {
    "file": "/var/run/secrets/kubernetes.io/serviceaccount/token",
    "format": {
      "type": "text"
    }
  }
}
````

cagataygurturk (changes requested):

PR #911, opened 2025-07-03
feat(pubsub): add support for snapshots

santiagofdialpad:
## Summary

Include methods to work with snapshots in `gcloud-aio-pubsub`.
the methods included are:
- seek
- create_snapshot
- delete_snapshot
- list_snapshots
- get_snapshot

### DIFF
--- a/pubsub/gcloud/aio/pubsub/subscriber_client.py
+++ b/pubsub/gcloud/aio/pubsub/subscriber_client.py
@@ -66,6 +66,10 @@ class SubscriberClient:
     def topic_path(cls, project: str, topic: str) -> str:
         return f'{cls.project_path(project)}/topics/{topic}'
 
+    @classmethod
+    def snapshot_path(cls, project: str, snapshot: str) -> str:
+        return f'{cls.project_path(project)}/snapshots/{snapshot}'
+
     async def _headers(self) -> Dict[str, str]:
         headers = {
             'Content-Type': 'application/json',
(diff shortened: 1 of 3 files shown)


pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -245,6 +249,107 @@ async def list_subscriptions(
...
+        timeout: int = 10,
+    ) -> None:
+        """
+        Seeks a subscription to a point in time or to a given snapshot.
+        """
+        url = f'{self._api_root}/{subscription}:seek'
+        headers = await self._headers()
+        encoded = json.dumps(body).encode()
+        s = AioSession(session) if session else self.session
+        await s.post(url, data=encoded, headers=headers, timeout=timeout)
```
santiagofdialpad, line 266: Not sure how we usually handle cases like this. Is it better to receive a body, or to use optional parameters like:

```
async def seek(
    self,
    subscription: str,
    *,
    snapshot: Optional[str] = None,
    time: Optional[str] = None,
    timeout: int = 10,
) -> None:
```

Note that only one of the two can be sent in the request

olidp: by "one of the two" what do you mean? which two?

pubsub/gcloud/aio/pubsub/subscriber_client.py
```diff
@@ -245,6 +249,107 @@ async def list_subscriptions(
...
+    ) -> Dict[str, Any]:
+        """
+        List snapshots
+        """
+        url = f'{self._api_root}/{project}/snapshots'
+        headers = await self._headers()
+        s = AioSession(session) if session else self.session
+
+        all_results: Dict[str, Any] = {'snapshots': []}
+        nextPageToken = None
```
shaundialpad, line 321: nit: snake case

shaundialpad (approved):

pubsub/gcloud/aio/pubsub/subscriber_client.py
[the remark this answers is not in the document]
santiagofdialpad: It seems so https://github.com/talkiq/gcloud-aio/blob/de77f6cf4beaf97beb804bfd8761cefcca189429/auth/gcloud/aio/auth/session.py#L180-L195

pubsub/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "gcloud-aio-pubsub"
-version = "6.2.0"
```
olidp, line 3: remember to remove alpha version

olidp (approved):

PR #919, opened 2025-07-21
perf(auth): improve latency by disabling netrc within requests lib

armenfirespotter:


armenfirespotter:
@TheKevJames sorry to waste your time on this one - I was able to get what I needed by other means

PR #922, opened 2025-07-23
feat(datastore): implement query explain

MarshallXie216:
## Summary

Added query explain support. Passing a valid `ExplainOptions` to `runQuery` will return a `QueryResult` container class with `ExplainMetrics`, which can help with debugging and optimizing queries.

## Changes

- Added query explain container classes that mirror the REST API: `QueryResult`, `ExplainMetrics`, `PlanSummary`, `ExecutionStats`, `IndexDefinition`, `ExplainOptions`
- Modified `runQuery` method to Datastore client to return a `QueryResult` container class that has `QueryResultBatch` and `ExplainMetrics`. This allows it to handle both regular and explain queries (default and analyze mode).
- Added some simple unit and integration tests

@talkiq/engineering @cphoward @TheKevJames - requesting codeowners approval to merge.

### DIFF
--- a/datastore/gcloud/aio/datastore/datastore.py
+++ b/datastore/gcloud/aio/datastore/datastore.py
@@ -22,9 +22,9 @@ from .entity import EntityResult
 from .key import Key
 from .mutation import MutationResult
 from .query import BaseQuery
+from .query import QueryResult
 from .query import QueryResultBatch
 from .query_explain import ExplainOptions
-from .query_explain import QueryResult
 from .transaction_options import TransactionOptions
 from .value import Value
 
@@ -352,13 +352,12 @@ class Datastore:
             result['transaction'] = new_transaction
         return result
 
+    # https://cloud.google.com/datastore/docs/reference/data/rest/v1/ReadOptions
     def _build_read_options(self,
                             consistency: Consistency,
                             newTransaction: Optional[TransactionOptions],
                             transaction: Optional[str]) -> Dict[str, Any]:
         # TODO: expose ReadOptions directly to users
-        # See
-        # https://cloud.google.com/datastore/docs/reference/data/rest/v1/ReadOptions
         if transaction:
             return {'transaction': transaction}
 
@@ -423,11 +422,9 @@ class Datastore:
         timeout: int = 10,
     ) -> QueryResult:
 
-        # get request endpoint
         project = await self.project()
         url = f'{self._api_root}/projects/{project}:runQuery'
 
-        # build request payload
         if transaction:
             options = {'transaction': transaction}
         else:
@@ -453,7 +450,6 @@ class Datastore:
             'Content-Type': 'application/json',
         })
 
-        # send the request
         s = AioSession(session) if session else self.session
         resp = await s.post(
             url, data=payload, headers=headers,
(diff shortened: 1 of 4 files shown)


datastore/gcloud/aio/datastore/query_explain.py
```diff
@@ -0,0 +1,266 @@
...
+        return debug_stats
+
+    @classmethod
+    def from_repr(cls, data: Dict[str, Any]) -> 'ExecutionStats':
+        return cls(
+            results_returned=int(data.get('resultsReturned', 0)),
+            execution_duration=cls._parse_execution_duration(
+                data.get('executionDuration')),
+            read_operations=int(data.get('readOperations', 0)),
+            debug_stats=cls._parse_debug_stats(data.get('debugStats', {}))
```
mwaehnerdialpad, line 162: The `_parse_debug_stats` method is mutating the input dictionary here instead of returning a copy. But not an issue if this is intended

mwaehnerdialpad (approved):
LGTM, great work!

Looks like we will need the maintainers to manually trigger the integration tests for us

https://github.com/talkiq/gcloud-aio/blob/master/.github/CONTRIBUTING.rst#pr-tests

datastore/gcloud/aio/datastore/query_explain.py
MarshallXie216: I decided to not create a copy since we don't really do anything else with the debug_stats dict.

datastore/gcloud/aio/datastore/datastore.py
```diff
@@ -409,45 +412,95 @@ async def rollback(
         s = AioSession(session) if session else self.session
         await s.post(url, data=payload, headers=headers, timeout=timeout)
 
-    # https://cloud.google.com/datastore/docs/reference/data/rest/v1/projects/runQuery
-    async def runQuery(
+    # pylint: disable=too-many-locals
+    async def _build_and_send_query(
```
santiagofdialpad: We can probably remove the comments on this function!

datastore/gcloud/aio/datastore/datastore.py
```diff
@@ -409,45 +412,95 @@ async def rollback(
...
+        data = await self._build_and_send_query(
+            query=query,
+            explain_options=explain_options,
+            transaction=transaction,
+            consistency=consistency,
+            session=session,
+            timeout=timeout
+        )
+
+        return self.query_explain_result_kind.from_repr(data)
```
santiagofdialpad: Maybe I’m missing something here.. but why don’t we include the `explain_options` optional parameter in the original `runQuery` function and change the result type to something like `QueryResult` (which I think is the same structure you already have for `QueryExplainResult`)?

datastore/gcloud/aio/datastore/query_explain.py
```diff
@@ -0,0 +1,266 @@
...
+        query_scope = ''
+        properties = []
+
+        if 'query_scope' in data:
+            query_scope = data['query_scope']
+        if 'properties' in data:
+            properties_str = data['properties']
+            properties = re.findall(r'\s*([^\s,()]+)\s+(ASC|DESC)\s*',
+                                    properties_str,
+                                    flags=re.IGNORECASE)
```
santiagofdialpad, line 68: nit: this could be compiled as a class attribute to avoid compiling it every time we use it

PR #925, opened 2025-08-08
feat(datastore): added newTransaction option for runQuery

mwaehnerdialpad:
## Summary

Similar to https://github.com/talkiq/gcloud-aio/pull/589 for `lookup`, this PR adds the `newTransaction` param to `runQuery`, which is a performance optimization that allows to avoid the beginTransaction RPC call. This changes `runQuery`’s return object, making it a breaking change that requires a major version bump.

Also added a missing setup step in the contributing guide for running the integration tests.

### DIFF
--- a/.github/CONTRIBUTING.rst
+++ b/.github/CONTRIBUTING.rst
@@ -24,6 +24,15 @@ And run tests with:
     poetry run pytest tests/unit  # unit tests only
     poetry run pytest tests/integration  # integration tests only
 
+To run the tests that require authentication with Google, you should set the
+GOOGLE_APPLICATION_CREDENTIALS environment variable to point to your credentials
+JSON file. If you are using the ``gcloud auth application-default login`` command,
+they can be found in ``~/.config/gcloud/application_default_credentials.json``.
+
+.. code-block:: console
+
+    export GOOGLE_APPLICATION_CREDENTIALS="$HOME/.config/gcloud/application_default_credentials.json"
+
 ``aio`` and ``rest``
 ~~~~~~~~~~~~~~~~~~~~
 
--- a/datastore/gcloud/aio/datastore/query.py
+++ b/datastore/gcloud/aio/datastore/query.py
@@ -261,3 +261,36 @@ class QueryResultBatch:
             data['snapshotVersion'] = self.snapshot_version
 
         return data
+
+
+class RunQueryResult:
+    def __init__(self, batch: QueryResultBatch,
+                 transaction: Optional[str] = None):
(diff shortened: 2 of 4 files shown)


datastore/tests/integration/smoke_test.py
```diff
@@ -128,6 +128,53 @@ async def test_start_transaction_on_lookup(creds: str,
         actual = await ds.lookup([key], session=s)
         assert actual['found'][0].entity.properties == {'animal': 'aardvark'}
 
+        # Clean up test data
+        await ds.delete(key, s)
```
MarshallXie216, line 132: good catch on this! We should probably check if there are any other tests that are leaving dirty state.

datastore/gcloud/aio/datastore/query.py
```diff
@@ -261,3 +261,36 @@ def to_repr(self) -> Dict[str, Any]:
...
+            self.batch == other.batch
+            and self.transaction == other.transaction
+        )
+
+    def __repr__(self) -> str:
+        return str(self.to_repr())
+
+    @classmethod
+    def from_repr(cls, data: Dict[str, Any]) -> 'RunQueryResult':
+        batch = QueryResultBatch.from_repr(data['batch'])
```
MarshallXie216, line 286: I think we should define `QueryResultBatch` as a class variable like `query_result_batch_kind`. This would allow downstream clients (e.g. ds lib) to override and add specific functionality.

So when we call
`batch = cls.query_result_batch_kind.from_repr(data['batch'])`
It will invoke the `from_repr` method of the client's implementation.

.github/CONTRIBUTING.rst
```diff
@@ -24,6 +24,15 @@ And run tests with:
...
 
+To run the tests that require authentication with Google, you should set the
+GOOGLE_APPLICATION_CREDENTIALS environment variable to point to your credentials
+JSON file. If you are using the ``gcloud auth application-default login`` command,
+they can be found in ``~/.config/gcloud/application_default_credentials.json``.
+
+.. code-block:: console
+
+    export GOOGLE_APPLICATION_CREDENTIALS="$HOME/.config/gcloud/application_default_credentials.json"
+
```
santiagofdialpad, line 35: We should remove this comment since it’s only valid for users within the dialpad org. We have access to the `dialpad-oss` gcp project used in the tests, but external accounts do not.

mwaehnerdialpad: understood, removed!

santiagofdialpad (approved):

PR #929, opened 2025-08-21
chore(datastore): bump version

santiagofdialpad:
## Summary

Given the changes in https://github.com/talkiq/gcloud-aio/pull/922, we need to release a new major version of datastore.
The result of `runQuery` has a breaking change


olidp (approved):
make sense! thanks santi

Issue #785, opened 2024-07-18
Chardet forces inheriting the LGPL-2.1 license 


TheCodeWrangler:
I am wanting to use `gcloud-aio` but  am concerned about inheriting the LGPL-2.1 license.

Is there any change of using an alternative to `chardet` for detecting character encodings which has a more permissive license?

PR #1010, opened 2026-03-16
chore(auth): upgrade chardet dependency upper bound to < 8.0

glinf:
## Summary

Raises the `chardet` dependency upper bound from `< 7.0` to `< 8.0` to allow chardet 7.x.

**Motivation:**
- [chardet 7.1.0](https://github.com/chardet/chardet) brings performance improvements
- License changed to MIT (more permissive)
- API is stable across the existing range, no breaking changes

Updates both `pyproject.toml` and `pyproject.rest.toml` (via `bin/build-rest`), along with their respective lock files.

Would also solve https://github.com/talkiq/gcloud-aio/issues/785


rad-pat:
It looks like requests and chardet have both moved forward, possible to also progress this now?

eedeebee:
+1 on getting this merged — the `chardet < 7.0` cap is currently the only thing forcing the LGPL `chardet` line on downstreams.

We had to work around it by pinning `chardet>=7.0.0` via a `uv` `override-dependencies` entry, which deliberately defeats `gcloud-aio-auth`'s declared `<7.0` bound. That works (chardet 7 is API-compatible — `detect()`/`detect_all()`/`UniversalDetector` are unchanged, and `gcloud-aio-auth` only uses the stable `detect()` path), but an override is a smell we'd rather not carry; relaxing the upper bound here removes the need for it entirely.

This PR looks like it just needs a rebase — it currently shows as conflicting. Would be great to get it unstuck.

One small correction for the PR description: chardet 7 relicensed to **0BSD** (public-domain-equivalent), not MIT — which makes the licensing motivation here even stronger.

TheKevJames:
Yup, we can definitely get this in now! I'll toss it into a separate commit since this has merge conflicts now.

@eedeebee note that v7 was originally MIT then got re-licensed to 0BSD instead.

TheKevJames:
Done in facd9935e9624599dec76cfe377609a78872908b, thanks!

PR #1074, opened 2026-06-26
feat(datastore): add _extra_request_fields() extension hook

fbalboaDialpad:
## Description

Adds a protected `_extra_request_fields()` extension method on the `Datastore` class, returning `{}` by default, and merges its result into the top-level request dict of each data-plane payload builder before serialization.

This is a generic subclass extension point for injecting extra top-level fields into Datastore requests without modifying the builders themselves. The default returns an empty dict, so the change is a no-op for all existing consumers — no behavior change unless a subclass overrides the method.

## Summary of changes

- Added `_extra_request_fields(self) -> dict[str, Any]` to `Datastore` with a `{}` default and a docstring describing it as a subclass extension point
- Merged the hook result into the top-level request dict in each of the seven data-plane and session-management builders (`:lookup`, `:runQuery`, `:commit`, `:beginTransaction`, `:allocateIds`, `:reserveIds`, `:rollback`). Not applied to `:export` or `:import`.
- `beginTransaction` now conditionally sends a body only when the merged dict is non-empty, keeping the current no-body behavior for the default case
- Added `datastore/tests/unit/extra_request_fields_test.py` asserting that a non-empty override appears at the top level of each data-plane body and that the default leaves bodies unchanged

### DIFF
--- a/datastore/gcloud/aio/datastore/datastore.py
+++ b/datastore/gcloud/aio/datastore/datastore.py
@@ -134,6 +134,13 @@ class Datastore:
             'Authorization': f'Bearer {token}',
         }
 
+    def _extra_request_fields(self) -> dict[str, Any]:
+        """Returns extra top-level fields to merge into every request body.
+
+        Override in a subclass to inject additional request-level fields.
+        """
+        return {}
+
     # TODO: support mutations w version specifiers, return new version (commit)
     @classmethod
     def make_mutation(
@@ -164,9 +171,11 @@ class Datastore:
         project = await self.project()
         url = f'{self._api_root}/projects/{project}:allocateIds'
 
-        payload = json.dumps({
+        body = {
             'keys': [k.to_repr() for k in keys],
-        }).encode('utf-8')
+        }
+        body.update(self._extra_request_fields())
+        payload = json.dumps(body).encode('utf-8')
 
         headers = await self.headers()
         headers.update({
@@ -192,13 +201,19 @@ class Datastore:
         project = await self.project()
         url = f'{self._api_root}/projects/{project}:beginTransaction'
         headers = await self.headers()
-        headers.update({
-            'Content-Length': '0',
-            'Content-Type': 'application/json',
-        })
+        headers['Content-Type'] = 'application/json'
+
+        extra = self._extra_request_fields()
+        kwargs: dict[str, Any] = {}
+        if extra:
+            payload = json.dumps(extra).encode('utf-8')
+            headers['Content-Length'] = str(len(payload))
+            kwargs['data'] = payload
+        else:
+            headers['Content-Length'] = '0'
 
         s = AioSession(session) if session else self.session
-        resp = await s.post(url, headers=headers, timeout=timeout)
+        resp = await s.post(url, headers=headers, timeout=timeout, **kwargs)
         data = await resp.json()
 
         transaction: str = data['transaction']
@@ -219,6 +234,7 @@ class Datastore:
             mutations, transaction=transaction,
             mode=mode,
         )
+        body.update(self._extra_request_fields())
         payload = json.dumps(body).encode('utf-8')
 
         headers = await self.headers()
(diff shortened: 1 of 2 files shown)


mwaehnerdialpad (approved):
LGTM!

datastore/gcloud/aio/datastore/datastore.py
```diff
@@ -134,6 +134,13 @@ async def headers(self) -> dict[str, str]:
             'Authorization': f'Bearer {token}',
         }
 
+    def _extra_request_fields(self) -> dict[str, Any]:
+        """Returns extra top-level fields to merge into every request body.
+
+        Override in a subclass to inject additional request-level fields.
+        """
+        return {}
+
```
shaundialpad, line 143: @fbalboaDialpad  Is the intention for these fields to be sent with every request to the datastore api? If so it might be clearer to put the final body-creation logic into a dedicated `_post` method or similar which handles the logic for us, instead of adding it _almost_ everywhere. E.g. I see `export` does not mimic the behavior, and it is not immediately clear why. Interested in hearing your thoughts since you are most familiar with how this is intended to be used!

datastore/tests/unit/datastore_test.py
```diff
@@ -41,6 +48,34 @@ async def test_build_read_options_priority():
...
+            return resp
+
+        key = Key('p', [PathElement('Kind')])
+        with patch.object(AioSession, 'post', _fake_post):
+            async with _WithSentinelField(project='p', api_root='http://dev/'
+                                          ) as ds:
+                # We use lookup as example but most methods do the same
+                await ds.lookup([key])
+
+        assert posted_body.get('sentinel') == 'injected'
```
shaundialpad, line 77: It may be worth including the other affected methods as well since, at the moment, there are subtle differences in how they construct the payload. 

As well, we check for existence of the new fields, but we don't have any guards for if any currently-existing fields get accidentally removed (for example, if the entire body is paved over or something similar). Might be worth including? Perhaps hold off until https://github.com/talkiq/gcloud-aio/pull/1074/changes#r3500771192 is addressed.

datastore/gcloud/aio/datastore/datastore.py
fbalboaDialpad: Since our use cases for this is are not publicly exposed by Google, I can't reference an official doc, but we've been told it will work for any operation handling data (lookup/commit) and transactions (beginTx, rollback). The features are related to fairness control between different workloads when the system is under pressure, so an offline export request is outside of the problem scope, which is why we don't send it there (not even sure it would be accepted in that endpoint)

datastore/tests/unit/datastore_test.py
fbalboaDialpad: > worth including the other affected methods

I'l leave that decision up to you, I can add them no problem, I just didn't want to bloat the tests with something that's mostly the same across methods, say the word and I'll push more tests.

> but we don't have any guards for if any currently-existing fields get accidentally removed

IMO that's a bit too defensive. If you prefer we can add an integration test, that one would break if for some reason we wiped out the payload dict. I don't have access to running the integration tests myself, so I couldn't verify it would work before asking for review

datastore/gcloud/aio/datastore/datastore.py
shaundialpad: > not even sure it would be accepted in that endpoint

Yeah that's what I was trying to figure out. If that's the case, then we might want to just update the methods in question to accept an optional argument for additional fields in the request, and supply those at every call. It's not _necessarily_ the case that we'd want constant parameter values for whatever it is we plan on passing in here for each method anyhow, if I am understanding the use case correctly (which I may not be). My main goal really is to demystify `_extra_request_fields` and make it useful to others using the library. If you can think of a way to do that, that would be great.

datastore/tests/unit/datastore_test.py
shaundialpad: Let's leave as is for now pending my above comment. The existing integration tests seem to cover the other concerns, IIUC.

fbalboaDialpad:
Closing in favor of #1076

shaundialpad:
@fbalboaDialpad Thanks for the ground work here!

PR #1080, opened 2026-07-08
fix: keep api_is_dev optional

shaundialpad:
## Summary

Some users currently make use of `init_api_root`, so extend #811 to
maintain backwards compatibility.


eddiedialpad (approved):
🙆‍♀️ approved: set kwarg `api_is_dev` default val is `None`

PR #1121, opened 2026-09-06
fix(storage): replace archived rsa dependency with cryptography

### LINKED ISSUE #1120: storage: replace archived `rsa` dependency with `cryptography`
`gcloud-aio-storage` depends on `rsa` and `pyasn1-modules` only to sign V4 signed URLs. [python-rsa was archived by its maintainer on 2025-04-20](https://github.com/sybrenstuvel/python-rsa) and is read-only.

Both can be dropped for `cryptography`, which is already installed transitively via `gcloud-aio-auth` (pyjwt RS256). Signatures are byte-identical, so it's a no-op for callers.

gorewilliams:
`python-rsa` was archived by its maintainer on 2025-04-20 and is read-only, so it will not receive future fixes, including security ones.

`Blob.get_pem_signature` was the only usage. `cryptography` is already installed transitively via `gcloud-aio-auth` (pyjwt RS256), and its `load_pem_private_key` reads both PKCS#1 and PKCS#8 keys directly. That removes the sole reason for the `pyasn1` machinery here: `rsa`'s `load_pkcs1` only takes PKCS#1 DER, while GCP service account keys are PKCS#8 PEM, so we had to unwrap them by hand.

PKCS#1 v1.5 is deterministic, so signatures are byte-identical to the previous implementation for both key formats, and malformed or non-RSA keys raise the same ValueError. The IAM signBlob path is unaffected.

Also adds the first credential-free unit coverage of this path, which was previously only exercised by the credentialed integration test.

N.B. the mypy-storage hook pinned gcloud-aio-auth==5.4.4, which caps cryptography at < 49 and conflicts with the pin used elsewhere in the config; bumped to 5.5.0 to match the repo.

Closes #1120.


TheKevJames (approved):
Thank you! Really appreciate the fix -- I'd totally missed the `rsa` archival.

gorewilliams:
Thx @TheKevJames! When do you expect this to be released?

